PR #38, opened 2024-09-28
Sprint 1 coursework solutions

sampennington:
## Changelist
Adds example solutions for first sprint coursework

## Questions
Do we want to keep this next to the questions, or organise all solutions elsewhere?


SallyMcGrath:
@sampennington should this be merging into a solutions branch instead?

PR #43, opened 2024-10-21
London| Elhadj Abdoul Diallo| Module-Structuring-and-Testing-Data| WEEK1

eediallo:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.
I tried to solve all the exercices of Sprint-1 of Module-Structuring-and-Testing-Data which includes the ones in the following folders:
- errors
- exercices
- explore
- interpret

## Questions

### DIFF
--- a/Sprint-1/errors/3.js
+++ b/Sprint-1/errors/3.js
@@ -1,5 +1,5 @@
-const cardNumber = 4533787178994213;
-const last4Digits = cardNumber.slice(-4);
+//const cardNumber = 4533787178994213;
+//const last4Digits = cardNumber.slice(-4);
 
 // The last4Digits variable should store the last 4 digits of cardNumber
 // However, the code isn't working
@@ -7,3 +7,11 @@ const last4Digits = cardNumber.slice(-4);
 // Then run the code and see what error it gives.
 // Consider: Why does it give this error? Is this what I predicted? If not, what's different?
 // Then try updating the expression last4Digits is assigned to, in order to get the correct value
+
+///==> my prediction: it won't work because carNumber is neither an array nor a string which means we cannot get the index the way we are doing now
+// One way to solve this is:
+// 1. Convert carNumber into a string
+// 2. subtracts 4 to the length of cardNumber to be able to extract the last 4 elements since we not dealing with and array
+const cardNumber = "4533787178994213";
+const last4Digits = +cardNumber.slice(cardNumber.length- 4);
+//console.log(last4Digits)
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -3,7 +3,10 @@ const num = 56.4567;
 // You should look up Math functions for this exercise https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math
 
 // Create a variable called wholeNumberPart and assign to it an expression that evaluates to 56 ( the whole number part of num )
+const wholeNumberPart = Math.floor(num);
+console.log(wholeNumberPart);
 // Create a variable called decimalPart and assign to it an expression that evaluates to 0.4567 ( the decimal part of num )
+
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -4,9 +4,10 @@ const num = 56.4567;
 
 // Create a variable called wholeNumberPart and assign to it an expression that evaluates to 56 ( the whole number part of num )
 const wholeNumberPart = Math.floor(num);
-console.log(wholeNumberPart);
+console.log("whole number part", wholeNumberPart);
 // Create a variable called decimalPart and assign to it an expression that evaluates to 0.4567 ( the decimal part of num )
-
+const decimalPart = num - wholeNumberPart;
+console.log("decimal part = ",decimalPart);
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -7,8 +7,8 @@ const wholeNumberPart = Math.floor(num);
 console.log("whole number part", wholeNumberPart);
 // Create a variable called decimalPart and assign to it an expression that evaluates to 0.4567 ( the decimal part of num )
 const decimalPart = num - wholeNumberPart;
-console.log("decimal part = ",decimalPart);
+console.log("decimal part = ", decimalPart);
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 const roundedNum = Math.ceil(num);
-console.log("rounded number = ",roundedNum);
+console.log("rounded number = ", roundedNum);
 // Log your variables to the console to check your answers
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -9,5 +9,6 @@ console.log("whole number part", wholeNumberPart);
 const decimalPart = num - wholeNumberPart;
 console.log("decimal part = ",decimalPart);
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
-
+const roundedNum = Math.ceil(num);
+console.log("rounded number = ",roundedNum);
 // Log your variables to the console to check your answers
--- a/Sprint-1/exercises/initials.js
+++ b/Sprint-1/exercises/initials.js
@@ -4,3 +4,5 @@ let lastName = "Johnson";
 
 // Declare a variable called initials that stores the first character of each string.
 // This should produce the string "CKJ", but you must not write the characters C, K, or J in the code of your solution.
+const initials = firstName[0] + middleName[0] + lastName[0]
+console.log(initials)
--- a/Sprint-1/exercises/initials.js
+++ b/Sprint-1/exercises/initials.js
@@ -4,5 +4,17 @@ let lastName = "Johnson";
 
 // Declare a variable called initials that stores the first character of each string.
 // This should produce the string "CKJ", but you must not write the characters C, K, or J in the code of your solution.
-const initials = firstName[0] + middleName[0] + lastName[0]
+//const initials = firstName[0] + middleName[0] + lastName[0]
+
+// make first, middle, and last name a string and convert it into an array
+const fullNameToArr = (firstName + " " + middleName + " " + lastName).split(' ')
+
+//to store the initials
+let initials = ''
+
+// get the initial of each item in the array and assign them to initials
+for (let key in fullNameToArr){
+    const value = fullNameToArr[key]
+    initials += value[0]
+}
 console.log(initials)
--- a/Sprint-1/exercises/paths.js
+++ b/Sprint-1/exercises/paths.js
@@ -1,3 +1,4 @@
+const { dir } = require('console');
 const exp = require('constants');
 const path = require('path')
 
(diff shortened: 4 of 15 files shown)


Sprint-1/errors/3.js
```diff
@@ -1,9 +1,17 @@
...
 // Then run the code and see what error it gives.
 // Consider: Why does it give this error? Is this what I predicted? If not, what's different?
 // Then try updating the expression last4Digits is assigned to, in order to get the correct value
+
+///==> my prediction: it won't work because carNumber is neither an array nor a string which means we cannot get the index the way we are doing now
+// One way to solve this is:
+// 1. Convert carNumber into a string
+// 2. subtracts 4 to the length of cardNumber to be able to extract the last 4 elements since we not dealing with and array
+const cardNumber = "4533787178994213";
+const last4Digits = +cardNumber.slice(cardNumber.length- 4);
```
illicitonion, line 16: This looks good and works, but can you think of a way of doing this which _doesn't_ require you to change how the `cardNumber` variable was declared?

i.e. so that the first line is still `const cardNumber = 4533787178994213;`, but `last4Digits` still ends up being a number containing `4213`?

Sprint-1/exercises/decimal.js
```diff
@@ -3,7 +3,12 @@ const num = 56.4567;
...
 // Create a variable called wholeNumberPart and assign to it an expression that evaluates to 56 ( the whole number part of num )
+const wholeNumberPart = Math.floor(num);
+console.log("whole number part", wholeNumberPart);
 // Create a variable called decimalPart and assign to it an expression that evaluates to 0.4567 ( the decimal part of num )
+const decimalPart = num - wholeNumberPart;
+console.log("decimal part = ", decimalPart);
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
-
+const roundedNum = Math.ceil(num);
+console.log("rounded number = ", roundedNum);
```
illicitonion, line 13: Ooh interesting! This is actually a bug in the exercise - what you've written here works for getting the right answer, but it actually doesn't quite solve the problem in the words.

I've just changed the number in https://github.com/CodeYourFuture/Module-Structuring-and-Testing-Data/pull/45 to `56.5678` - can you also change the number in your file here, and find the solution to the stated problem in words? `Math.ceil` does something slightly different from rounding to "the nearest whole number" - can you explain what it does instead?

Sprint-1/exercises/initials.js
```diff
@@ -4,3 +4,17 @@ let lastName = "Johnson";
 
 // Declare a variable called initials that stores the first character of each string.
 // This should produce the string "CKJ", but you must not write the characters C, K, or J in the code of your solution.
+//const initials = firstName[0] + middleName[0] + lastName[0]
+
+// make first, middle, and last name a string and convert it into an array
+const fullNameToArr = (firstName + " " + middleName + " " + lastName).split(' ')
```
illicitonion, line 10: This is a really interesting approach! It works, but I have some questions! Can you talk through how you came up with this solution? Can you explain a few other ways you could have solved this problem, and why this is the one you chose?

Sprint-1/exercises/paths.js
```diff
@@ -9,10 +9,15 @@
...
-const filePath = "/Users/mitch/cyf/Module-JS1/week-1/interpret/file.txt";
+const filePath = "/Users/mitch/cyf/Module-JS1/week-1/interpret/file.db";
 const lastSlashIndex = filePath.lastIndexOf("/");
 const base = filePath.slice(lastSlashIndex + 1);
 console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
+const dirPart = filePath.slice(0, lastSlashIndex)
+console.log(`The dir part of ${filePath} is ====> ${dirPart}`);
 // Create a variable to store the ext part of the variable
```
illicitonion, line 20: This works, good job!

What do you think would happen if the filename didn't contain a `.`?

What do you think _should_ happen if the filename doesn't contain a `.`?

Sprint-1/interpret/percentage-change.js
```diff
@@ -12,11 +12,38 @@ console.log(`The percentage change is ${percentageChange}`);
...
+ * let priceAfterOneYear = "8,543";
+ * const priceDifference = carPrice - priceAfterOneYear;
+ * const percentageChange = (priceDifference / carPrice) * 100;
+ * 
+ */
 
 // e) Describe what the expression Number(carPrice.replaceAll(",","")) is doing - what is the purpose of this expression?
+/*this line of code does the following:
+1. carPrice.ReplaceAll(",", ""): removes the comma in the carPrice variable / replace the comma with the no space
+2.Number(carPrice.ReplaceAll(",", ""))): convert the value into an integer.
```
illicitonion, line 47: It's really good that you're thinking about this in this order. An important thing to understand wide software is the _order_ things are evaluated, and as you identify, the `carPrice.replaceAll(...)` call happens _before_ the `Number(...)` call, even though the the `Number(...)` call is written first, because it is needed in order to have an argument value to pass to it! Nicely written up!

Sprint-1/interpret/time-format.js
```diff
@@ -1,24 +1,30 @@
...
 
 // c) Using documentation, explain what the expression movieLength % 60 represents
+    // The the expression movieLength % 60 represents calculate the number of seconds remaining after converting  the movieLength from seconds to minutes
 
 // d) Interpret line 4, what does the expression assigned to totalMinutes mean?
+ // Line 4 converts movieLength(seconds) to minutes
 
 // e) What do you think the variable result represents? Can you think of a better name for this variable?
+    // The variable result formats the date in this format: h:minutes:seconds
+    // Yes, we can give it the name formateDate for example
```
illicitonion, line 28: I'm not sure about `formateDate` as a name - can you think of any more clear names? What do you think I don't like about `formateDate`?

Sprint-1/interpret/time-format.js
```diff
@@ -1,24 +1,30 @@
...
 
 // f) Try experimenting with different values of movieLength. Will this code work for all values of movieLength? Explain your answer
```
illicitonion, line 30: Do you think the code will work for all values? It looks like you experimented, but didn't write it up?

illicitonion (commented):
This is generally looking really good, congrats! It's clear you're grasping the problems being asked, and able to turn them into correct solutions. Congrats!

If you're racing ahead, I'd encourage you to take some time to investigate the stretch exercises - they are designed to get you thinking and exploring new ideas that will help you grow :)

Please do respond do the comments, and we can discuss anything that's unclear!

Sprint-1/errors/3.js
eediallo: Yes, I think I could use the toString() method to convert the variable into a string first instead of doing it manually. Like the following
```
const last4Digits = +cardNumber.toString().slice(- 4);
```

Sprint-1/exercises/decimal.js
eediallo: Ooh yeah, actually I have found out that I should have used ```Math.round``` which returns the value of a number rounded to the nearest integer instead of ```Math.ceil``` which always rounds up and returns the smallest integer greater than or equal to a given number.

Sprint-1/exercises/initials.js
eediallo: Ohh yeah I knew I could have just gotten the first character of each string(firstName[0] , middleName[0], lastName[0]) using the their indeces and concatenate them as follows: 
```const initials = firstName[0] + middleName[0] + lastName[0]```
This is actually what I did in the first place but I want to explore more ways to do it. That's when I came up with the idea of using an array. 

This line here ```const fullNameToArr = (firstName + " " + middleName + " " + lastName).split(' ')``` concatenate the 3 strings and use the split method to convert them into an array. 

And in here ```for (let key in fullNameToArr){
    const value = fullNameToArr[key]
    initials += value[0]
}``` I am using a for in loop to get the values of the array which are going to be : Creola, Katherine and Johnson. and finally get the initial character of each value

Any other way I could have done it better?

Sprint-1/exercises/paths.js
eediallo: ```lastDotIndex``` would be -1  and ```extPart``` would return whatever the base is.

Sprint-1/interpret/time-format.js
eediallo: Maybe because we are not dealing with date here instead with time? how about formatedTime then?

Sprint-1/interpret/time-format.js
eediallo: Yeah given the fact that we are given the lenght of the movie in seconds, I would say the result variable represents the lenght of the movie formated in ```hours:minutes:seconds```.

Sprint-1/exercises/initials.js
illicitonion: Sounds good!

You'll learn more about arrays later, but one other thing you could've done is created the array directly rather than by concatenating the strings and splitting them, i.e. `const namesArray = [firstName, middleName, lastName];` - this is a bit simpler and quicker than making a string and splitting it. While it would work exactly the same in _this_ example, it does not actually do exactly the same thing. Can you think what cases would be handled _differently_ by the two approaches?

Also, note how we used different variable names - you used `fullNameToArr` and I used `namesArray` (or honestly probably just `names`). Can you think why I prefer my variable name to yours? :)

Sprint-1/exercises/paths.js
illicitonion: Yeah, that's what _would_ happen. What do you think _should_ happen? If I have a file just named `image`, and ask you what its extension is, what do you think you'd say?

Sprint-1/exercises/initials.js
eediallo: Thank your for sharing this approach. 

I think in yours, we do not need to add the seperator the array is automatically created while in mine the seperator had to be added while concatinating the names for the split method to work as intended. 

Also we would use the ```split``` method if we wanted to create a new array so that a change in the new array will not affect the original array.

I am not sure why you prefer your variable name instead of mine. Maybe because it is more meaningful ?

Sprint-1/exercises/paths.js
eediallo: if the file is of type image the extension would be either svg, png, jpeg etc... I am not sure I get the question right but that's what I would say

Sprint-1/exercises/initials.js
illicitonion: In general variable names should represent _what_ is in the variable, not _how_ it got there.

Imagine you're making a website that allows you to log in by email or by Facebook. In most of the code of your website, you would want to be able to get the ID of the user. But in most places, it doesn't matter how they logged in. You'd call the variable `userId` not `userIdFromFacebook`.

(This isn't _always_ true - in your code that specifically handles log-in, may be both a `userIdFromFacebook` variable and a `userIdFromEmail` variable specifically so you know the difference between them. But if you only have _one_ variable representing some data, you probably want a really general name describing what the data is).

Sprint-1/exercises/paths.js
illicitonion: But I'm asking: what if the file _doesn't_ have an extension?

I can create a file on my computer just named `image` that _doesn't_ have an extension.

I agree that most files _should_ have an extension, but there is nothing that _guarantees_ that. What do you think should happen if we are given a file name that _doesn't_ match our expectations?

eediallo: In this case the exPart variable will evaluate to ```image```.

illicitonion: Is that what we _want_ though? If I asked you as a human, rather than code, "What is the extension of this file name", what would you answer?

eediallo: I would say it does not have an extension.

illicitonion: Cool - what value do you think we could return from this function so that the code does the same as you would as a human? (I can think of three better values than "image"!)

eediallo: I think we could return an error message stating that the file does not have an extension or we could just return an empty string

illicitonion: An empty string I think makes sense. One disadvantage of this is that it means that getting the extension of a file named "image" or a file named "image." would give the same result - it is sometimes useful to know the difference.

I think `null` ("There was no extension") would also be a valid thing to return (compared with the empty string, when there was a `.` but nothing after it).

An error may also make sense, but we wouldn't want to _return_ an error message (because if we just return a string, we'll assume that string was the extension!) - have a read of https://medium.com/@dilarauluturhan/javascript-error-handling-try-catch-throw-finally-de5d26f16790 to learn about error handling :)

eediallo: Great! the artilce is very helpful. Thank you.

PR #47, opened 2024-10-23
Westmidlands  | Danial Bashirzadeh | Module-Structuring-and-Testing-Data | WEEK 3- Sprint 1

bashirzadehdanial:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/errors/4.js
+++ b/Sprint-1/errors/4.js
@@ -1,2 +1,5 @@
-const 12HourClockTime = "20:53";
-const 24hourClockTime = "08:53";
+const twelveHourClockTime = "20:53"; // Renamed to start with a letter, not a number
+const twentyhourClockTime = "08:53";
+console.log(twelveHourClockTime);
+console.log(twentyhourClockTime);
+
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -7,3 +7,14 @@ const num = 56.5678;
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
+
+let wholeNumberPart = Math.floor(num);
+let decimalPart = (num - wholeNumberPart).toFixed(4);
+let roundedNum = Math.ceil(num);
+
+console.log(wholeNumberPart);
+console.log(decimalPart);
+console.log(roundedNum);
+
+
+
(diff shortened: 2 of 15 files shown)


Sprint-1/errors/4.js
a-robson, line 1: Well done - you got the main point. Javascript variables can't start with a number. But are we happy with the names of the variables? Can you think how they might be improved?

a-robson:
This is looking really good. You've demonstrated you understand a lot of Javascript fundamentals. 

Please let me know if there is anything you want to discuss or clarify in my comments.

Sprint-1/errors/4.js
a-robson, line 1: Well done - you got the main point. Javascript variables can't start with a number. But are we happy with the names of the variables? Can you think how they might be improved?

Sprint-1/exercises/decimal.js
a-robson, line 1: Good job. There is an interesting ambiguity/fault in this exercise. What would the value of `roundedNum` be if the value of `num` was `56.4678`? `Math.ceil()` doesn't always round "to the nearest whole number". Can you find another method in the `Math` class that would work for all possible values of `num`?

Sprint-1/exercises/paths.js
a-robson, line 1: Ok. There are a few things going on here. 

I think you were a bit misled by the instruction. You didn't need to remove spaces from the filepath though you show nice use of a regular expression.   

When you log your results it's helpful if you include some explanation. e.g. `The dir part of ${filePath} is ${dir}` 

Can you see another way to extract the file extension without creating an array?

Sprint-1/interpret/time-format.js
a-robson, line 1: I like that you experimented with what would happen if the value of `movieLength` was negative or zero. This is exactly the sort of thing we need to consider when trying to write robust code. What do you think should happen in these cases? (I'm not sure I would really want to say the code works!).

What is the difference between movieDuration - let's follow camelCase convention :) - and movieLength? What have we accomplished between the first and last lines of the code?

Sprint-1/errors/4.js
bashirzadehdanial: Thank you for the feedback! I’ve updated the variable names to be more descriptive and consistent, using twelveHourFormatTime and twentyFourHourFormatTime to indicate the time formats.

Sprint-1/exercises/decimal.js
bashirzadehdanial: We should use Math. round(num).

Sprint-1/interpret/time-format.js
bashirzadehdanial: Thanks for the feedback! I’ve added conditions to handle negative and zero movieLength. I also updated MovieDuration to movieDuration for camelCase consistency.

PR #48, opened 2024-10-24
North West | Zabihollah Namazi | Module-Structuring-and-Testing-Data | Sprint 1 | Week 4

ZabihollahNamazi:
I will be happy if you check my answers.

## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/errors/4.js
+++ b/Sprint-1/errors/4.js
@@ -1,3 +1,3 @@
-const HourClockTime12 = "20:53";
-const hourClockTime24 = "08:53";
-console.log(hourClockTime24)
+const hourClockTime24 = "20:53";
+const hourClockTime12 = "08:53";
+console.log(hourClockTime24);
--- a/Sprint-1/errors/4.js
+++ b/Sprint-1/errors/4.js
@@ -1,2 +1,3 @@
-const 12HourClockTime = "20:53";
-const 24hourClockTime = "08:53";
+const HourClockTime12 = "20:53";
+const hourClockTime24 = "08:53";
+console.log(hourClockTime24)
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -1,5 +1,13 @@
 const num = 56.5678;
 
+let wholeNumberPart = Math.floor(num) ;
+let decimalPart = num - Math.trunc(num) ;
+let roundedNum = Math.round(num) ;
+
+console.log(wholeNumberPart)
+console.log(decimalPart)
+console.log(roundedNum)
+
 // You should look up Math functions for this exercise https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math
 
 // Create a variable called wholeNumberPart and assign to it an expression that evaluates to 56 ( the whole number part of num )
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -5,7 +5,7 @@ let decimalPart = num - Math.trunc(num) ;
 let roundedNum = Math.round(num) ;
 
 console.log(wholeNumberPart)
-console.log(decimalPart)
+console.log(decimalPart.toFixed(4))
 console.log(roundedNum)
 
 // You should look up Math functions for this exercise https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math
--- a/Sprint-1/explore/chrome.md
+++ b/Sprint-1/explore/chrome.md
@@ -16,3 +16,7 @@ Now try invoking the function `prompt` with a string input of `"What is your nam
 
 What effect does calling the `prompt` function have?
 What is the return value of `prompt`?
+
+// alert shows a small window saying "Hello World!"
+// prompt shows a small window with a input area 
+// what you write in the input area in prompt is the return value of prompt
--- a/Sprint-1/explore/objects.md
+++ b/Sprint-1/explore/objects.md
@@ -12,7 +12,7 @@ Try also entering `typeof console`
 
 Answer the following questions:
 
-What does `console` store? Answer: console is not designed to store any data. it output with messages , error and other useful data
+What does `console` store?- object -  Answer: console is not designed to store any data. it output with messages , error and other useful data
 
 What does the syntax `console.log` or `console.assert` mean? In particular, what does the `.` mean?
 console.log : to output messages,errors and variebles
--- a/Sprint-1/explore/objects.md
+++ b/Sprint-1/explore/objects.md
@@ -4,13 +4,17 @@ In this activity, we'll explore some additional concepts that you'll encounter i
 
 Open the Chrome devtools Console, type in `console.log` and then hit enter
 
-What output do you get?
+What output do you get? Answer : ƒ log() { [native code] }
 
-Now enter just `console` in the Console, what output do you get back?
+Now enter just `console` in the Console, what output do you get back? Answer: {debug: ƒ, error: ƒ, info: ƒ, log: ƒ, warn: ƒ, …}
 
 Try also entering `typeof console`
 
 Answer the following questions:
 
-What does `console` store?
+What does `console` store? Answer: console is not designed to store any data. it output with messages , error and other useful data
+
 What does the syntax `console.log` or `console.assert` mean? In particular, what does the `.` mean?
+console.log : to output messages,errors and variebles
+console.assert :  prints a message to the console only if an expression evaluates to false
+`.` : it calls functions which belongs to that object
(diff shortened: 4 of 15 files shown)


Sprint-1/interpret/to-pounds.js
```diff
@@ -15,7 +15,7 @@ const pence = paddedPenceNumberString
   .substring(paddedPenceNumberString.length - 2)
   .padEnd(2, "0");
```
cjyuan, line 16: Just an extra question. Do you think .padEnd(2, "0") is necessary?

Sprint-1/interpret/percentage-change.js
```diff
@@ -20,3 +20,17 @@ console.log(`The percentage change is ${percentageChange}`);
...
 // e) Describe what the expression Number(carPrice.replaceAll(",","")) is doing - what is the purpose of this expression?
+
+
+// a) 1-carPrice = Number(carPrice.replaceAll(",", ""));   2-priceAfterOneYear = Number(priceAfterOneYear.replaceAll("," ""));
+//       3- console.log(`The percentage change is ${percentageChange}`);
+
+// b) there should be a ',' between the argument and the replacement
+// replaceAll is not a function, you should use replace instead
+
+// c) 1- carPrice   2-priceDifference
```
cjyuan, line 31: These are not quite the "lines" with the declarations.

Sprint-1/interpret/time-format.js
```diff
@@ -22,3 +22,15 @@ console.log(result);
 // e) What do you think the variable result represents? Can you think of a better name for this variable?
 
 // f) Try experimenting with different values of movieLength. Will this code work for all values of movieLength? Explain your answer
+
+// a) 6 varieble declaration
+
+// b) one function call - console.log()
+
+// c) it divides movielength by 60 which will separate the seconds which is not enough to be one minute
```
cjyuan, line 30: can you try explain in terms of the term "remainder".

Sprint-1/interpret/time-format.js
```diff
@@ -22,3 +22,15 @@ console.log(result);
...
+
+// d) it removes remaining seconds which is not enough to be a minute and then divide the seconds to 60 to make total mitutes
```
cjyuan, line 32: You probably know what the code means, but the explanation is still not quite clear.
(Would you try ChatGPT to help you explain better?)

Sprint-1/interpret/percentage-change.js
```diff
@@ -20,3 +20,17 @@ console.log(`The percentage change is ${percentageChange}`);
 // d) Identify all the lines that are variable declarations
 
 // e) Describe what the expression Number(carPrice.replaceAll(",","")) is doing - what is the purpose of this expression?
+
+
+// a) 1-carPrice = Number(carPrice.replaceAll(",", ""));   2-priceAfterOneYear = Number(priceAfterOneYear.replaceAll("," ""));
+//       3- console.log(`The percentage change is ${percentageChange}`);
```
cjyuan, line 26: A function call is an expression and therefore can appear as a subexpression in another expression or as a parameter of 
     another function call. For example, there are 3 function calls happening in `console.log(Math.round(Number("12.34")))`.

Sprint-1/exercises/decimal.js
```diff
@@ -1,5 +1,13 @@
 const num = 56.5678;
 
+let wholeNumberPart = Math.floor(num) ;
+let decimalPart = num - Math.trunc(num) ;
+let roundedNum = Math.round(num) ;
+
+console.log(wholeNumberPart)
+console.log(decimalPart)
```
cjyuan, line 8: Your solution is correct because of round-off error with floating point number arithmetic. However, just for some challenge, can you find a way to output the decimal part of the `num` as .5678? (Currently, it is output as 0.5677999999999983)

Sprint-1/errors/4.js
```diff
@@ -1,2 +1,3 @@
-const 12HourClockTime = "20:53";
-const 24hourClockTime = "08:53";
\ No newline at end of file
+const HourClockTime12 = "20:53";
```
cjyuan, line 1: Probably a CYF mistake, but did you notice the variable names (besides the syntax error) do not quite match the values assigned to them?

Sprint-1/explore/chrome.md
```diff
@@ -16,3 +16,7 @@ Now try invoking the function `prompt` with a string input of `"What is your nam
 
 What effect does calling the `prompt` function have?
 What is the return value of `prompt`?
+
+// alert shows a small window saying "Hello World!"
+// prompt shows a small window with a input area 
+// what you write in the input area in prompt is the return value of prompt
```
cjyuan, line 22: What value does prompt() return if the user clicked "Cancel" instead of "OK"?

Sprint-1/explore/objects.md
```diff
@@ -4,13 +4,17 @@ In this activity, we'll explore some additional concepts that you'll encounter i
...
 
-Now enter just `console` in the Console, what output do you get back?
+Now enter just `console` in the Console, what output do you get back? Answer: {debug: ƒ, error: ƒ, info: ƒ, log: ƒ, warn: ƒ, …}
 
 Try also entering `typeof console`
 
 Answer the following questions:
 
-What does `console` store?
+What does `console` store? Answer: console is not designed to store any data. it output with messages , error and other useful data
```
cjyuan, line 15: The value stored in the variable `console` is an **object**. What can be found inside an object? 
What does the output `{debug: ƒ, error: ƒ, info: ƒ, log: ƒ, warn: ƒ, …}` tells you about the object stored in variable `console`?

cjyuan (commented):
Good job. I left the comments in the changed files.

Sprint-1/interpret/to-pounds.js
ZabihollahNamazi: it is not necessary but when its only one character in the pence side it will add a "0"
for example if the pence side is only a "6" pence it makes it "60" pence

Sprint-1/interpret/percentage-change.js
ZabihollahNamazi: fixed

Sprint-1/interpret/time-format.js
ZabihollahNamazi: we divided movielength by 60, every 60 will be a minute, the remainder is the the seconds not enough to be a minute.  for example 2:43:4  the 4 is teh remainder seconds

Sprint-1/interpret/time-format.js
ZabihollahNamazi: This gives the remainder when movieLength is divided by 60, which is the number of seconds left after counting whole minutes. So remainingSeconds is the number of seconds in the final minute.

Here, we subtract remainingSeconds from movieLength to get only the full minutes, then divide by 60 to convert those seconds to minutes.

Sprint-1/exercises/decimal.js
ZabihollahNamazi: I fixed it

Sprint-1/errors/4.js
ZabihollahNamazi: I fixed the variables name

Sprint-1/explore/chrome.md
ZabihollahNamazi: null

Sprint-1/explore/objects.md
ZabihollahNamazi: The {debug: ƒ, error: ƒ, info: ƒ, log: ƒ, warn: ƒ, …} output provides a quick overview of some core methods available on the console object. Each method (like ƒ for function) allows for logging different types of messages, helping with debugging, information display, and error reporting directly within the browser's console.

Sprint-1/interpret/time-format.js
cjyuan: To me "to calculate number of whole minutes" would be clear enough. (Before using ChatGPT, I wasn't aware of the term "whole minute".)

Sprint-1/interpret/to-pounds.js
cjyuan: Have you tried changing the value of `penceString` to check whether your expectation holds ( without `.padEnd(2, "0")` )?

cjyuan:
Have you pushed your changes to Github?

ZabihollahNamazi:
yes I pushed them all

Sprint-1/interpret/to-pounds.js
ZabihollahNamazi: Yes I changed it to different numbers and the answer were correct

cjyuan (commented):
Code looks good.

- May I suggest utilising ChatGPT to find out how else the same code can be described or explained (as a second opinion)? It could help you write more concise comment in long run.

PR #57, opened 2024-10-29
West Midlands | Gabriel Deng | Module-Structuring-and-Testing-Data | Sprint 1

gai93003:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/errors/3.js
+++ b/Sprint-1/errors/3.js
@@ -1,6 +1,6 @@
 const cardNumber = 4533787178994213;
-cardNumber = cardNumber.toString();
-const last4Digits = cardNumber.slice(-4);
+const cardNumberStrings = cardNumber.toString();
+const last4Digits = cardNumberStrings.slice(-4);
 
 // The last4Digits variable should store the last 4 digits of cardNumber
 // However, the code isn't working
@@ -13,3 +13,5 @@ const last4Digits = cardNumber.slice(-4);
 // I believe the the code is correct. However, the first variable is considered as a string. If operation were to be done on it, the output should first be converted into an integer.
 
 //I used toString method on the cardNumber to turn it into strings and then use slice to retrieve the last 4 digits.
+
+// I converted the cardNumber integers into a string and assigned them to a variable called cardNumberStrings, this way, I can continue using the 'const' declaration instead of using the 'let'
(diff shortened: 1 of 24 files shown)


CE-0 (commented):
errors/3.js: Consider that  the "slice" function works with the string data type. How would you convert cardNumber to a string?

errors/4.js: Your assessment is correct, but is that the only error in the original declarations? I advise that you look at the variable names in relation to their respective values and see if anything's wrong.

gai93003:
Thanks for the review. For errors/3.js: I used toString method on the cardNumber to turn it into strings and then use slice to retrieve the last 4 digits.
In the meantime, for errors/4.js: I believe the problem was that the variable names were not descriptive enough. I have changed that.
Should I open and pull request for these changes?

Sprint-1/errors/3.js
```diff
@@ -1,4 +1,5 @@
 const cardNumber = 4533787178994213;
+cardNumber = cardNumber.toString();
```
CE-0, line 2: Constants cannot be reassigned. There must be a different way to go about this.

CE-0 (changes requested):
I left some feedback. Please take a look at it.

Sprint-1/errors/3.js
gai93003: I believe the toString content can be stored in a new variable. Or I could use the let declaration instead of the const7

Sprint-1/errors/3.js
```diff
@@ -1,6 +1,6 @@
 const cardNumber = 4533787178994213;
-cardNumber = cardNumber.toString();
-const last4Digits = cardNumber.slice(-4);
+const cardNumberStrings = cardNumber.toString();
```
CE-0, line 2: Method is different from what I would've done, but ultimately gets the job done. Good work.

cjyuan:
Why not try also the questions in "explore" and "interpret" folders?

gai93003:
Alright, thank. I was planning to give them a try as well. Let me work on
them so that you can help me review it.

Sprint-1/errors/3.js
gai93003: What other method were you suggesting that I could use, I believe that it could help knowing as well.

CE-0: const cardNumber = 4533787178994213;

gai93003: Oh this way is much better since it reduces the lines of code, thanks for that. I didn't think of it.

PR #59, opened 2024-10-30
London | Forogh Aghaeiyarijani | Module-Structuring-and-Testing-Data | sprint1| week1

Forogh-Aghaeiyarijani:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/errors/2.js
+++ b/Sprint-1/errors/2.js
@@ -1,5 +1,5 @@
 // Currently trying to print the string "I was born in Bolton" but it isn't working...
 // what's the error ?
-
-console.log(`I was born in ${cityOfBirth}`);
 const cityOfBirth = "Bolton";
+console.log(`I was born in ${cityOfBirth}`);
+
--- a/Sprint-1/errors/2.js
+++ b/Sprint-1/errors/2.js
@@ -3,3 +3,4 @@
 const cityOfBirth = "Bolton";
 console.log(`I was born in ${cityOfBirth}`);
 
+//Computer reads code from the line by line, so we need to first define variables and then get console.
--- a/Sprint-1/errors/3.js
+++ b/Sprint-1/errors/3.js
@@ -9,3 +9,10 @@ const last4Digits = cardNumber.slice(-4);
 // Then try updating the expression last4Digits is assigned to, in order to get the correct value
 
 console.log(last4Digits);
+
+// If cardNumber were a number (not a string), this would raise a TypeError because numbers don't have a slice() method.
+// If cardNumber were a number, JavaScript would throw a TypeError because slice() is not a method for numbers.
+// Why String Conversion Might Not Always Be Ideal:
+// 1-Loss of Numerical Context: After converting to a string, performing arithmetic operations requires converting back to a number.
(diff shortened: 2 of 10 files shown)


Sprint-1/errors/2.js
```diff
@@ -1,5 +1,5 @@
 // Currently trying to print the string "I was born in Bolton" but it isn't working...
 // what's the error ?
```
SallyMcGrath, line 2: Try writing in your explanation as well as fixing the code - we are communicating about code here.

Sprint-1/errors/3.js
```diff
@@ -1,4 +1,4 @@
-const cardNumber = 4533787178994213;
+const cardNumber = "4533787178994213";
```
SallyMcGrath, line 1: The exercise asks you to update "the expression last4Digits is assigned to, in order to get the correct value"

Can you try doing this? In your code you have changed the data type to a string. Can you think why or when it might not be a good idea to change a number to a string?

Sprint-1/errors/4.js
```diff
@@ -1,2 +1,2 @@
-const 12HourClockTime = "20:53";
-const 24hourClockTime = "08:53";
\ No newline at end of file
+const HourClockTime = "20:53";
```
SallyMcGrath, line 1: What does 12 hour clock time mean?  What about 24 hour clock time?

Does it matter what we call things? Why do we assign labels to values, functions, expressions and so on in coding?

SallyMcGrath (commented):
Hi @Forogh-Aghaeiyarijani . Is there a bit more to do here? It looks like there are many more exercises in the folders.

Sprint-1/errors/3.js
Forogh-Aghaeiyarijani: If cardNumber were a number (not a string), this would raise a TypeError because numbers don't have a slice() method.
If cardNumber were a number, JavaScript would throw a TypeError because slice() is not a method for numbers.
 Why String Conversion Might Not Always Be Ideal:
 1-Loss of Numerical Context: After converting to a string, performing arithmetic operations requires converting back to a number.
 2-Performance Considerations: String manipulation can be less efficient than numerical operations, especially in large-scale data processing.
 3-Clarity: Maintaining a variable as a number can make the code easier to understand in contexts where numeric operations are expected. ​​

Sprint-1/errors/4.js
Forogh-Aghaeiyarijani: Definition: A system of timekeeping where the day is divided into two 12-hour periods: AM (Ante Meridiem, before noon) and PM (Post Meridiem, after noon).
 Example: 8:53 PM in 12-hour format is written as 08:53 PM.

 Definition: A system where the day runs from 00:00 (midnight) to 23:59 (one minute before the next midnight), without using AM or PM.
 Example: 20:53 in 24-hour format is 8:53 PM in 12-hour format.
 labels in coding is matters!Why? 
 1. Consistency in Naming: Helps with understanding the flow of data.
 2. Avoiding Ambiguity: Mislabeling can lead to incorrect assumptions about the data type or its usage.
 3. Efficient Communication: When collaborating, meaningful labels reduce the need for excessive comments or explanations.

PR #70, opened 2024-11-01
London 11 | Pezhman-Azizi | Structuring-and-testing-data | week1 | sprint 1 

Pezhman-Azizi:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/errors/0.js
+++ b/Sprint-1/errors/0.js
@@ -1,2 +1,7 @@
-This is just an instruction for the first activity - but it is just for human consumption
-We don't want the computer to run these 2 lines - how can we solve this problem?
+// This is just an instruction for the first activity - but it is just for human consumption
+// We don't want the computer to run these 2 lines - how can we solve this problem?
+
+// We can take an approach called `Commenting`. In different programming languages, there are different ways to solve this.
+// And it is just a matter of Syntax. In case of JavaScript for Mac we can do `command + /`
+// it means the line is not executed and it is just for sake of reading it.
+// So in the case of this exercise I want to comment out the 2 lines (line1 and line2) by selecting both lines and pressing `command + /`
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -13,7 +13,7 @@ const num = 56.5678;
 const wholeNumberPart = Math.floor(num);
 
 // the decimal part of num
-const decimalPart = num - wholeNumberPart
+const decimalPart = +(num - wholeNumberPart).toFixed(4); //.toFixed limits the decimal places in form of string. And the + sign turns the string to an integer.
 
 // num rounded to the nearest whole number
 const roundedNum = Math.round(num);
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -7,3 +7,20 @@ const num = 56.5678;
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
+
+
+// the whole number part of num
+const wholeNumberPart = Math.floor(num);
+
+// the decimal part of num
+const decimalPart = num - wholeNumberPart
+
+// num rounded to the nearest whole number
+const roundedNum = Math.round(num);
+
+
+//logging the variables to the console
+console.log(num);
+console.log(wholeNumberPart);
+console.log(decimalPart);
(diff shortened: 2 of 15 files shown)


Sprint-1/exercises/initials.js
sajadsoorni, line 1: Attention, Commiser Gorden! Nice chap chap on that code! Your use of charAt() to grab the initials is as smooth as a well-oiled machine. It helped me compare it to my approach, which is as follows:
```
let firstChar = firstName[0] + middleName[0] + lastName[0];
console.log(firstChar);
```
It's a bit like a clumsy recruit trying to find the right gear with direct indexing. Let’s keep charging ahead in this coding mission!
Stay sharp and carry on...

Sprint-1/exercises/decimal.js
```diff
@@ -7,3 +7,20 @@ const num = 56.5678;
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
+
+
+// the whole number part of num
+const wholeNumberPart = Math.floor(num);
+
+// the decimal part of num
+const decimalPart = num - wholeNumberPart
```
CE-0, line 16: Result is not what is expected. Value contains too many decimal places.

Sprint-1/exercises/random.js
```diff
@@ -3,7 +3,27 @@ const maximum = 100;
...
+
+// 1. Math.random():
+// This part, Math.random(), gives us a random decimal number between 0 and 1. It could be something like 0.234 or 0.857—always between 0 and just under 1.
+
+// 2. Making the random number bigger:
+// We then multiply that random decimal by 100 (since (maximum - minimum + 1) calculates to 100).
+// This turns the decimal into a bigger random number that’s somewhere between 0 and almost 100. For example, it could now be 45.67 or 98.23.
+
+// 3. Rounding down to a whole number:
+// Next, Math.floor(...) rounds down whatever number we have to the nearest whole number. If we had 45.67, it becomes 45. If it was 98.23, it becomes 98.
```
CE-0, line 23: Good assessment, but might I suggest that you explain the purpose behind rounding _down_ the value rather than _up_?

CE-0 (changes requested):
I have left you a few suggestions for your work. Please have a look at them.

Sprint-1/exercises/decimal.js
Pezhman-Azizi: Thanks a ton for the feedback! Just to clarify, when you mentioned "result is not what was expected," did you mean only the number of decimal places? If that’s the case, I looked into it and learned about .toFixed(4), which keeps things cleaner by limiting the decimal places.

I also learned that .toFixed() has a small catch—it actually returns the result as a string rather than a number. So, to make sure I get a number again, I’m using the + sign before the expression. It’s a quick way to convert it back to a number without affecting the value. I will go ahead and make the required changes. Thanks again for your time.

Sprint-1/exercises/random.js
Pezhman-Azizi: Thanks for the feedback! So the reason we’re rounding down with Math.floor() is to make sure we stay within the 1 to 100 range. If we rounded up, we’d risk getting a number outside that range—like going over 100. By rounding down and adding + minimum, we keep everything nicely contained.

Hope that clears it up! Let me know if there’s anything else you think I should add. Thanks again!

Sprint-1/exercises/decimal.js
```diff
@@ -13,7 +13,7 @@ const num = 56.5678;
 const wholeNumberPart = Math.floor(num);
 
 // the decimal part of num
-const decimalPart = num - wholeNumberPart
+const decimalPart = +(num - wholeNumberPart).toFixed(4); //.toFixed limits the decimal places in form of string. And the + sign turns the string to an integer.
```
CE-0, line 16: Good work on display here

Sprint-1/errors/0.js
```diff
@@ -1,2 +1,7 @@
-This is just an instruction for the first activity - but it is just for human consumption
```
Pezhman-Azizi, line 1: by commenting them out

Sprint-1/errors/0.js
```diff
@@ -1,2 +1,7 @@
...
-We don't want the computer to run these 2 lines - how can we solve this problem?
```
Pezhman-Azizi, line 2: by commenting them out

PR #80, opened 2024-11-04
CYF London 11 | Nadika Zavodovska | Structuring-and-Testing-Data | Week 1

nadika-zavodovska:
Self checklist

## Changelist

I have completed the Sprint 1 exercises.

### DIFF
--- a/Sprint-1/errors/3.js
+++ b/Sprint-1/errors/3.js
@@ -1,5 +1,5 @@
 const cardNumber = 4533787178994213;
-const last4Digits = Number(cardNumber.toString().slice(-4));
+const last4Digits = cardNumber.toString().slice(-4);
 
 // The last4Digits variable should store the last 4 digits of cardNumber
 // However, the code isn't working
--- a/Sprint-1/errors/3.js
+++ b/Sprint-1/errors/3.js
@@ -1,5 +1,5 @@
 const cardNumber = 4533787178994213;
-const last4Digits = cardNumber.slice(-4);
+const last4Digits = Number(cardNumber.toString().slice(-4));
 
 // The last4Digits variable should store the last 4 digits of cardNumber
 // However, the code isn't working
--- a/Sprint-1/exercises/random.js
+++ b/Sprint-1/exercises/random.js
@@ -1,7 +1,15 @@
 const minimum = 1;
 const maximum = 100;
 
-const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
+// Calculate how many numbers we can pick
+const valueRange = maximum - minimum + 1;
+// Get a random number and make it fit our range
+const scaledRandomNumber = Math.random() * (valueRange);
+// Round the random number down to the nearest whole number
+const roundedDownValue = Math.floor(scaledRandomNumber);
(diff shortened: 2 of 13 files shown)


Sprint-1/errors/3.js
```diff
@@ -1,5 +1,5 @@
 const cardNumber = 4533787178994213;
-const last4Digits = cardNumber.slice(-4);
+const last4Digits = Number(cardNumber.toString().slice(-4));
```
cjyuan, line 2: Is it practical to convert the last 4 digits back to Number type? What if the last 4 digits is 0000 
(No change needed, just a question to think about)

Sprint-1/exercises/random.js
```diff
@@ -1,7 +1,15 @@
...
-const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
+// Calculate how many numbers we can pick
+const valueRange = maximum - minimum + 1;
+// Get a random number and make it fit our range
+const scaledRandomNumber = Math.random() * (valueRange);
+// Round the random number down to the nearest whole number
+const roundedDownValue = Math.floor(scaledRandomNumber);
+//Add the smallest number to the rounded number to get the final number
+const num = roundedDownValue + minimum;
+console.log(num);
```
cjyuan, line 12: To test your understanding, how would you write an expression (without using any variable in the expression) that yields a random integer between -5 and 3?

Sprint-1/interpret/to-pounds.js
```diff
@@ -4,24 +4,53 @@ const penceStringWithoutTrailingP = penceString.substring(
...
 const pounds = paddedPenceNumberString.substring(
   0,
   paddedPenceNumberString.length - 2
 );
+// This code takes everything except the last two digits
 
 const pence = paddedPenceNumberString
   .substring(paddedPenceNumberString.length - 2)
   .padEnd(2, "0");
+//  This code takes the last two numbers for pence and makes sure it has 2 digits
```
cjyuan, line 22: Can we expect the code to work as intended even if we deleted `.padEnd(2, "0")` from the code?

cjyuan (commented):
Code and explanation look good. Just a few questions for you to think about.

Sprint-1/interpret/to-pounds.js
nadika-zavodovska: Hello @cjyuan,
Yes, you are right. The code will still work as intended without .padEnd(2, "0"). That is extra logic.

Sprint-1/exercises/random.js
nadika-zavodovska: `Math.floor ( Math.random( ) * (3 - (-5) + 1) ) + (-5);`

Sprint-1/errors/3.js
nadika-zavodovska: That's a good point. It's impractical to convert it to a number because if I convert '0000' to a number, it becomes 0, which isn't what I intended.

nadika-zavodovska:
Hello @cjyuan,
Thank you for reviewing my code. Your questions are very helpful. They help me understand my code better.

PR #81, opened 2024-11-04
London| Donara_Blanc | Structuring-and-Testing-Data | Week 1

donarbl:
Hello,

Self checklist

- [ x] I have committed my files one by one, on purpose, and for a reason
- [x ] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [x ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [x ] My changes meet the [requirements](./README.md) of this task

## Changelist

I have worked on the first sprint of week 1 on structuring data. 

## Questions
1. I have a question regarding the random.js. Math.random() Returns a pseudo-random number between 0 and 1. 
(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math) says this website. In the exercise we have the min =1 and max =100. My understanding is of this function 
Is that Math. random will multiple the decimals starting from 0.01 or until 0.999? However, the exercise has given min and max within the range from 1 to 100. Does this mean that the math.random will calculate from this range? Also does random is inclusive of the last number such as 1 or 100 in our case?
2. In to-pound.js the variables' names are so long that I had difficulties reading let alone understanding them. Is it a common practice to assign long variables name? 

Thank you.


Sprint-1/interpret/to-pounds.js
```diff
@@ -25,3 +25,11 @@ console.log(`£${pounds}.${pence}`);
...
 // To begin, we can start with
 // 1. const penceString = "399p": initialises a string variable with the value "399p"
+// 2. const penceStringWithoutTrailingP = penceString.substring(
+// 0, penceString.length - 1); remove the "p" in penceString so it return "399"
+// 3. const paddedPenceNumberString = penceStringWithoutTrailingP.padStart(3, "0");
+//  the variable paddedPenceNumberString has a length of 3 caracters starting from "0"-3
+// 4. const pounds = paddedPenceNumberString.substring(0,paddedPenceNumberString.length - 2);
+// the variable paddedPenceNumberString has now "3" as the value since the last digit "99" were removed by the function 
+// 5. const pence = paddedPenceNumberString .substring(paddedPenceNumberString.length - 2).padEnd(2, "0");
+// allows to have 2 digits in pence
```
cjyuan, line 35: Can we expect the code to work as intended even if we deleted `.padEnd(2, "0")` from the code?

cjyuan (commented):
Good job. I just mainly have a few questions for you to think about.

cjyuan:
Is that Math. random will multiple the decimals starting from 0.01 or until 0.999?  (Nope)

Good observation. The documentation of `Math.random()` should state clearly whether the range of the random numbers include 0/1 or not.  The expression used to generate a random numbers from 1 to 100 will make more sense if you know the exact range of random numbers `Math.random()` can return.


In to-pound.js the variables' names are so long that I had difficulties reading let alone understanding them. Is it a common practice to assign long variables name?

I would say no (especially for local variables) because it tends to make the expression involving the variables very long.
In that example, I think `paddedPenceStr` is as good as `paddedPenceNumberString`.
If needed, I would just use comments to describe the variable.

PR #83, opened 2024-11-05
WestMidlands | Danial Bashirzadeh | Module-Structuring-and-Testing-Data | Week 2 exercises ( Sprint 2 )

bashirzadehdanial:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/extend/format-time.js
+++ b/Sprint-2/extend/format-time.js
@@ -3,10 +3,16 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
-  if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+
+  if (hours === 0) {
+    return `12:00 am`; // Midnight
+  } else if (hours === 12) {
+    return `12:00 pm`; // Noon 
+  } else if (hours > 12) {
+    return `${hours - 12}:00 pm`; // Afternoon
+  } else {
+    return `${hours}:00 am`; // Morning 
   }
-  return `${time} am`;
 }
 
 const currentOutput = formatAs12HourClock("08:00");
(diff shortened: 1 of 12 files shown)


Sprint-2/extend/format-time.js
```diff
@@ -3,10 +3,16 @@
...
+
+  if (hours === 0) {
+    return `12:00 am`; // Midnight
+  } else if (hours === 12) {
+    return `12:00 pm`; // Noon 
+  } else if (hours > 12) {
+    return `${hours - 12}:00 pm`; // Afternoon
+  } else {
+    return `0${hours}:00 am`; // Morning 
   }
```
YoanHlebarov, line 15: Well done, this works well in the test cases described. Can you think about how we can improve this to handle inputs such as `formatAs12HourClock("7:30")` for example? 
Consider what other ways you might be able to split the string, where you can be certain which part is the hours and which part is the minutes.
Add test to assert that these inputs will be handled correctly.

Sprint-2/implement/bmi.js
YoanHlebarov, line 1: Great job, this works really well. What would happen if the arguments to the function were `weight = -80` and `height = 185`? Can you think about ways you can metigate that, or log out what is/not supported?

Sprint-2/implement/vat.js
YoanHlebarov, line 1: That is good, well done. This works well if the VAT is a decimal number above 1. In most cases people might add VAT as apercentage, for example 20%. Can you think of a way  were we could input something like `priceWithVatCalculation(99, 20);` where the `rate` is 20%?

Sprint-2/extend/format-time.js
bashirzadehdanial: Thank you for the feedback, Yoan! I propose splitting the input string into hours and minutes based on the colon (":") to address your point. I also added some tests to assert that these inputs will be handled correctly.

Sprint-2/implement/bmi.js
bashirzadehdanial: Thank you for the positive feedback! I have added a condition to validate inputs at the top part of the function.

Sprint-2/implement/vat.js
bashirzadehdanial: Thanks for your feedback! I've added a condition to check if the input is with a percentage sign, to handle this scenario as well.

YoanHlebarov (commented):
Nicely done, that looks much better!

PR #88, opened 2024-11-05
Structuring and testing exercise 1

HaftamuKebedew:
London | Haftamu Kebedew |  Module-Structuring-and-Testing-Data-Sprint-1 | WEEK 1

## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.
I completed the questions included in the Exercises, Errors, and Interpret folders in Sprint-1. I try to answer all of the Javascript files in this folder.So please take a look, and I would appreciate a feedback.
## Questions

In this activity, I saw some functions as replace, subString, slice, and Math.round. Can we write our own functions as well?

### DIFF
--- a/Sprint-1/errors/3.js
+++ b/Sprint-1/errors/3.js
@@ -1,4 +1,4 @@
-const cardNumber = 4533787178994213;
+const cardNumber = '4533787178994213';
 const last4Digits = cardNumber.slice(-4);
 
 // The last4Digits variable should store the last 4 digits of cardNumber
@@ -7,3 +7,10 @@ const last4Digits = cardNumber.slice(-4);
 // Then run the code and see what error it gives.
 // Consider: Why does it give this error? Is this what I predicted? If not, what's different?
 // Then try updating the expression last4Digits is assigned to, in order to get the correct value
+ console.log(last4Digits)
+
+/*
+TypeError. cardNumber is not an array or string being slice works with those types. so I get the same result with what I gess before when i run the code
+To get the last four digits of the cardNumber I have to change the cardNumber to String.
+
+*/
--- a/Sprint-1/exercises/count.js
+++ b/Sprint-1/exercises/count.js
@@ -1,6 +1,6 @@
 let count = 0;
 
-count = count + 1;
+count = count + 1; //count increased by one so addion operaiton is perfomed here.
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -7,3 +7,9 @@ const num = 56.5678;
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
+
+const wholeNumberPart=parseInt(num);
+const decimalPart=(num%1).toFixed(4);// const decimalPart=(num-Math.floor(num)).toFixed(4);
+const roundedNum=Math.round(num);
+
+console.log(` wholeNumberPart= ${wholeNumberPart}\n decimalPart= ${decimalPart}\n roundedNum= ${roundedNum}`)
--- a/Sprint-1/exercises/paths.js
+++ b/Sprint-1/exercises/paths.js
@@ -16,3 +16,7 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
+
+const dir=filePath.slice(0, lastSlashIndex).slice(1);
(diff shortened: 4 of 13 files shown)


Sprint-1/errors/3.js
```diff
@@ -7,3 +7,10 @@ const last4Digits = cardNumber.slice(-4);
 // Then run the code and see what error it gives.
 // Consider: Why does it give this error? Is this what I predicted? If not, what's different?
 // Then try updating the expression last4Digits is assigned to, in order to get the correct value
+ console.log(last4Digits)
+
+/*
+TypeError. cardNumber is not an array or string being slice works with those types. so I get the same result with what I gess before when i run the code
```
Emiliecg, line 13: This works! Can you figure out another way to fix the problem by changing the expression last4Digits is assigned to instead of changing the type of cardNumber?

Sprint-1/exercises/count.js
```diff
@@ -1,6 +1,6 @@
 let count = 0;
 
-count = count + 1;
+count = count + 1; //count increased by one so addion operaiton is perfomed here.
```
Emiliecg, line 3: That's the right idea. Can you get more specific about what "=" is doing in this line?

Sprint-1/exercises/decimal.js
```diff
@@ -7,3 +7,9 @@ const num = 56.5678;
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
+
+const wholeNumberPart=parseInt(num);
+const decimalPart=(num%1).toFixed(4);// const decimalPart=(num-Math.floor(num)).toFixed(4);
```
Emiliecg, line 12: Think about how you might be able to modify the expression to get the decimal part so that it would work for any value of num, even if it has a different number of digits. Could you make use of your previous result?

Sprint-1/exercises/paths.js
```diff
@@ -16,3 +16,7 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
+
+const dir=filePath.slice(0, lastSlashIndex).slice(1);
```
Emiliecg, line 20: dir includes the root, so you don't have to get rid of the first '/' character here.

Sprint-1/interpret/time-format.js
```diff
@@ -22,3 +23,22 @@ console.log(result);
...
+
+b) 1
+
+c) This uses a  modulus operator (%) to find the remainder when movieLength is divided by 60. 
+
+d) Calculating the total number of full minutes in movieLength. 
+movieLength represents the total duration of the movie in seconds ,remainingSeconds is the number of seconds left after accounting for all full minutes (from the previous line).
+By subtracting remainingSeconds from movieLength, we get only the seconds that make up complete minutes, effectively rounding down to the nearest minute.
+
+e)movieDuration
```
Emiliecg, line 40: The original movieLength variable also represented the duration of the movie. What's different about the result variable, and how might we call it to better communicate that?

Sprint-1/interpret/time-format.js
```diff
@@ -22,3 +23,22 @@ console.log(result);
...
+
+f) Yes, it works correctly for all values.
```
Emiliecg, line 42: Try exploring different, potentially unexpected values for the movieLength variable. Can you come up with something that this code wouldn't handle?

Emiliecg (commented):
Great work 👏 I've just left a few small suggestions for improvement.

Sprint-1/errors/3.js
HaftamuKebedew: I have updated. by converting the number to string and extracting its last 4 characters instead of change its type. Thank you!

Sprint-1/exercises/count.js
HaftamuKebedew: I make some modifications. thank you.

Sprint-1/exercises/decimal.js
HaftamuKebedew: I have used the previous expressions for the decimalPart.

Sprint-1/exercises/paths.js
HaftamuKebedew: I have removed slice to include the root directory . so now dir value will include root path

Sprint-1/interpret/time-format.js
HaftamuKebedew: I have updated movieDuration to  timer instead of result.

Sprint-1/interpret/time-format.js
HaftamuKebedew: Here it works fine for positive  numbers. If the value of movieDuration is negative or fraction it will generate error

HaftamuKebedew: I endeavour to respond to every comments according to my comprehension, thus I would appreciate any feedback on my responses.

I greatly appreciate your time and constructive input!!!

PR #91, opened 2024-11-06
CYF - Lndon | Anna Fedyna | Module-Structuring-and-Testing-Data | Week 2

annafedyna:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions


Sprint-1/errors/4.js
```diff
@@ -1,2 +1,2 @@
-const 12HourClockTime = "20:53";
-const 24hourClockTime = "08:53";
\ No newline at end of file
+const hourClockTime12 = "20:53";
+const hourClockTime24 = "08:53";
```
CE-0, line 2: Assessment of problem is correct, but there seem further issues with the original variable declarations. I suggest taking a closer at the values of the variables in relation to the variable names to find the issue.

Sprint-1/exercises/count.js
```diff
@@ -4,3 +4,5 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+
+/* the = operator here is taking the result of the expression on the right and assigning it to variable count */
```
CE-0, line 8: I feel it would help to provide slightly more explanation as to what's happening here.

Sprint-1/exercises/decimal.js
```diff
@@ -7,3 +7,11 @@ const num = 56.5678;
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
+ 
+const wholeNumberPart = Math.floor(num);
+const decimalPart = num - Math.floor(num);
```
CE-0, line 12: This does not give the expected result, as it has too many decimal places. Perhaps there is a way to limit them.

Sprint-1/exercises/decimal.js
```diff
@@ -7,3 +7,11 @@ const num = 56.5678;
...
+const roundedNum = Math.round(num);
+
+
+
+console.log(roundedNum);
```
CE-0, line 17: You were asked to log all the variables, not just one.

Sprint-1/exercises/paths.js
```diff
@@ -15,4 +15,8 @@ const base = filePath.slice(lastSlashIndex + 1);
 console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
+const dir = filePath.slice(0, lastSlashIndex);
+
 // Create a variable to store the ext part of the variable
+const extPart = base.slice(base.lastIndexOf('.')+1)
```
CE-0, line 21: Look at the diagram above and see if it matches up exactly with your solution.

Sprint-1/exercises/random.js
```diff
@@ -7,3 +7,15 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+/*
+    (maximum - minimum + 1) = 100
+    Math.random() - generates a random floating-point number between (0, 1)
+    Math.floor() - rounds down to the nearest whole number
```
CE-0, line 14: What is the necessity for rounding _down_ rather than rounding _up_?

Sprint-1/exercises/random.js
```diff
@@ -7,3 +7,15 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+/*
+    (maximum - minimum + 1) = 100
```
CE-0, line 12: What is the purpose behind the multiplication?

CE-0 (changes requested):
I have left you some feedback.

Sprint-1/errors/4.js
annafedyna: Right,  thank you,  the variable names and values were mismatched. 
It should be :
`const hourClockTime24 = "20:53";
const hourClockTime12 = "08:53";`

Sprint-1/exercises/count.js
annafedyna: Of course, I will add here a bit more detailed explanation about what's happening with the = operator in JavaScript. So, in line 3, the  = operator is used to assign the value on the right-hand side to the variable on the left-hand side. We have  count = count + 1 . First, the value of count (which is initially 0) is accessed. Then, 1 is added to it, resulting in 1.This value (1) is assigned back to the variable count. The = operator here doesn't mean "equal" . Instead, it acts as an assignment operator which updates the variable with the new value.

Sprint-1/exercises/decimal.js
annafedyna: We can use toFixed() to limit decimal places

Sprint-1/exercises/decimal.js
annafedyna: Thank you for pointing out!

Sprint-1/exercises/random.js
annafedyna: The purpose is to scale the random value

Sprint-1/exercises/random.js
annafedyna: So that num is not bigger than 100

Sprint-1/exercises/paths.js
annafedyna: Yes, after reviewing the diagram and running the code it matches,  please let me know if solution not aligns correctly

CE-0: Please take note of all the _characters_ under the 'base' column

cjyuan:
I suppose this PR is for sprint-3 (since it says week 3)?
However, the current branch was created from Sprint-2 branch, and not from main. As a result, it contain modified files in both Sprint-1 and Sprint=2. Can you try to rebase the current branch so that it is branched out from main?

Apparently, this is a common mistake. See if can fix the issue from the suggestions found on Slack:
https://codeyourfutur-yov6609.slack.com/archives/C07PVEVQG9X/p1731018581934559

annafedyna:
> I suppose this PR is for sprint-3 (since it says week 3)? However, the current branch was created from Sprint-2 branch, and not from main. As a result, it contain modified files in both Sprint-1 and Sprint=2. Can you try to rebase the current branch so that it is branched out from main?
> 
> Apparently, this is a common mistake. See if can fix the issue from the suggestions found on Slack: https://codeyourfutur-yov6609.slack.com/archives/C07PVEVQG9X/p1731018581934559

Thank you very much for pointing out ! I rebased two branches.

PR #105, opened 2024-11-08
North West | Zabihollah Namazi | Module-Structuring-and-Testing-Data | Sprint 2

ZabihollahNamazi:
I'll appreciate it if you can review my pull request.

### DIFF
--- a/Sprint-2/debug/1.js
+++ b/Sprint-2/debug/1.js
@@ -1,8 +1,10 @@
 // Predict and explain first...
 
 function sum(a, b) {
-  return;
-  a + b;
+  return a + b;
 }
 
 console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
+
+// teh result will be the text and "undefined" 
+// because we returned the function without giving it functionality, (a+b) should be infront of return
--- a/Sprint-2/debug/1.js
+++ b/Sprint-2/debug/1.js
@@ -1,10 +1,11 @@
 // Predict and explain first...
 
 function sum(a, b) {
-  return a + b;
+  return a + b; 
 }
 
-console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
+
+console.log(`The sum of 10 and 30 is ${sum(10, 30)}`);
 
 // teh result will be the text and "undefined" 
 // because we returned the function without giving it functionality, (a+b) should be infront of return
--- a/Sprint-2/debug/2.js
+++ b/Sprint-2/debug/2.js
@@ -2,8 +2,8 @@
 
 const num = 103;
 
-function getLastDigit() {
-  return num.toString().slice(-1);
+function getLastDigit(number) {
+  return number.toString().slice(-1);
 }
 
 console.log(`The last digit of 42 is ${getLastDigit(42)}`);
@@ -12,3 +12,7 @@ console.log(`The last digit of 806 is ${getLastDigit(806)}`);
 
 // This program should tell the user the last digit of each number.
 // Explain why getLastDigit is not working properly - correct the problem
+
+// it will show "3" for all each number. 
+// because we defined num variable and gave it a number and each time the function is using that number
+// we have to use a variable which we give the number each time and the function can use it
--- a/Sprint-2/debug/2.js
+++ b/Sprint-2/debug/2.js
@@ -1,9 +1,7 @@
 // Predict and explain first...
 
-const num = 103;
-
-function getLastDigit(number) {
-  return number.toString().slice(-1);
+function getLastDigit(num) {
+  return num.toString().slice(-1);
 }
 
 console.log(`The last digit of 42 is ${getLastDigit(42)}`);
--- a/Sprint-2/errors/0.js
+++ b/Sprint-2/errors/0.js
@@ -4,6 +4,12 @@
 // interpret the error message and figure out why an error is occurring
 
 function capitalise(str) {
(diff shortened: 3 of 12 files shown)


Sprint-2/debug/1.js
```diff
@@ -1,8 +1,10 @@
...
 function sum(a, b) {
-  return;
-  a + b;
+  return a + b;
 }
 
 console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
+
+// teh result will be the text and "undefined" 
+// because we returned the function without giving it functionality, (a+b) should be infront of return
```
jonnywyatt, line 10: Yes, the ; after the return on line 4, finishes the statement so the function just returns `undefined`

Sprint-2/debug/2.js
```diff
@@ -2,8 +2,8 @@
 
 const num = 103;
 
-function getLastDigit() {
-  return num.toString().slice(-1);
+function getLastDigit(number) {
```
jonnywyatt, line 5: This works fine although `number` is very close to the Javascript function `Number` so might be a bit confusing for someone to read, how about using `num` as the parameter name instead?

jonnywyatt: Also, is the `num` declaration on line 3 still needed? If not, it's best to remove it.

Sprint-2/errors/0.js
```diff
@@ -4,6 +4,12 @@
...
 function capitalise(str) {
-  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+  str = `${str[0].toUpperCase()}${str.slice(1)}`;
   return str;
 }
+
+console.log(capitalise("hello"));
+
+// it gives an error "str has already been declared" so we don't need to define it again in line 7
+// in line 6 - we give the function the value or variable so in line 7- we don't nee to define the variable
```
jonnywyatt, line 14: Yes correct, `str` in line 6 is the function's 'parameter'

Sprint-2/extend/format-time.js
```diff
@@ -4,13 +4,14 @@
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
   if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+    let x = `${hours - 12}`.padStart(2,"0");
+    return x+`:00 pm`;
```
jonnywyatt, line 8: What if the `time` passed to this function has minutes that aren't zero?
eg '13:30'

Sprint-2/implement/bmi.js
```diff
@@ -13,3 +13,11 @@
 // Given someone's weight in kg and height in metres
 // Then when we call this function with the weight and height
 // It should return their Body Mass Index to 1 decimal place
+
+function bmi(weight, height){
```
jonnywyatt, line 17: If weight should always be in kg and height in m, how could you rename these two parameters to make that a bit clearer to someone else who might use your function in future?

Sprint-2/implement/cases.js
```diff
@@ -13,3 +13,21 @@
 
 // You will need to come up with an appropriate name for the function
 // Use the string documentation to help you find a solution
+
+var caps = ["A", "B", "C", "D", "E", "F", "G", "H", "K", "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z"]
```
jonnywyatt, line 17: This function works, well done.
Is there a way to simplify it? You could think about what the function needs to do, in two steps:
1. convert the whole `txt` string to upper case
2. replace any spaces with `_`

There's a string method called `toUpperCase()`, can you use it to do the case conversion in one line?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase

Sprint-2/implement/vat.js
```diff
@@ -8,3 +8,10 @@
 // Given a number,
 // When I call this function with a number
 // it returns the new price with VAT added on
+
+function price_with_VAT(price){
```
jonnywyatt, line 12: If price should be in £, could you rename the parameter `price` to make that clearer to someone who might use your function?

jonnywyatt (commented):
Well done! You've thought through all the problems.
I've made some suggestions, see what you think?

Sprint-2/debug/1.js
ZabihollahNamazi: I checked it and it returns correct result

Sprint-2/debug/2.js
ZabihollahNamazi: thanks. I applied your suggestions

Sprint-2/errors/0.js
ZabihollahNamazi: thats right

Sprint-2/extend/format-time.js
ZabihollahNamazi: I changed it and now you can put minutes too, thanks

Sprint-2/implement/bmi.js
ZabihollahNamazi: I changed the parameters name

Sprint-2/implement/cases.js
ZabihollahNamazi: I made it more simple and used toUpperCase() too. thanks

Sprint-2/implement/vat.js
ZabihollahNamazi: updated

PR #107, opened 2024-11-08
time-format

Amir200524:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/extend/format-time.js
+++ b/Sprint-2/extend/format-time.js
@@ -22,3 +22,34 @@ console.assert(
   currentOutput2 === targetOutput2,
   `current output: ${currentOutput2}, target output: ${targetOutput2}`
 );
+
+//If the input time is "00:00", the function returns "0:00 am", which is incorrect. Midnight should be represented as "12:00 am".
+//If the input is "12:00", the function returns "12:00 am", which is incorrect. Noon should be represented as "12:00 pm".
+//If the time has a single digit for the hour (e.g., "03:00"), the function will return it in the format "3:00 am", which is missing a leading zero. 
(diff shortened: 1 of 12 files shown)


Sprint-2/extend/format-time.js
```diff
@@ -22,3 +22,34 @@ console.assert(
...
+//If the time has a single digit for the hour (e.g., "03:00"), the function will return it in the format "3:00 am", which is missing a leading zero. 
+//There should be some validation for invalid input (e.g., strings not in a valid time format, non-numeric strings, etc.).
+//The function doesn't handle minutes correctly.
+//Fixed code:
+function formatAs12HourClock(time) {
+  if (!/^\d{2}:\d{2}$/.test(time)) {
+    throw new Error("Invalid time format. Use 'hh:mm' format.");
+  }
+  const hours = Number(time.slice(0, 2));
+  const minutes = time.slice(3, 5);
```
YoanHlebarov, line 37: This works really well, but can you think of a more effecient and easier to read way you can split a string such as "HH:MM"?

Sprint-2/extend/format-time.js
YoanHlebarov, line 1: Great job here. You are validating user input first which is fantastic. One edge case to think about, what happens when you call `formatAs12HourClock("27:30");?

Sprint-2/implement/bmi.js
YoanHlebarov, line 1: That's very good, in your `formatAs12HourClock` function you are validating user input which is best practice. Do you think it would be wise to do the same here? For example if someone puts negavite integer, or puts their height as an integer (eg 185) rather than decimal?

Amir200524: Hi @YoanHlebarov. Thanks for your the tips.

PR #121, opened 2024-11-11
ITP-2 Sprint01 | Stanislav Starishko | Module-Structuring-and-Testing-Data | Week 1

StanStarishko:
## Learners, PR Template

Self checklist

- [✅] I have committed my files one by one, on purpose, and for a reason
- [✅] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [✅] I have tested my changes
- [✅] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [✅] My changes meet the [requirements](./README.md) of this task

## I have completed the exercises for **Sprint-1**
in the folders:
- **errors**
- **exercises**
- **explore**
- **interpret**


Sprint-1/errors/3.js
```diff
@@ -1,5 +1,6 @@
-const cardNumber = 4533787178994213;
-const last4Digits = cardNumber.slice(-4);
+//const cardNumber = 4533787178994213;
+const cardNumber = "4533787178994213"; // slice() is a method for Array, and string is an Array of characters
```
cjyuan, line 2: If you are not allowed to modify the original value of `cardNumber` (which is of a number type), how would you rewrite your code to get the last 4 digits?

Sprint-1/exercises/count.js
```diff
@@ -2,5 +2,10 @@ let count = 0;
 
 count = count + 1;
 
+/* The current value of the variable count 
+is taken and 1 is added to it. The result 
+is assigned to the variable count
```
cjyuan, line 7: There is a shorter way to describe statement like this. May I suggest feeding the code to ChatGPT to see how else the code can be described?

Sprint-1/exercises/decimal.js
```diff
@@ -1,4 +1,24 @@
+// Create a variable called wholeNumberPart and assign to it an expression that evaluates to 56 ( the whole number part of num )
+function wholeNumberPart(inputNumber) {
+    return Math.trunc(inputNumber);
+}
+
+// Create a variable called decimalPart and assign to it an expression that evaluates to 0.5678 ( the decimal part of num )
+function decimalPart(inputNumber) {
```
cjyuan, line 7: This is a clever method but it won't work properly when inputNumber is `0.000000002`.

Sprint-1/exercises/random.js
```diff
@@ -1,7 +1,19 @@
...
+
+console.log(`Random = ${randomNum}`);
+console.log(`Rounder = ${maximum - minimum + 1}`);
+console.log(`Multiplied = ${randomNum * (maximum - minimum + 1)}`);
+console.log(`Rounded num = ${Math.floor(randomNum * (maximum - minimum + 1))}`);
+console.log(`num = ${num}`);
+
+/* I saw here getting a random number from 1 to 100, 
+to which the minimum number is added
+ */
```
cjyuan, line 16: To test your understanding, how would you write an expression (without using any variable in the expression) that yields a random integer between -5 and 3 (including -5 and 3)?

Sprint-1/explore/chrome.md
```diff
@@ -16,3 +16,7 @@ Now try invoking the function `prompt` with a string input of `"What is your nam
 
 What effect does calling the `prompt` function have?
 What is the return value of `prompt`?
+
+
+**The *"input"* is executed as seen in the screenshot above. The *string value* of the input is returned:**
+![alt text](image.png)
```
cjyuan, line 22: If you were writing a program that uses `prompt()` to ask for an input value, how can your program tell if the user clicked "OK" or "Cancel"?

Sprint-1/interpret/time-format.js
```diff
@@ -13,12 +13,76 @@ console.log(result);
...
+const remainingMinutes
+const totalHours
+const result
+ */
+
 // b) How many function calls are there?
 
+/* Taking into account the string formatting with `$`, 
+there are 4 in total:
+line 9: $ 3(three) times
```
cjyuan, line 29: Why do you consider `${...}` in the string template literal a function call? And what function was called?

Sprint-1/interpret/time-format.js
```diff
@@ -13,12 +13,76 @@ console.log(result);
...
+line 10: log
+ */
+
 // c) Using documentation, explain what the expression movieLength % 60 represents
 
+/* The expression `movieLength % 60` uses the modulo operator (`%`) 
+to calculate the remainder when `movieLength` 
+(representing a movie's length in minutes) is divided by 60.
```
cjyuan, line 37: The comment in line 1 says the number stored in `movieLength` is of a different unit (not minutes)

Sprint-1/interpret/to-pounds.js
```diff
@@ -25,3 +25,27 @@ console.log(`£${pounds}.${pence}`);
...
+3. Pad the pence number string with leading zeros to make it 3 characters long:
+   3.1. Use the `padStart(3, "0")` method on `penceStringWithoutTrailingP` 
+   to add leading zeros, saving the result in `paddedPenceNumberString`
+
+4. Extract the number of pounds:
+   4.1. Get the substring of the first 1-2 digits from 
+   `paddedPenceNumberString`, which will give the number 
+   of pounds, saved as `pounds`
+
+5. Extract the number of pence:
```
cjyuan, line 45: Can we expect the code to work as intended even if we deleted `.padEnd(2, "0")` from the code?

cjyuan (commented):
Good job. Just a few questions for you to think about.

Sprint-1/errors/3.js
StanStarishko: Thanks CJ for such interesting questions
I'll start with the fact that theory is not my strong point, but I'll try to give more precise answers
In the programming language I'm switching from now, I prefer these constructions:
```
```
this guarantees that it will work for any type of `cardNumber`. But we can still come up with "scenarios" when this code won't work

Sprint-1/exercises/count.js
StanStarishko: Without ChatGPT I would assume that it is
`count += 1;`

But just in case I will check if I am right

Hmm, and an interesting result, besides my version, there is another version:
`count++;`

This is basically logical, because Java also uses such an increment construct. But a long time ago I used it in` C/C++`. Thanks CJ for such pleasant nostalgia

Sprint-1/exercises/decimal.js
StanStarishko: My original version of this function was:
`return Math.abs(inputNumber - wholeNumberPart(inputNumber));`

but it gave the result:
`decimalPart from 56.5678 = 0.56779999999999983`

which in my opinion in the context of the problem statement is not correct

if we take your example with `0.000000002`, then my version, and I think you hinted at my original version, both options will give the result but with a number in the exponent format. To do everything "nicely" you need to write a full-fledged conversion function from the "exponent". But there is one "but", if the bit depth is more than 15, then you need to connect, for example, the `decimal.js` library
In **Python**, I use a similar library for such "tricky" problems. I suggest limiting ourselves to this comment here because I demonstrated an understanding of the problem and the way to solve it.

Sprint-1/exercises/random.js
StanStarishko: `Math.floor(Math.random() * 9) - 5`

`Math.floor(Math.random() * 9)` - getting random numbers from 0 to 8
Then, 0 - 5 = -5 and 8 - 5 = 3

Sprint-1/explore/chrome.md
StanStarishko: - If the user clicks "OK" without entering anything, `prompt()` returns an empty string `""`.
- If the user clicks "OK" after entering a value, `prompt()` returns that value as a string.
- If the user clicks "Cancel", `prompt()` returns `null`

Sprint-1/interpret/time-format.js
StanStarishko: I already wrote above that theory is not my strong point. I can only answer that it is not me who considers it a function. Perhaps, say that the JS interpreter consider such as, it will be more correct. Maybe the language design, honestly I do not know how to explain it correctly. I just know that is a function, if you do a decomposition, for example, to the level of Assembler or machine code. I hope I was able to defend my opinion, assertion and understanding

If I am wrong, I will be happy to hear your more expert opinion and will take it into service

Sprint-1/interpret/to-pounds.js
StanStarishko: Yes, I think that if you remove this part of the code, everything will work as intended.
When I was designing the algorithm for this script, I noticed this and thought "why?". After all, we first make a string of length 3 in case there is one significant penny. There are no separators, therefore we will not have a situation where we need to, for example, convert 2 to 20. If we had, for example, 3.2 at the input, where 3 is pounds and 2 is 20 pence, this code would be simply necessary. But here, this is just an extra load on the processor, system interfaces and data buses. In the context of our course, this is **dead code**.

Sprint-1/exercises/count.js
cjyuan: I was not concerning about the code or your understanding (which you are doing great); I was focusing on "describing code" in English. For example, a more concise description would be "increment count by 1".

Sprint-1/exercises/decimal.js
cjyuan: `Math.abs(inputNumber - wholeNumberPart(inputNumber))`, even if it yield 0.56779999999999983, is perfectly 
normal and acceptable solution. That's just the nature of floating point arithmetic. You will probably get the same result in any language that uses double precision floating point number. 

If we want to show only 4 decimal places, we can "format" the output to have only 4 decimal places without changing the value of the variable. 

However, if we were to use an approach that depends on converting the number to a String first, we need to be aware how number type is converted to string. In the example I showed you, the number is converted to a string in scientific format and as a result your code won't work properly.

Sprint-1/interpret/time-format.js
cjyuan: You can always try ChatGPT for a second opinion, even though it may not always give correct answer.

And that's why I did not say you were right or wrong because I don't know exactly how the JS interpreter do when it parse the template string literal. For all I know, the interpreter can indeed be calling some function to obtain the string value  in the ${....] expression.

Sprint-1/exercises/count.js
StanStarishko: Thank you, CJ for developing my vocabulary. I am very grateful for that
Tell me, is my phrase "increment construct" bad English? Is it better to say "increment count by N"?

Sprint-1/exercises/decimal.js
StanStarishko: The whole point is that any solution path will be converted to scientific format if it is an exponent. The conversion occurs when the result is outputting to the console. Most likely, a background conversion to a "string" occurs, that's why. I tried to explain this in my comment, but apparently, I didn't do it very cleanly.
example:
`decimalPart from 56.5678 = 0.567799999999999983`
was given to explain why I converted to a string right away in my solution. There is no doubt that this is normal from a mathematical point of view, but I tried to make a "universal" visualisation, forgetting that the exponent is visualised in scientific format by default in many programming languages

For numbers like `0.000000002`, you can play around with methods `.toFixed()` and/or `.toPrecision()` not forgetting about `.substring()`
But if such a number has more than 14 digits we get a "cut" number to an integer, after that we need to connect a library, for example `decimal .js`, as I wrote above.

That is, I wanted to say that I could not find a solution for outputting the fractional part for numbers like `0.000000002` without writing a conversion function from scientific format.

I came to this conclusion after a series of studies and experiments. Perhaps I did them wrong.

Shortly, I tried to get a string from the float and cut off only the decimal part, forgetting numbers like `0.000000002`. If I had remembered it, I would have written a solution that performs the conversion from scientific format. But then we still run into the float limitation of no more than 17 decimal places. Frankly, I did not dig further to see if it was possible to get a "longer" float. On the stack I worked on before, I had a float limitation of 26 decimal places. If I understood correctly, then JavaScript has a limitation of 17.

Sprint-1/interpret/time-format.js
StanStarishko: Thanks to CJ for this discussion.

I tortured ChatGPT with questions and this is what I got in the end:
I understand what you mean by an "internal function" of the interpreter. Yes, ***interpolation*** can be thought of as a kind of built-in interpreter function that automatically evaluates expressions and converts them to strings. Although interpolation isn’t a standalone function, the interpreter does handle ${...}, evaluating the expression inside and inserting the result into the string. In this sense, it’s an internal mechanism, somewhat like a “hidden” function that the interpreter executes to simplify string handling.  😊

PR #144, opened 2024-11-14
CYF London | Sajad Soornisofla | Module-Structuring-and-Testing-Data | Sprint 3 | WEEK 3

sajadsoorni:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/implement/get-angle-type.js
+++ b/Sprint-3/implement/get-angle-type.js
@@ -29,17 +29,20 @@
 function getAngleType(angleDegree) {
   if (angleDegree === 90) {
     return "Right angle";
-  } else if (angleDegree < 90) {
-    return "Acute angle";
-  } else if (angleDegree > 90 && angleDegree < 180) {
-    return "Obtuse angle";
-  } else if (angleDegree === 180) {
-    return "Straight angle";
-  } else if (angleDegree > 180 && angleDegree < 360) {
-    return "Reflex angle";
-  } else {
-    return "Invalid angle";
   }
+  if (angleDegree < 90) {
+    return "Acute angle";
+  }
+  if (angleDegree > 90 && angleDegree < 180) {
+    return "Obtuse angle";
+  }
+  if (angleDegree === 180) {
+    return "Straight angle";
+  }
+  if (angleDegree > 180 && angleDegree < 360) {
+    return "Reflex angle";
+  }
+  return "Invalid angle";
 }
 
 console.log(getAngleType(361));
--- a/Sprint-3/implement/get-angle-type.js
+++ b/Sprint-3/implement/get-angle-type.js
@@ -25,3 +25,21 @@
 // Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
+
+function getAngleType(angleDegree) {
+  if (angleDegree === 90) {
+    return "Right angle";
+  } else if (angleDegree < 90) {
+    return "Acute angle";
+  } else if (angleDegree > 90 && angleDegree < 180) {
+    return "Obtuse angle";
+  } else if (angleDegree === 180) {
+    return "Straight angle";
+  } else if (angleDegree > 180 && angleDegree < 360) {
+    return "Reflex angle";
+  } else {
+    return "Invalid angle";
+  }
+}
+
+console.log(getAngleType(361));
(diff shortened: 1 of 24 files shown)


Sprint-3/implement/get-angle-type.js
```diff
@@ -25,3 +25,21 @@
 // Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
+
+function getAngleType(angleDegree) {
```
SallyMcGrath, line 29: Nice! Take a moment to refactor this. Could you make this simpler? What would happen if you just made the if statements and didn't bother with the else?

Sprint-3/implement/rotate-char.js
```diff
@@ -41,3 +41,46 @@ console.log(rotateCharacter("7", 5)); // Output: "7" (unchanged, not a letter)
 // And the function should return the rotated character as a string (e.g., 'z' rotated by 3 should become 'c', 'Z' rotated by 3 should become 'C').
 console.log(rotateCharacter("z", 1)); // Output: "a" (preserves case, but wraps around)
 console.log(rotateCharacter("Y", 2)); // Output: "A" (preserves case, but wraps around)
+
+function rotateCharacter(char, n) {
+  if (!/^[a-zA-Z]$/.test(char)) return "Unchanged, not a letter";
```
SallyMcGrath, line 46: I think you're adding in quite a few extra conditions here. Could you simplify this code a bit? How would you go about doing that?

Sprint-3/revise/implement/count.test.js
```diff
@@ -15,3 +15,96 @@
 // And a character char that does not exist within the case-sensitive str,
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of the char were found in the case-sensitive str.
+//---------------------------------------------------------------------------------------------------
```
SallyMcGrath, line 18: 😵 This is so much code! I definitely think it would make sense to stop and check before writing epics like this.

Go test first, writing the simplest test you can think of, that satisfies the simplest condition defined in the criteria. Just write one test and then just enough code to pass that test. 

Spend the time thinking of the simplest thing that can possibly work.

Sprint-3/revise/implement/get-ordinal-number.test.js
```diff
@@ -2,3 +2,62 @@
...
+describe("getOrdinalNumber", () => {
+  test("returns 1st for 1", () => {
+    expect(getOrdinalNumber(1)).toBe("1st");
+  });
+
+  test("returns 2nd for 2", () => {
+    expect(getOrdinalNumber(2)).toBe("2nd");
+  });
+
+  test("returns 3rd for 3", () => {
```
SallyMcGrath, line 16: Can you think of a description here that adds MORE information instead of just restating the assertion?

What is the underlying pattern you are looking for? What do these tests really mean?

SallyMcGrath (changes requested):
Some lovely work here, @sajadsoorni . Thanks for this.

I think the key thing for you to work on is simplicity and refactoring. You have written absolutely pages of code. Is this necessary? Is this desirable? Is this the simplest thing that can possibly work? 

Take a look at some of the sample solutions and revise some of your programs. The sample solutions are not perfect or anything, but are often quite a bit simpler. Simplicity, clarity, and correctness are your goals here. Good luck!

Sprint-3/implement/get-angle-type.js
sajadsoorni: Thank you for the suggestion
I've refactored the code as per your advice and pushed the changes. Let me know if there's anything else to improve.

Sprint-3/implement/rotate-char.js
sajadsoorni: Thank you for the feedback.
I’ve simplified the code by reducing the number of conditions and using regular expressions to check if the character is a letter (uppercase or lowercase).

Sprint-3/revise/implement/get-ordinal-number.test.js
sajadsoorni: I've updated the test descriptions to be more descriptive and explain the underlying patterns I'm testing for. Each test now describes the expected ordinal number and includes details about edge cases like 11th, 12th, and 13th.

Sprint-3/revise/implement/count.test.js
sajadsoorni: I've simplified the code and focused on writing smaller, simpler tests first, as suggested. I'll continue using this approach going forward. 

Thank you @SallyMcGrath  for your valuable feedback.

Sprint-3/revise/implement/count.test.js
```diff
@@ -15,3 +15,96 @@
...
+
+//// Converts a number into its English word representation
+function numberToWords(num) {
```
SallyMcGrath: I think you can safely delete all of this

sajadsoorni: Oops, you're right! 🙈 I've deleted the lines as suggested

PR #146, opened 2024-11-15
Sprint 2

Della-Bella:
LONDON | GISLAINE_DELLA_BELLA | MODULE-STRUCTURING-AND-TESTING-DATA | SPRINT-2

Self checklist

- [X ] I have committed my files one by one, on purpose, and for a reason
- [X ] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [X ] I have tested my changes
- [X ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [X ] My changes meet the [requirements](./README.md) of this task

## Questions

Hello Volunteer, good morning
Here are my exercises for review.
Thank you for taking the time to review them. I appreciate it,
Gislaine


Sprint-2/debug/0.js
```diff
@@ -1,7 +1,22 @@
...
+the first console.log (a * b) which will call the function inputs 10, 32 = 320
 function multiply(a, b) {
   console.log(a * b);
 }
 
 console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
+
+DEBUG*/
+
+//The result of multiplying 10 and 32 is 320
```
SallyMcGrath, line 13: Did this prediction bear out, or did it print undefined? Why did it do this?

Sprint-2/debug/1.js
```diff
@@ -1,8 +1,21 @@
...
   return;
   a + b;
 }
 
+console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);*/
+
+//FIX://
+
+function sum(a, b) {
+  return; a + b;
```
SallyMcGrath, line 17: When you executed this function, did this fix work? 🤔

Sprint-2/debug/2.js
```diff
@@ -10,5 +10,41 @@ console.log(`The last digit of 42 is ${getLastDigit(42)}`);
 console.log(`The last digit of 105 is ${getLastDigit(105)}`);
 console.log(`The last digit of 806 is ${getLastDigit(806)}`);
 
-// This program should tell the user the last digit of each number.
-// Explain why getLastDigit is not working properly - correct the problem
+/* This program should tell the user the last digit of each number.
+ the function doesn’t have a parameter to receive the num variable it’s trying to process.
```
SallyMcGrath, line 14: This is true, but not complete. It doesn't have a parameter, so why is it returning 3? Where does the 3 come from?

Sprint-2/errors/1.js
```diff
@@ -3,11 +3,14 @@
 // Why will an error occur when this program runs?
 // Try playing computer with the example to work out what is going on
```
SallyMcGrath, line 4: What do you think? Why is this happening? What is the fix?

Sprint-2/errors/2.js
```diff
@@ -1,10 +1,14 @@
...
 
-function square(3) {
-    return num * num;
-}
+ }
+ console.log (square (3))
 
 
+/* this function should square any number but instead we're going to get an error
+ Fix 3 is not valid parameter name. should change function name to square(num)
```
SallyMcGrath, line 12: What is the error, specifically?

Sprint-2/implement/bmi.js
```diff
@@ -13,3 +13,24 @@
 // Given someone's weight in kg and height in metres
 // Then when we call this function with the weight and height
 // It should return their Body Mass Index to 1 decimal place
+
+
+function ibmCalculator() {
```
SallyMcGrath, line 18: What would happen if you had a different height or weight?

Sprint-2/myexc2.js
```diff
@@ -0,0 +1,59 @@
...
+
+const currentOutput = formatAs12HourClock("09:00");
+const targetOutput = "09:00 am";
+
+console.assert(
+  currentOutput === targetOutput,
+  `current output: ${currentOutput}, target output: ${targetOutput}`
+);*/
+
+function formatAs12HourClock(time) {
```
SallyMcGrath, line 23: What would happen if the time was in the afternoon?

SallyMcGrath (commented):
Thanks for your hard work here, @Della-Bella .

I think it would be a good idea to spend a bit more time on this. There are some foundational concepts that you need to work on here. Book some time in with me or with another volunteer in the pair programming channel and we can walk through this coursework step by step. 

You can do this. You will get there.

Sprint-2/debug/0.js
Della-Bella: The function wasn't called initially, so it didn't display the result. However, it has now been called and is returning the result.

Sprint-2/debug/1.js
Della-Bella: This fix didn't work because the sum of "a" and "b" wasn't included. Now I have fixed the code, and it is returning the result.

Sprint-2/errors/1.js
Della-Bella: Remove the duplicated declaration. With [this commit ](https://github.com/CodeYourFuture/Module-Structuring-and-Testing-Data/pull/146/commits/17d1618436ef72fbdd53e66334675b2dfb7fe2ce)

Sprint-2/errors/2.js
Della-Bella: The error was with parameter 3, which always gave me the same result. By changing it to "num," the function now works with any number. Also 3 is used as the parameter name for the function, which is not valid in JavaScript.

Sprint-2/implement/bmi.js
Della-Bella: I understand now why it was giving the same results. I removed the constant from inside the function and fixed the height units.

Della-Bella: Fix the units/ removed the constant values

Sprint-2/myexc2.js
Della-Bella: It wouldn't work. Fix the function to the 12 hours format.

PR #151, opened 2024-11-15
CYF London 11 | Nadika Zavodovska | Structuring-and-Testing-Data | Week 3

nadika-zavodovska:
Self checklist

## Changelist

I have completed the Sprint 3 exercises.

Thank you.


Sprint-3/implement/get-angle-type.test.js
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// console.assert tests
+// Test 1
+const currentOutput1 = getAngleType(90);
+const targetOutput1 = "Right angle";
+console.log(currentOutput1);
+console.assert(currentOutput1 === targetOutput1, `${currentOutput1} is not equal ${targetOutput1}`);
+
+// Test 2
```
SallyMcGrath, line 51: Instead of writing Test 2, what else could you say about this test?

Sprint-3/implement/get-angle-type.test.js
```diff
@@ -0,0 +1,88 @@
...
+// Then the function should return "Straight angle"
+
+// Identify Reflex Angles:
+// When the angle is greater than 180 degrees and less than 360 degrees,
+// Then the function should return "Reflex angle"
+
+function getAngleType(angle) {
+    if (angle === 90) {
+        return "Right angle";
+    } else if (angle < 90) {
```
SallyMcGrath, line 32: Nice! 

If you were to revise this function, how could you make it simpler and easier to read?

Sprint-3/implement/is-proper-fraction.test.js
```diff
@@ -0,0 +1,88 @@
...
+// target output: true
+// Explanation: The fraction -4/7 is a proper fraction because the absolute value of the numerator (4) is less than the denominator (7). The function should return true.
+
+// Equal Numerator and Denominator check:
+// Input: numerator = 3, denominator = 3
+// target output: false
+// Explanation: The fraction 3/3 is not a proper fraction because the numerator is equal to the denominator. The function should return false.
+// These acceptance criteria cover a range of scenarios to ensure that the isProperFraction function handles both proper and improper fractions correctly and handles potential errors such as a zero denominator.
+
+function isProperFraction(numerator, denominator){
```
SallyMcGrath, line 36: Typically, we would write the function in the file is-proper-fraction.js and the test in the paired filed is-proper-fraction.test.js

Then we would import the function into the test file to make it available to the tests (just like in html we import the css file to make it work on the html).

I'm not sure this was completely clear in the coursework; what do you think? How could we revise it to make it clearer?

Sprint-3/implement/is-valid-triangle.test.js
```diff
@@ -33,3 +33,48 @@
...
+        return true;
+    }
+}
+
+let triangle = isValidTriangle(-2, 4, 5);
+console.log(triangle);
+
+// console.assert tests
+// Test 1
+const currentOutput1 = isValidTriangle(-2, 4, 5);
```
SallyMcGrath, line 52: To make this better, bring some more semantic depth into your descriptions, comments, and variable names.

When we write code, we are acting like a translator between the computer and other humans. We have to always think about it like communication: what are we telling not just the computer, but other humans reading our conversation.

Sprint-3/implement/rotate-char.test.js
```diff
@@ -41,3 +41,56 @@ console.log(rotateCharacter("7", 5)); // Output: "7" (unchanged, not a letter)
 // And the function should return the rotated character as a string (e.g., 'z' rotated by 3 should become 'c', 'Z' rotated by 3 should become 'C').
 console.log(rotateCharacter("z", 1)); // Output: "a" (preserves case, but wraps around)
 console.log(rotateCharacter("Y", 2)); // Output: "A" (preserves case, but wraps around)
+
+function rotateCharacter(char, shiftNum) {
```
SallyMcGrath, line 45: There's a lot of repetition in here. I wonder, could you extract a function to reduce some of the repetition? Put a note in your calendar to come back in a couple of weeks and refactor this with a helper function.

Sprint-3/revise/implement/password-validator.test.js
```diff
@@ -14,3 +14,37 @@ To be valid, a password must:
...
+        return false;
+    }
+    return true;
+}
+
+let userPassword = checkPassword("te^A");
+// console.log(userPassword);
+
+// jest tests 
+test("check password", () => {
```
SallyMcGrath, line 40: Again, really want some explanations in your tests. Imagine your colleague doesn't know the password rules, but they've come to this program to find out why a failing test is taking down your application. How many steps would it take them to find out that 

`expect(checkPassword("te^")).toBe(false);` is failing BECAUSE it's too short. 

They shouldn't have to read outside the test to figure this out.

SallyMcGrath (commented):
Lovely work here @nadika-zavodovska . Thanks so much for your time on this.

Overall my main advice is to work more on organisation and communication. That doesn't mean you need to write gobs of comments every time, but spending time on 

- clear simple variable names
- coherent grouping and ordering of your operations and assertions
- clear descriptions in your tests, like a readable sentence that explains not just what but why you are testing

Will really bring you up to the next level. Good luck! Keep going! ⭐

nadika-zavodovska:
Thanks so much for the review and really helpful feedback, @SallyMcGrath  😊 I truly appreciate you taking the time to look over my work.

You’re right, it's a great practice to explain in comments why we need a particular test, especially for someone who might not be familiar with the context. I'll work on improving how I organise and explain things in my code. I'll also focus on using clearer, more meaningful variable names and grouping things in a way that’s easier to follow.

Your advice is super helpful, and I'm happy to implement it. Thanks again for reviewing my work 😌

Sprint-3/implement/rotate-char.test.js
nadika-zavodovska: Thank you @SallyMcGrath I put a note in my calendar. That is a great idea to come back in a few weeks

Sprint-3/implement/is-proper-fraction.test.js
nadika-zavodovska: Yes, I started to create separate files for tests and functions from the next sprint after this. It's really more clear.

Sprint-3/implement/get-angle-type.test.js
nadika-zavodovska: I can add to the description "Test to check if the input angle is an acute angle".

Sprint-3/implement/get-angle-type.test.js
nadika-zavodovska: Probably this version of the function is easier to read.
Thank you.

    if (angle < 90) return "Acute angle";
    if (angle === 180) return "Straight angle";
    if (angle < 180) return "Obtuse angle";
    if (angle < 360) return "Reflex angle";
    return "Invalid angle";
}

PR #164, opened 2024-11-18
CYF-ITP-South Africa | Natalie Patel | Module-Structuring-and-Testing-Data | Sprint 2

NataliePatel:
Self checklist

## Changelist

I have completed all the exercises for Sprint 2.

## Questions

### DIFF
--- a/Sprint-2/errors/0.js
+++ b/Sprint-2/errors/0.js
@@ -7,3 +7,8 @@ function capitalise(str) {
   let str = `${str[0].toUpperCase()}${str.slice(1)}`;
   return str;
 }
+
+/* In line 7, there is a declaration of the variable str but it is also in the scope of itself.
+Either a new variable must be declared or the existing variable should be modified by removing
+let.
+*/
--- a/Sprint-2/extend/format-time.js
+++ b/Sprint-2/extend/format-time.js
@@ -3,8 +3,9 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
+
   if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+    return `${hours - 12} pm`; // I've removed the :00 in the return for minutes to be output and asserted
   }
   return `${time} am`;
 }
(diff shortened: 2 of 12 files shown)


Sprint-2/extend/format-time.js
```diff
@@ -22,3 +23,26 @@ console.assert(
   currentOutput2 === targetOutput2,
   `current output: ${currentOutput2}, target output: ${targetOutput2}`
 );
+ 
+// My code starts here
+
+const currentOutput3 = formatAs12HourClock("14:00");
```
SallyMcGrath, line 29: Great! What would happen if the user sent "28:17"?

Sprint-2/implement/bmi.js
```diff
@@ -13,3 +13,14 @@
 // Given someone's weight in kg and height in metres
 // Then when we call this function with the weight and height
 // It should return their Body Mass Index to 1 decimal place
+
+// My code starts here 
+
+function BMIcalc(w, h) {
```
SallyMcGrath, line 19: Typically, we write function names like this: verbNoun, or what it _does_ and what it does it to. 

computeValue , convertName, getProperty. What might you name this?

SallyMcGrath (commented):
Thanks for your hard work here @NataliePatel . I've left just a few comments about working on making your code and explanations clearer and more complete. Have a look and a think.

Sprint-2/errors/0.js
```diff
@@ -7,3 +7,8 @@ function capitalise(str) {
   let str = `${str[0].toUpperCase()}${str.slice(1)}`;
   return str;
 }
+
+/* In line 7, there is a declaration of the variable str but it is also in the scope of itself.
```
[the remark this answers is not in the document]
NataliePatel: We could remove the "let" because str is already an existing parameter in the function, it does not need to be redeclared. When we remove "let" it would then mean a modification of the existing variable to allow it to be returned when calling "return str"

Sprint-2/extend/format-time.js
NataliePatel: Ha ha I did not write code to throw an error message for the edge case where users input a time that is not possible. A 24hr clock can only go up to to 23:59 so if they input hours beyond 23 or minutes beyond 59, there should be a message to them with feedback and instructions.

Out of curiosity, should I write code that actually prevents the function from running or is an error message sufficient? I'm thinking about instances where that input data might be stored in a database and I should disallow that.

Sprint-2/implement/bmi.js
NataliePatel: Thank you, I will in future write a function name like "calculateBMI" in future :-)
I can see why reading it the way I initially wrote it, if I continued a pattern like that, would be difficult to understand.

Sprint-2/extend/format-time.js
SallyMcGrath: Think back to the forms project in the first module. We write validations on the inputs and reject values that don't match the pattern. So long as there is a user, it's better to communicate with them in this way: throw an error and pass a message explaining what to do. 

Sometimes there isn't a user - there's just messy data you have to reconcile. In this case you might choose to clean and transform the data yourself.

PR #183, opened 2024-11-22
london|saraamiri|module-structuring-and-testing-data|sprint3|week3

saraamiiri:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions


Sprint-3/implement/get-angle-type.test.js
```diff
@@ -0,0 +1,31 @@
...
+  it('should return "Straight angle" for angle equal to 180', () => {
+    expect(getAngleType(180)).toBe('Straight angle');
+  });
+
+  it('should return "Reflex angle" for angles between 180 and 360', () => {
+    expect(getAngleType(270)).toBe('Reflex angle');
+  });
+
+  it('should return "Invalid angle" for angles less than 0 or greater than 360', () => {
+    expect(getAngleType(400)).toBe('Invalid angle');
```
mhmdk, line 28: why delete the other invalid angle test case?

Sprint-3/implement/is-valid-triangle.js
```diff
@@ -33,3 +33,43 @@
...
+    if (a + b > c && a + c > b && b + c > a) {
+        return true;
+    } else {
+        return false;
+    }
+}
+module.exports =  {isValidTriangle} ;
+// Test Cases
+function runTests() {
+    console.log("Test 1: Valid Triangle (a = 3, b = 3, c = 3)");
```
mhmdk, line 52: can you please move these to the test file, so that the expected value is automatically checked?

Sprint-3/implement/rotate-char.js
```diff
@@ -41,3 +41,35 @@ console.log(rotateCharacter("7", 5)); // Output: "7" (unchanged, not a letter)
...
 console.log(rotateCharacter("z", 1)); // Output: "a" (preserves case, but wraps around)
 console.log(rotateCharacter("Y", 2)); // Output: "A" (preserves case, but wraps around)
+
+
+function rotateCharacter(char, shift) {
+    // Check if the character is a lowercase letter
+    if (char >= 'a' && char <= 'z') {
+        let start = 'a'.charCodeAt(0); // ASCII code for 'a'
+        let offset = (char.charCodeAt(0) - start + shift) % 26;
+        return String.fromCharCode(start + (offset >= 0 ? offset : offset + 26));
```
mhmdk, line 51: could you explain why is the offset>=0? check required? what test cases would fail without it?

Sprint-3/implement/rotate-char.test.js
```diff
@@ -0,0 +1,49 @@
+// rotateCharacter.test.js
+const rotateCharacter = require('./rotate-char.js'); // Import the function
+
+describe('rotateCharacter', () => {
+    // Test for lowercase letters
+    test('should rotate lowercase letter "a" by 3 positions', () => {
```
mhmdk, line 6: There are other interesting cases that should be tested. In particular, a wider shift value range.

Sprint-3/revise/implement/count.test.js
```diff
@@ -15,3 +15,29 @@
...
+  });
+
+  // Scenario 3: Single Occurrence
+  test('counts a single occurrence', () => {
+    expect(countChar("hello world", "o")).toBe(2);  // 'o' appears 2 times
+  });
+
+  // Scenario 4: Empty String
+  test('returns 0 for empty string', () => {
+    expect(countChar("", "a")).toBe(0);  // No characters to count
```
mhmdk, line 40: good catch!

Sprint-3/revise/implement/count.test.js
```diff
@@ -15,3 +15,29 @@
 // And a character char that does not exist within the case-sensitive str,
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of the char were found in the case-sensitive str.
```
mhmdk, line 17: note the case-sensitive part, so the tests should cover a mix of capital and small letters

mhmdk:
All in all, good work!

Sprint-3/implement/rotate-char.js
saraamiiri: thak you for reviewing ,Without the offset >= 0 check, rotating a character with a negative shift would give a negative offset.

By adding offset >= 0 ? offset : offset + 26, we ensure that the offset always falls within the valid range of 0 to 25

Sprint-3/implement/is-valid-triangle.js
saraamiiri: thank you for reviewing

mhmdk: you welcome!
 I think you just removed these cases from is-valid-triangle.js, but forgot to add them to s-valid-triangle.test.js

Sprint-3/implement/rotate-char.js
mhmdk: got it. The specs mention negative shifts only for non letter input, but this is solid as well.

PR #193, opened 2024-11-24
ITP-2 Sprint02 | Stanislav Starishko | Module-Structuring-and-Testing-Data | Week 2

StanStarishko:
## Learners, PR Template

Self checklist

- [✅] I have committed my files one by one, on purpose, and for a reason
- [✅] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [✅] I have tested my changes
- [✅] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [✅] My changes meet the [requirements](./README.md) of this task

## I have completed the exercises for **Sprint-2**
in the folders:
- **debug**
- **errors**
- **extend**
- **implement**
- **interpret**

### DIFF
--- a/Sprint-2/debug/0.js
+++ b/Sprint-2/debug/0.js
@@ -1,7 +1,22 @@
 // Predict and explain first...
 
 function multiply(a, b) {
-  console.log(a * b);
+  //console.log(a * b);
+  return a * b;
 }
 
 console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
+
+/* 
+Frankly speaking, I did this task in reverse order. 
+I first launched it, and then reviewed the code.
+If we imagine that I first reviewed it to predict, 
+then at first glance everything should work correctly.
+
+But after launching, we see that the result of calling 
+the multiply function is not displayed in the console. 
+This means that I did not review it carefully. 
+Our function does not return any value, but outputs 
+information to the console, so we get its value undefined. 
+We need to fix the code so that the function returns a value.
+ */
--- a/Sprint-2/debug/1.js
+++ b/Sprint-2/debug/1.js
@@ -1,8 +1,27 @@
 // Predict and explain first...
 
 function sum(a, b) {
-  return;
-  a + b;
+  //return;
+  //a + b;
+  return a + b;
 }
 
 console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
+
+
+/* 
+My prediction:
+We should initially get the result as in previous exercise, 
+i.e. the value of the sum function should be undefined, 
+because in line 4 of our script the return operator does 
+not return anything, i.e. an undefined value.
+Line 5 of our script will be ignored, because 
+in the previous line there is a return from the function.
+In my opinion, there should be no system crash.
+
+Let's check if my theory is correct,  and just fix the code. 
+
+
+Test results:
+As expected, testing confirmed my theory, so I just fix the code.
+*/
--- a/Sprint-2/errors/0.js
+++ b/Sprint-2/errors/0.js
@@ -4,6 +4,64 @@
 // interpret the error message and figure out why an error is occurring
 
 function capitalise(str) {
-  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
-  return str;
+  let strOut = `${str[0].toUpperCase()}${str.slice(1)}`;
+  return strOut;
 }
+
+console.log(`Test with normal data: "hello guys" is ${capitalise("hello guys")}`);
+console.log(`Test with abnormal data: "h" is ${capitalise("h")}`);
+//console.log(`Test with abnormal data: "" is ${capitalise("")}`);
+//console.log(`Test with extreme data: 234 is ${capitalise(234)}`);
+console.log(`Test with normal data: "hello guys again" is ${capitalise("hello guys again")}`);
+
+/* 
+My prediction:
+I will give the prediction in the order of theories, 
+i.e. what I noticed first, etc. I will only analyze line 7:
+let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+because there is nothing else to analyze
+
+1. Judging by the code, we should accept the incoming string 
+and return the same string, but converting the first character 
+of the string to uppercase.
+
+2. The first potential error is if the incoming string is empty 
+or one character long, i.e. the minimum length of the incoming string 
+should be at least 2, otherwise it will cause a critical error. 
+Maybe not critical, I don't know node.JS behavior that well yet
+
+3. We should have an error because we redefine the "let str" 
+variable with the same identifier as the parameter of our function. 
+If there is no error in this place, then the JS interpreter 
+or compiler will surprise me greatly.
+
+To confirm my theories I will modernize the code, but not much. 
+To write reliable code I would check the length of the string 
+and if it is empty, then return an empty string, if the length 
+of the string is 1, then return only the first character 
+in uppercase. To save time I will make cosmetic changes 
+to confirm my theories and detect all errors if possible, 
+if there are others. If the input parameter is a number 
+consisting of 2 digits or more, this should not cause errors.
+ 
+
+Test results:
+1. My theory about the string length is partially correct. 
+When the string is one character long or an empty string, 
+the .slice() method returns an empty string. Therefore, 
+it was confirmed only on an empty string and this caused 
+a critical error. But this happened because str[0] did not work
+
+2. My theory is that if there is another data type instead 
+of a string, I chose a number to test the theory, then an implicit 
+type conversion will occur. The theory was not confirmed, 
+because I was inattentive str[0] in this case does not perform 
+an implicit conversion, so my theory failed and its testing 
+caused a critical system error.
+
+Conclusions: for more stable operation of the function, 
+it is necessary to convert the input parameter to a string, 
+but this will not give a magic pill. If you specify more complex 
+types, unpredictable behavior is also possible, which 
+we will not check and fix now.
+*/
--- a/Sprint-2/errors/1.js
+++ b/Sprint-2/errors/1.js
@@ -4,10 +4,62 @@
 // Try playing computer with the example to work out what is going on
 
 function convertToPercentage(decimalNumber) {
-  const decimalNumber = 0.5;
+  //const decimalNumber = 0.5;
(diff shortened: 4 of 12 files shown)


Sprint-2/debug/0.js
```diff
@@ -1,7 +1,22 @@
...
 function multiply(a, b) {
-  console.log(a * b);
+  //console.log(a * b);
+  return a * b;
 }
 
 console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
+
+/* 
+Frankly speaking, I did this task in reverse order.
```
SallyMcGrath, line 11: Yes, lots of people do this. We ask you to predict-explain because there's reasonable evidence it increases retention, understanding, and test scores. 

https://files.eric.ed.gov/fulltext/EJ1341654.pdf

Sprint-2/debug/1.js
```diff
@@ -1,8 +1,27 @@
...
 console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
+
+
+/* 
+My prediction:
+We should initially get the result as in previous exercise, 
+i.e. the value of the sum function should be undefined, 
+because in line 4 of our script the return operator does 
+not return anything, i.e. an undefined value.
+Line 5 of our script will be ignored, because
```
SallyMcGrath, line 18: Yes - what do we call code like this?

Sprint-2/errors/0.js
```diff
@@ -4,6 +4,64 @@
...
+of the string to uppercase.
+
+2. The first potential error is if the incoming string is empty 
+or one character long, i.e. the minimum length of the incoming string 
+should be at least 2, otherwise it will cause a critical error. 
+Maybe not critical, I don't know node.JS behavior that well yet
+
+3. We should have an error because we redefine the "let str" 
+variable with the same identifier as the parameter of our function. 
+If there is no error in this place, then the JS interpreter
```
SallyMcGrath, line 35: Hee! Love it

Sprint-2/errors/1.js
```diff
@@ -4,10 +4,62 @@
...
-console.log(decimalNumber);
+//let decimalNumber = 0.1;
+
+//console.log(convertToPercentage(decimalNumber));
+console.log(convertToPercentage(0.1));
+
+
+/* My prediction:
+
+There should be an error in line 13, because
```
SallyMcGrath, line 21: What kind of error?

Sprint-2/errors/1.js
```diff
@@ -4,10 +4,62 @@
...
+3. Declare decimalNumber as a global variable or constant 
+and in line 13 call the convertToPercentage(decimalNumber) function. 
+In many programming languages ​​that have implicit type casting, 
+you can set the same identifiers for the global and local contexts. 
+When compiling, the identifiers are converted to their internal 
+identifiers, thereby avoiding the identifier overriding error. 
+A good example, I'm interested in how this works in JS. 
+But I want to note that I am not a supporter of this practice, 
+I prefer to assign different identifiers so that my code is working 
+and does not depend on the current features of the compiler
```
SallyMcGrath, line 50: 100%

Sprint-2/errors/2.js
```diff
@@ -3,8 +3,35 @@
...
+
+
+/* My predictions:
+
+I hope everything is simple and obvious here. 
+Somehow, the exercises went from complex to simple, 
+usually the other way around.
+
+Here the error will be due to the fact that 
+the identifier of the square() function parameter
```
SallyMcGrath, line 23: Why do you think it cannot start with a number?

Sprint-2/extend/format-time.js
```diff
@@ -2,23 +2,108 @@
...
+// Check normal format when 12:01 and more in 24 hours format
+currentTestOutput = formatAs12HourClock("12:01");
+targetTestOutput = "12:01 pm";
+console.assert(
+  currentTestOutput === targetTestOutput,
+  `current output: ${currentTestOutput}, target output: ${targetTestOutput}`
+);
+
+// Check normal format when 24:00 in 24 hours format
+currentTestOutput = formatAs12HourClock("24:00");
```
SallyMcGrath, line 86: Is there a 24:00 in 24h format? And is it 12pm?

Sprint-2/extend/format-time.js
```diff
@@ -2,23 +2,108 @@
...
+// Check normal format with minutes
+let currentTestOutput = formatAs12HourClock("05:45");
+let targetTestOutput = "05:45 am";
+console.assert(
+  currentTestOutput === targetTestOutput,
+  `current output: ${currentTestOutput}, target output: ${targetTestOutput}`
+);
+
+// Check abnormal format when one digital of hours and minutes
+currentTestOutput = formatAs12HourClock("1:0");
```
SallyMcGrath, line 54: Should this assume 1am? I'm not sure it looks like a time at all.

Sprint-2/extend/format-time.js
```diff
@@ -2,23 +2,108 @@
...
 );
+
+// my testing block
+
+/* I hope I have checked all possible options 
+for the error to occur when converting time 
+from 24-hour format to 12-hour format
+*/
+
+// Check normal format with minutes
```
SallyMcGrath, line 45: Loving all these tests. Good work here!

SallyMcGrath (commented):
Thanks for this @StanStarishko . Your detailed explanations are high quality. I've dropped a few little comments and questions for you to think about. Onward!

Sprint-2/debug/1.js
StanStarishko: We call it dead code

Sprint-2/errors/0.js
StanStarishko: I won't hide it, I sometimes make such mistakes myself, so I found it here easily and quickly 😉

Sprint-2/errors/1.js
StanStarishko: Frankly speaking, I do not understand the essence of the question. I gave an exhaustive description of what kind of error it is:

There should be an error in line 13, because
the variable decimalNumber is not defined anywhere
and is not declared in the global context.
This variable is a parameter
of the function convertToPercentage(),
which apparently was intended to convert
a decimal number to a percentage. For example:
0.1 is 10%, 0.05 is 5%, and 0.0025 is 0.25%

Please clarify your question, I really do not understand what you are asking me about

Sprint-2/errors/2.js
StanStarishko: I don't think, I know and I wrote about it. 😊

The Identifier name cannot start with a number. This is a Rule or axiom of any programming language, this rule was thought up by the creators of the first programming language and this rule is followed by all popular languages ​​that were created and are created after that

Sprint-2/extend/format-time.js
StanStarishko: There are different interpretations of 24:00 in the 24:00 format, namely, it can be 24:00, it can be 00:00, and it can be 23:59 or 23:59:59. I consider such questions provocative, not having an exact and unambiguous answer. 😉

Sprint-2/extend/format-time.js
StanStarishko: In my 30 years of experience in the IT field, I have seen even worse things from users, so such troubleshooting is now just as a reflex for me. 😊🕵️

Sprint-2/extend/format-time.js
StanStarishko: Honestly, I hate them 😊
But without them, it is impossible to get a good and reliable product 😉

Sprint-2/debug/1.js
SallyMcGrath: Yep, dead or unreachable code.

Sprint-2/errors/1.js
SallyMcGrath: syntaxError, referenceError etc https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors

But to be honest I only asked because I always try to find SOMETHING to press, extend, or develop, no matter how complete the answer (because high performers should get the same support to go even further), and this was all I could think of! 😂

Sprint-2/errors/2.js
SallyMcGrath: So, another way to say this might be, is there a difference between this cannot be "of the data type number" and this cannot be "a string that begins with a character that represents a number" ? Would the following produce different or the same errors? (And why)

```
function num(3_set) { return true}
function num(3set) { return true}
function num(3) { return true}
```

Sprint-2/extend/format-time.js
SallyMcGrath: That's the more interesting debate: when should you fix the error for the user and when should you make the user fix it themselves.

Sprint-2/errors/2.js
StanStarishko: Sorry, maybe I'm not expressing myself clearly. English is not my native language. But I'll just try to repeat it again, I don't know any other answer. Identifiers cannot be specified starting with a number in any programming language I know. This is the rule for assigning names to identifiers and all your 3 lines:
```
function num(3_set) { return true}
function num(3set) { return true}
function num(3) { return true}
```
should give a syntax error, because all three identifiers start with a number. I wrote about this both in the comment to the task and in our discussion. To avoid errors, you need to assign other identifiers to the parameters of all your functions. For example, these:
```
function num(_3_set) { return true}
function num(_3set) { return true}
function num(_3) { return true}
```

Sprint-2/extend/format-time.js
StanStarishko: In the comment above, I gave a real experience of how it happens in my country.
But clearly understanding that in your country there is no 24:00, in order to answer your question and make the user behave well 😉, I need to clearly know how it happens in your country, i.e. clearly understand the task, otherwise we will continue to go in circles.
When I clearly understand the task and conditions, then I will be able to compose a pool of tests correctly and "configure" the program so that it helps the user behave well based on my "little" knowledge of User Experience design. 😊

Sprint-2/extend/format-time.js
StanStarishko: My opinion is that the program should be as usable as possible and understand the user from the "half-symbol". That is, in our case, understand that 1:0 can be a real case and it should turn into 01:00 without the user's participation. 😉

PR #197, opened 2024-11-25
disambiguate password validator requirement

mhmdk:
It was not clear whether all non-alphanumeric symbols were allowed or not.


illicitonion (approved):
Thanks!

PR #198, opened 2024-11-26
CYF-ITP-South Africa | Serna Malala| Module-Structuring-and-Testing-Data | Week 3

sernamalala:
## Learners, PR Template

Self checklist

- [y] I have committed my files one by one, on purpose, and for a reason
- [y] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [y] I have tested my changes
- [y] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [y] My changes meet the [requirements](./README.md) of this task

## Changelist

Briefly explain your PR.

**I have completed all the exercises, I have implemented functions as required according to the conditions given and I have also created test files for these functions by using Jest.**
## Questions

**Regarding commenting in the test function, how descriptive should they be?**


Sprint-3/revise/implement/card-validator.js
```diff
@@ -0,0 +1,55 @@
+function cardValidator(input) {
+    //in case of whitespace, we can remove it by using trim
+    input = input.trim();
+
+    //checks if all the characters are digits
+    if (!/^\d+$/.test(input)) return false;
+
+
+    let count = 0;
+    //counts the number of duplicates, it should be more than 1
```
cjyuan, line 10: You can also find out the same info from `duplicates.length`.

Sprint-3/revise/implement/card-validator.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+console.log(cardValidator("4444444444444444"));
+console.log(cardValidator("9999777788880000"));
+console.log(cardValidator("6666666666661666"));
+
+console.log(cardValidator("a92332119c011112"));
+console.log(cardValidator("1111111111111110"));
+console.log(cardValidator("6666666666666661"));
```
cjyuan, line 55: Why not check also card numbers that does not have exactly 16 digits?

Sprint-3/revise/implement/count.test.js
```diff
@@ -15,3 +15,40 @@
...
+test('This test will check if the function counts multiple occurrences of a number', () => {
+    expect(currentOutput4).toEqual(targetOutput4);
+});
+
+const currentOutput5 = countChar("", "1");
+const targetOutput5 = 0;
+
+test('This test will check if the function counts multiple occurrences of a number', () => {
+    expect(currentOutput5).toEqual(targetOutput5);
+});
```
cjyuan, line 54: Since the test at line 45 and the test at line 52 are doing the same types of test, you can consider combining them into the one test, but with two `expect(...)` statements.

Sprint-3/revise/implement/get-ordinal-number.js
```diff
@@ -0,0 +1,34 @@
+function getOrdinalNumber(num) {
+    //function will add the correct suffix to the appropriate number with certain conditions
+    let ordinal = "";
+    if (num > 0) {
+        let StringNum = num.toString();
```
cjyuan, line 5: Common practice is to begin a variable name with lowercase letter.

Sprint-3/revise/implement/get-ordinal-number.js
```diff
@@ -0,0 +1,34 @@
...
+        //Have to account for 11-19 which end with 'th'
+        if (num % 100 === 11 || num % 100 === 12 || num % 100 === 13) {
+            ordinal = "th";
+        }
+        else if (Number(StringNum[StringNum.length - 1]) === 1 || Number(num) === 1) {
```
cjyuan, line 10: Why do you need to specifically check for `Number(num) === 1`?

Sprint-3/revise/implement/is-prime.js
```diff
@@ -0,0 +1,36 @@
+function isPrime(num) {
+    //function will check whether the given number is a prime number according to the conditions given
+    let count = 0;
+    if (num >= 1) {
+
+        //divide the given number by the iterations if the number is divisible by (1 and itself ONLY) then it is prime and will have a count of 2
+        //Otherwise it is not a prime number
+        for (let i = 1; i <= num; i++) {
```
cjyuan, line 8: This approach works and your code is correct, but the approach is not very efficient. Can you try to improve its performance? Since you may not be familiar with the mathematical properties of prime numbers, just use Google or ChatGPT to learn a better approach (not JS implementation) to determine if a number is prime, and then try to implement the approach.

Sprint-3/revise/implement/password-validator.js
```diff
@@ -0,0 +1,44 @@
...
+
+function passwordValidator(input) {
+
+    // have to go learn about regex or validation methods for this exercise
+    //Cant use this: /[a-z]{1,}[A-Z]{1,}[0-9]{1,}[!#$%.*&]{1,}/
+    //The above is for a password that goes in that specific order
+
+    //const regexPattern = /[a-z]{1,}[A-Z]{1,}[0-9]{1,}[!#$%.*&]{1,}/;
+    //write a regex for each condition
+    const lowercaseRegex = /^(?=.*[a-z])/;
```
cjyuan, line 11: Why not just specify the regular expression as `/[a-z]/` instead of `/^(?=.*[a-z])/`?

Sprint-3/revise/implement/password-validator.test.js
```diff
@@ -14,3 +14,52 @@ To be valid, a password must:
...
+test('This test will check if the password is valid according to the criteria given', () => {
+    expect(currentOutput).toEqual(targetOutput);
+})
+
+// short password
+const currentOutput1 = passwordValidator("S#2");
+const targetOutput1 = false;
+
+test('This test will check if the password is invalid because it is too short', () => {
+    expect(currentOutput1).toEqual(targetOutput1);
```
cjyuan, line 32: "S#2" can also be detected as "invalid" because it lacks a lowercase letter. So with this test value, we cannot be certain if the function can really check if a password is too short.

cjyuan (commented):
I left comments and suggestions in the code.

cjyuan:
Regarding your question, "Regarding commenting in the test function, how descriptive should they be?"

Write it in such a way that
- You think another programmer (i.e., PR reviewer) would understand
- You would still understand what you wrote after revisiting your code 6 months from now.

I personally prefer concise description. Complete sentence is not needed.

Sprint-3/implement/get-angle-type.js
```diff
@@ -25,3 +25,31 @@
...
+    else if (angle < 90 && angle >= 0) {
+        result = "Acute angle";
+    }
+    else if (angle > 90 && angle < 180) {
+        result = "Obtuse angle";
+    }
+    else if (angle === 180) {
+        result = "Straight angle";
+    }
+    else if (angle > 180) {
```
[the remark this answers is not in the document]
sernamalala: I've fixed the reflex angle condition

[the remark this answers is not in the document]
sernamalala:
```
    else if (angle > 180 && angle < 360) {
        result = "Reflex angle";
    }
```

Sprint-3/implement/get-angle-type.test.js
```diff
@@ -0,0 +1,72 @@
...
+
+test('Checks for an reflex angle', () => {
+    expect(currentOutput7).toEqual(targetOutput7);
+});
+
+const currentOutput8 = getAngleType(260);
+const targetOutput8 = "Reflex angle";
+
+test('Checks for a reflex angle', () => {
+    expect(currentOutput8).toEqual(targetOutput8);
```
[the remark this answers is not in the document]
sernamalala: Yes, I definitely see how that could make my code less cluttered, I was just being cautious and I tried having things laid out by using variables instead. If it is a must then I can change it

Sprint-3/implement/get-card-value.js
```diff
@@ -29,3 +33,38 @@
...
+    }
+    return result;
+}
+
+console.log(getCardValue("5♠"));  // 5
+console.log(getCardValue("J♦"));  // 10
+console.log(getCardValue("A♣"));  // 11
+console.log(getCardValue("Q♠"));  // 10
+// console.log(getCardValue("Z♠"));  
+
```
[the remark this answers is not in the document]
sernamalala: I see that the previous program was allowing decimal numbers, I have adapted it by using regex to allow only the specified integers, the above calls you provided will all lead to errors.

Sprint-3/implement/is-proper-fraction.test.js
```diff
@@ -0,0 +1,19 @@
+const isProperFraction = require('./is-proper-fraction');
+
+const currentOutput = isProperFraction(2, 7);
```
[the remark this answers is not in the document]
sernamalala: I've added those test cases to the is-proper-fraction.test.js file

Sprint-3/revise/implement/get-ordinal-number.js
sernamalala: A mistake on my part, I've fixed it

Sprint-3/revise/implement/get-ordinal-number.js
sernamalala: I added it as a precaution, not realising that it was already being covered by the condition give. I'll remove that. Thanks for the heads-up. I must have written it out when thinking of the specific conditions to give the if-else statement

Sprint-3/revise/implement/password-validator.js
sernamalala: I tried specifying it that way as `/[a-z]/ ` but it only worked when certain characters were in a specific order, it could not be randomized and looking at the password validator I don't belive that the passwords requires certain characters to come after each other

Sprint-3/revise/implement/password-validator.test.js
sernamalala: The first requirement that would be flagged would be the length. Are you saying that I should make my descriptions more descriptive by saying it will be false because of its length, not having a lowercase letter etc.?

sernamalala:
Hi @cjyuan I've looked at your review comments and I've adjusted those that I could. I left some questions for the comments I found a bit more difficult to understand.

Sprint-3/revise/implement/password-validator.test.js
cjyuan: In practice, over time, someone could have modified your implementation or reimplement the same function. So the tests you prepared should need to be able to check every implementation of the same function thoroughly. That means when you prepare the tests, you should not assume anything abou how the function is implemented, but only what is expected when the parameters are of certain values. In fact, you should prepare the tests as if you were trying to break  someone's implementation.

Sprint-3/revise/implement/password-validator.js
cjyuan: What string value was you testing when the test gives you a different result between
`/[a-z]/.test(str)` and `/^(?=.*[a-z])/.test(str)` (where `str` holds a string value)?

You can also try using ChatGPT to help you understand what these expressions mean and find out their difference.

Sprint-3/revise/implement/is-prime.js
cjyuan: This is an opportunity to practice "reading code".
May I suggest this: Pick some numbers (e.g., 15, 17, 19, 21, 23) and trace the code to see how it works,
with or without using the Python visualizer, https://pythontutor.com/javascript.html#mode=edit

Have you tried asking GPT to explain the code for you?

There is also a simpler version that skips only even numbers. You can ask GPT to present you this simpler version and try to understand this simpler version first. It could help you understand the version you showed here.

Sprint-3/implement/get-angle-type.test.js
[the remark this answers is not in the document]
cjyuan: Not a must.

Sprint-3/revise/implement/password-validator.js
cjyuan: I am not sure how you will get notified if I just leave my comments here. So I will also tag you  @sernamalala.

Sprint-3/revise/implement/is-prime.js
sernamalala: Hi. I've implemented the simpler version in my code that skips even numbers. I also took the version above and put it into a visualiser. Made it a lot easier to understand.  I prefer the version that skips even numbers, it's just a bit more straightforward

Sprint-3/revise/implement/password-validator.js
sernamalala: Hi. Yes I see what you mean, my understanding of the regex and how it works was off. There is a difference but I can use this `/[a-z]/.test(str)` expression for my code.

Sprint-3/revise/implement/password-validator.test.js
sernamalala: Ah I see. This took me a while to understand but I assume you're saying that each test should specifically check for a single case where things can go wrong, not mixing two cases in one

Sprint-3/implement/get-angle-type.js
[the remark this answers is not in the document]
sernamalala: Hi @KudahShambare I am only seeing your review now. I already responded to CJ's review

PR #203, opened 2024-11-27
solutions for sprint 2 debug

SallyMcGrath:
add some mdn links

solutions for sprint 2 errors

extend and interpret tasks

also add exhortation, but this should go to main I think

### DIFF
--- a/Sprint-2/debug/0.js
+++ b/Sprint-2/debug/0.js
@@ -4,7 +4,7 @@
 // This function logs out the result of multiplying a and b to the console.
 // When invoked it will log out the result of multiplying 10 and 32: 320
 // BUT it has no return statement, so it will RETURN undefined.
-// So when the function is called in the log statement, it will not return anything, the console.log statement will read
+// So when the function is called in the log statement, it will not return anything, the console.log statements will read:
 // 320
 // "The result of multiplying 10 and 32 is undefined"
 
--- a/Sprint-2/debug/0.js
+++ b/Sprint-2/debug/0.js
@@ -1,7 +1,25 @@
 // Predict and explain first...
 
+// My prediction:
+// This function logs out the result of multiplying a and b to the console.
+// When invoked it will log out the result of multiplying 10 and 32: 320
+// BUT it has no return statement, so it will RETURN undefined.
+// So when the function is called in the log statement, it will not return anything, the console.log statement will read
+// 320
+// "The result of multiplying 10 and 32 is undefined"
+
 function multiply(a, b) {
   console.log(a * b);
 }
 
 console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
+
+// Actual:
+// I was correct in my prediction.
+
+// When I ran this code in node REPL, the output was:
+// 320
+// console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
+// When I executed this file in VSCode  -> % node 0.js, the output was the same:
+// 320 <== executed from the function multiply when I executed the script file
+// The result of multiplying 10 and 32 is undefined
--- a/Sprint-2/debug/1.js
+++ b/Sprint-2/debug/1.js
@@ -1,8 +1,20 @@
 // Predict and explain first...
 
-function sum(a, b) {
-  return;
-  a + b;
-}
+// My prediction:
+// This function has a return statement with no value after it, so it will return undefined.
+// The log will say "The sum of 10 and 32 is undefined"
 
 console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
+
+// Actual:
+// I was correct in my prediction.
+// When I ran this code in VSCcode terminal -> % node 1.js, the output was:
+// The sum of 10 and 32 is undefined
+// This is because https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions#return_value
+// There's no value after the keyword return, so the function returns undefined
+// The code stops executing at the return statement, so the a + b is never executed
+// I could fix this by removing the return statement and moving the a + b to the same line as the return statement
+
+// function sum(a, b) {
+//   return a + b;
+// }
--- a/Sprint-2/debug/1.js
+++ b/Sprint-2/debug/1.js
@@ -13,7 +13,7 @@ console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
 // This is because https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions#return_value
 // There's no value after the keyword return, so the function returns undefined
 // The code stops executing at the return statement, so the a + b is never executed
-// I could fix this by removing the return statement and moving the a + b to the same line as the return statement
+// I could fix this by moving the a + b to the same line as the return statement before the ;
 
 // function sum(a, b) {
 //   return a + b;
--- a/Sprint-2/errors/0.js
+++ b/Sprint-2/errors/0.js
@@ -1,9 +1,28 @@
 // Predict and explain first...
+// Prediction:
+// Given the function capitalise
+// When I call the function capitalise("hello")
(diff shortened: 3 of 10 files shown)


Sprint-2/debug/0.js
```diff
@@ -1,7 +1,25 @@
...
 }
 
 console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
+
+// Actual:
+// I was correct in my prediction.
+
+// When I ran this code in node REPL, the output was:
+// 320
+// console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
```
illicitonion, line 22: This doesn't look like repl output?

Sprint-2/debug/1.js
```diff
@@ -1,8 +1,20 @@
...
 console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
+
+// Actual:
+// I was correct in my prediction.
+// When I ran this code in VSCcode terminal -> % node 1.js, the output was:
+// The sum of 10 and 32 is undefined
+// This is because https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions#return_value
+// There's no value after the keyword return, so the function returns undefined
+// The code stops executing at the return statement, so the a + b is never executed
+// I could fix this by removing the return statement and moving the a + b to the same line as the return statement
```
illicitonion, line 16:
```suggestion
// I could fix this by moving the a + b to the same line as the return statement before the ;
```

Sprint-2/errors/0.js
```diff
@@ -1,9 +1,28 @@
...
+);
+
+// Explain:
+// I executed and got this -> % node 0.js
+// /Users/SallyMcGrath/CYF/ITP/Module-Structuring-and-Testing-Data/Sprint-2/errors/0.js:13
+//   let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+//       ^
+// There's a little caret pointing to the place I need to fix.
+// I fixed it by returning the value of the expression directly instead of redeclaring the variable str
+// I tested my code with the assertion and it passed
```
illicitonion, line 28:
```suggestion
// I tested my code with the assertion and it passed
// I could also have renamed the variable, if I really wanted to use a variable.
```

Sprint-2/errors/1.js
```diff
@@ -1,13 +1,37 @@
...
+
+function convertToPercentageFixed(decimalNumber) {
+  // keyword "function" starts a function declaration, convertToPercentage is the function identifier, decimalNumber is the parameter
+  const percentage = `${decimalNumber * 100}%`; // compute the percentage of the decimalNumber and assign it to the constant variable percentage
 
   return percentage;
 }
 
-console.log(decimalNumber);
+console.log(convertToPercentageFixed(0.5));
```
illicitonion, line 33:
```suggestion
// We also need to call the function in order to log its return value - decimalNumber only exists inside the function, we can't see it outside here - decimalNumber is not in scope.
console.log(convertToPercentageFixed(0.5));
```

Sprint-2/extend/format-time.js
```diff
@@ -1,24 +1,94 @@
 // This is the latest solution to the problem from the prep.
+// https://programming.codeyourfuture.io/structuring-data/sprints/2/prep/
 // Your task is to write tests for as many different groups of input data or edge cases as you can, and fix any bugs you find.
 
 function formatAs12HourClock(time) {
+  // check my time is valid first
+  validateTimeInput(time);
   const hours = Number(time.slice(0, 2));
+  // handle midnight
+  if (hours === 0) return "12:00 am";
```
illicitonion, line 10: Woah there, why did all the minutes become `00`?

Could a few tests with different hour and minute values have helped to detect this bug? :)

Sprint-2/extend/format-time.js
```diff
@@ -1,24 +1,94 @@
...
+  const [hours, minutes] = time.split(":").map(Number);
+
+  // 3. Check if we got valid numbers within range
+  if (
+    hours === undefined ||
+    minutes === undefined ||
+    hours > 23 ||
+    minutes > 59
+  ) {
+    throw new Error("Invalid time - use hours 00-23 and minutes 00-59");
```
illicitonion, line 37: Your error message suggests there are lower bounds on these values, but your if statement doesn't check for them :)

Sprint-2/interpret/time-format.js
```diff
@@ -20,12 +18,27 @@ function formatTimeDisplay(seconds) {
...
+
+// This is because the time format is "HH:MM:SS" and each number needs padding!
+
 // Call formatTimeDisplay with an input of 61, now answer the following:
 
-// b) What is the value assigned to num when pad is called for the first time?
+// b) What is the value assigned to num when pad is called for the first time? 0
 
-// c) What is the return value of pad is called for the first time?
+// c) What is the return value of pad when pad is called for the first time?  "00"
```
illicitonion, line 33:
```suggestion
// c) What is the return value of pad when pad is called for the first time? "00"
```

Sprint-2/interpret/time-format.js
```diff
@@ -20,12 +18,27 @@ function formatTimeDisplay(seconds) {
...
 
 // d) What is the value assigned to num when pad is called for the last time in this program?  Explain your answer
 
+// d) The value assigned to num when pad is called for the last time in this program is 1.
```
illicitonion, line 37:
```suggestion
// d) What is the value assigned to num when pad is called for the last time in this program?  Explain your answer
// The value assigned to num when pad is called for the last time in this program is 1.
```

Sprint-2/interpret/time-format.js
```diff
@@ -20,12 +18,27 @@ function formatTimeDisplay(seconds) {
...
+// This is because pad is called with the value of remainingSeconds, which is 1, or 61 % 60.
+// 61 seconds is 1 minute and 1 second in total.
+
 // e) What is the return value assigned to num when pad is called for the last time in this program?  Explain your answer
```
illicitonion, line 41: I don't love the wording of this question. How would you feel about: "What does the pad function return the last time it is called in this program?"?

SallyMcGrath: Happy to change the questions but probably not in the solutions! That would be annoying for the learner. 

I think overall this module needs slimming down a LOT - there's too much work here and people won't have time to think about what they are doing or really explore and enjoy the work. I ploughed through this in between meetings and emails and a broken down boiler and I got super bored and inattentive.

Sprint-2/extend/format-time.js
SallyMcGrath: You're meant to build up the tests in the next sprint is the issue really.

illicitonion: This bug specifically was what made me create this exercise 😂  I don't believe the 12-hour-clock exercise comes back in sprint 3?

Sprint-2/extend/format-time.js
illicitonion: What are you expecting to change next sprint? I don't think there's a "revise your answers" thing in sprint 3? (Or honestly that there's more relevant knowledge next sprint - sprint 3 really just teaches Jest...)

(Also, FWIW, I did not originally expect validation / throwing to be a part of this exercise, but I'm not unhappy with it! But I do think adding a `hours < 0 || minutes < 0` check to the `if` is reasonable with the knowledge trainees have at this point)

Sprint-2/interpret/time-format.js
illicitonion: Yeah - personally I'd be pretty happy to remove most of the "playing computer" exercises - I think they're quite long-winded, fiddly to assess, and it's a better exercise to do in person with a person interacting with you to say "you skipped a line, slow down, I know it's frustrating"...

(An agree about changing in the questions not the solutions!)

PR #204, opened 2024-11-27
CYF-ITP-South Africa | Rashaad Ebrahim | Module-Structuring-and-Testing-Data | Week 1

Rashaad-Ebrahim:
## Learners, PR Template

Self checklist

## Changelist

Sprint 1 exercises complete

## Questions

N/A

### DIFF
--- a/Sprint-1/exercises/count.js
+++ b/Sprint-1/exercises/count.js
@@ -4,3 +4,6 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+
+
+// Line 3 is reassigning the value of the variable "count", to the value of "count + 1"
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -8,6 +8,6 @@ const num = 56.5678;
 
 // Log your variables to the console to check your answers
 
-const wholeNumberPart = Math.floor(56.7891)
-const decimalPart = (34.5678 - Math.floor(34.5678)).toFixed(4)
-const roundedNum = Math.round(56.7891)
+const wholeNumberPart = Math.floor(num)
+const decimalPart = (num - Math.floor(num)).toFixed(4)
+const roundedNum = Math.round(num)
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -7,3 +7,7 @@ const num = 56.5678;
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
+
+const wholeNumberPart = Math.floor(56.7891)
+const decimalPart = (34.5678 - Math.floor(34.5678)).toFixed(4)
+const roundedNum = Math.round(56.7891)
--- a/Sprint-1/exercises/paths.js
+++ b/Sprint-1/exercises/paths.js
@@ -16,3 +16,7 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
+
+const dir = filePath.slice(0, filePath.lastIndexOf(base));
+
+const ext = filePath.slice(filePath.lastIndexOf("."));
--- a/Sprint-1/exercises/random.js
+++ b/Sprint-1/exercises/random.js
@@ -7,3 +7,14 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+/* 
+
+num represents a random number between 1 and 100.
+
+Math.random() returns a random floating-point number between 0 and 1.
+(maximum - minimum + 1) scales up the returned number to the range that you would like.
+Math.floor() rounds down the number that you receive.
++ minimum is added to account for Math.random() giving a number BETWEEN 0 and 1 but not including 1.
+
+*/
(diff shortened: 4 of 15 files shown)


Sprint-1/exercises/count.js
```diff
@@ -4,3 +4,6 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+
+
+// Line 3 is reassigning the value of the variable "count", to the value of "count + 1"
```
cjyuan, line 9: There is a shorter ***programming term*** to describe operations like `count = count + 1` or `count++`. May I suggest feeding the code to ChatGPT to see how else the code can be described? From time to time you may learn new terminology or more concise ways to describe code.

Sprint-1/exercises/decimal.js
```diff
@@ -7,3 +7,7 @@ const num = 56.5678;
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
+
+const wholeNumberPart = Math.floor(56.7891)
+const decimalPart = (34.5678 - Math.floor(34.5678)).toFixed(4)
```
cjyuan, line 12: It would be nicer if your code can be reused (without any change) to show the "whole number part", "decimal part", and "rounded value" of any number (not just `56.5678`).

Currently, if we change the initial value of `num` to `46.5678`, your code would not be able to produce the correct result.

Can you change your code (to take advantage of the value stored in variables) so that the code can work for any number assigned to `num`?

Sprint-1/exercises/paths.js
```diff
@@ -16,3 +16,7 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
+
+const dir = filePath.slice(0, filePath.lastIndexOf(base));
```
cjyuan, line 20: Have you tried executing this program and output the value of the variables using `console.log()` to see if the code can produce the result you expect?

I don't think the value assigned `dir` is correct.

Sprint-1/exercises/random.js
```diff
@@ -7,3 +7,14 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+/* 
+
+num represents a random number between 1 and 100.
+
+Math.random() returns a random floating-point number between 0 and 1.
```
cjyuan, line 15: The phrase "between 0 and 1" alone is not precise enough in program specification because it does not state clearly whether 0 and 1 are included in the range.
 
Does the random value returned by `Math.random()` include 0? Does the return value include 1?

After you find out the answer, may I suggest asking ChatGPT how to precisely describe a range of numbers from 0 to 1 that include/exclude 0/1.

Sprint-1/exercises/random.js
```diff
@@ -7,3 +7,14 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
...
+(maximum - minimum + 1) scales up the returned number to the range that you would like.
+Math.floor() rounds down the number that you receive.
++ minimum is added to account for Math.random() giving a number BETWEEN 0 and 1 but not including 1.
+
```
cjyuan, line 19: To test your understanding, how would you write an expression (without using any variable in the expression) that yields a random integer between -5 and 3 (including -5 and 3)?

cjyuan (commented):
I left my comments in the code.

Why not also try the exercises/questions in folder `errors`, `interpret`, `explore`?

Sprint-1/exercises/paths.js
cjyuan: Sorry. I overlooked. Code is correct

Sprint-1/exercises/count.js
Rashaad-Ebrahim: Yes, I am familiar with the ++ operator as well as the --.

Sprint-1/exercises/random.js
Rashaad-Ebrahim: Math.random() includes 0 but excludes 1.

Rashaad-Ebrahim:
I've updated my code as suggested in the PR.

Sprint-1/exercises/count.js
cjyuan, line 9: I was referring to the term "increment". We can describe `count = count +1` as "increment the value of count by 1".

Sprint-1/explore/chrome.md
```diff
@@ -9,10 +9,16 @@ Let's try an example.
...
+* It gives you an alert which will remain there until you acknowledge it by clicking ok.
+
 
 Now try invoking the function `prompt` with a string input of `"What is your name?"` - store the return value of your call to `prompt` in an variable called `myName`.
 
 What effect does calling the `prompt` function have?
+* * A popup apprears with the question, an input field, and an ok and cancel button.
+
 What is the return value of `prompt`?
+* The return value is undefined. However, the name I added to the input is saved in the variable myName
```
cjyuan, line 24: `prompt()` should never return `undefined`. If the user clicked "Cancel", `prompt()` would return `null` (which is not equal to `undefined` in JavaScript).

Sprint-1/explore/objects.md
```diff
@@ -5,12 +5,17 @@ In this activity, we'll explore some additional concepts that you'll encounter i
...
 
 Try also entering `typeof console`
+* It returns 'object'
 
 Answer the following questions:
 
 What does `console` store?
+* The console object contains methods (or functions)
 What does the syntax `console.log` or `console.assert` mean? In particular, what does the `.` mean?
+* The . calls upon the different methods available to the console object.
```
cjyuan, line 21: The description does not seem quite correct. Even if you know what `.` is, I still think you can ask ChatGPT to learn how else we can describe it.

cjyuan (commented):
Code is good after changes. I just left some suggestions.

PR #207, opened 2024-11-28
West Midlands | Gabriel Deng | Module-Structuring-and-Testing-Data | Sprint 3

gai93003:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/implement/get-card-value.js
+++ b/Sprint-3/implement/get-card-value.js
@@ -29,3 +29,26 @@
 // Given a card with an invalid rank (neither a number nor a recognized face card),
 // When the function is called with such a card,
 // Then it should throw an error indicating "Invalid card rank."
+
+
+const getCardValue = (card) => {
+  const cardValue = card.slice(0, -1);
+
+  if(/^\d+$/.test(cardValue)) {
+    if (Number(cardValue) >= 2 && Number(cardValue) <= 10) {
+      return Number(cardValue);
+    }
+  }
+  else if (/^[JQK]$/.test(cardValue)) {
+    return 10;
+  }
+  else if (cardValue === 'A') {
+    return 11;
+  }
+}
+
+console.assert(getCardValue('2♠') === 2, 'The given value of 2♠ returns 2');
+console.assert(getCardValue('A♠') === 11, 'The given value of "A" returns 11');
+console.assert(getCardValue('8♠') === 8, 'The given value of 8♠ returns 8');
+console.assert(getCardValue('J♠') === 10, 'The given value of J returns 10');
+console.assert(getCardValue('Q♠') === 10, 'The given value of 2♠ returns 10');
--- a/Sprint-3/implement/is-proper-fraction.js
+++ b/Sprint-3/implement/is-proper-fraction.js
@@ -32,3 +32,44 @@
 // target output: false
 // Explanation: The fraction 3/3 is not a proper fraction because the numerator is equal to the denominator. The function should return false.
 // These acceptance criteria cover a range of scenarios to ensure that the isProperFraction function handles both proper and improper fractions correctly and handles potential errors such as a zero denominator.
+
+function isProperFraction(numerator, denominator) {
+  if (denominator === 0) {
+    return false;
+  }
+  else if (numerator < 0 && denominator > 0) {
+    return true;
+  }
+  else if(numerator < denominator) {
+    return true;
+  }
+  else if (numerator > denominator) {
+    return false;
+  }
+  else if (numerator === denominator) {
+    return false;
+  }
+}
+
+console.assert(isProperFraction(3, 0) === false, `No target output: Error (Denominator cannot be zero), Explanation: The function should throw an error when the denominator is zero, as it's not a valid fraction.`);
+
+// console.assert(isProperFraction(-4, 7), `target output: false, Explanation: The fraction  is not a proper fraction because the numerator is equal to the denominator. The function should return false.`);
+
+console.assert(isProperFraction(-4, 7) === true, `target output: false, Explanation: The fraction  is a proper fraction because the absolute value of the numerator (4) is less than the denominator (7). The function should return true.`);
+
+console.assert(isProperFraction(5, 2) === false, `target output: false, Explanation: The fraction  is an improper fraction, where the numerator is greater than or equal to the denominator. The function should return false.`);
+
+console.assert(isProperFraction(2, 3) === true, `target output: true, Explanation: The fraction  is a proper fraction, where the numerator is less than the denominator. The function should return true.`);
+// isProperFraction(3, 5);
+
+
+// function formatAs12HourClock(time) {
+//   return `${time} am`;
+// }
+
+const currentOutput = isProperFraction(2, 4);
+const targetOutput = true;
+console.assert(
+  currentOutput === targetOutput,
+  `current output: ${currentOutput}, target output: ${targetOutput}`
+);
--- a/Sprint-3/implement/is-valid-triangle.js
+++ b/Sprint-3/implement/is-valid-triangle.js
@@ -33,3 +33,17 @@
 // Then it should return true because the input forms a valid triangle.
 
 // This specification outlines the behavior of the isValidTriangle function for different input scenarios, ensuring it properly checks for invalid side lengths and whether they form a valid triangle according to the Triangle Inequality Theorem.
+
+function isValidTriangle(a, b, c) {
+  if (a + b <= c ||  a + c <= b || b + c <= a) {
+    return false;
+  }
+  else if (a <= 0 || b <= 0 || c <= 0) {
+    return false;
+  }
+  else if (a + b > c || a + c > b || b + c > a) {
+    return true;
+  }
+}
+
+isValidTriangle(0, 4, 6)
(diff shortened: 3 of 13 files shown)


cjyuan:
You have modified files from Sprint-1 and Sprint-2 in your branch. Can you rebase your Sprint-3 branch to main so that the files in Sprint-1 and Sprint-2 remains the same as those in main?

cjyuan:
Did you run `git push --force origin Your_sprint_branch_name?` to force your local change onto remote repository after you rebased your sprint 3 branch? I can still see the changed files in sprint 1 and 2 folder in your sprint 3 branch. We can try to resolve this on Slack or over huddle if you can post a question or DM me on Slack.

gai93003:
Alright, I will DM you on slack.

Sprint-3/implement/rotate-char.js
```diff
@@ -41,3 +41,36 @@ console.log(rotateCharacter("7", 5)); // Output: "7" (unchanged, not a letter)
...
+        return lowerCaseLetters[j + num];
+      }
+      else if (char !== upperCaseLetters[j] || char !== lowerCaseLetters[i]) {
+        return char;
+      }
+    }
+  }
+}
+
+console.log(rotateCharacter("a", 3)); //Output: "d"
```
cjyuan, line 70: I wonder what your function will return when it is called with the following parameters:
```
console.log(rotateCharacter("x", 3)); 
console.log(rotateCharacter("Y", 1));  
```

You can try making your function to work on lowercase letters first, and then duplicate your code with tiny modification to apply them to uppercase letters. 

Also, try taking advantage of the `%` operator so that you don't to use any loop in your solution.

Sprint-3/revise/implement/count.test.js
```diff
@@ -15,3 +15,10 @@
 // And a character char that does not exist within the case-sensitive str,
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of the char were found in the case-sensitive str.
+
+const countChar = require('./count');
+
+test('description', () => {
+  expect(countChar('aaaabfdjaa', 'a')).toBe(6);
+  expect(countChar('bbbbbbbbb', 'a')).toBe(0);
+});
```
cjyuan, line 24: Have you tried running this test? It could help you detect a bug in your code.

Sprint-3/revise/implement/get-ordinal-number.js
```diff
@@ -0,0 +1,20 @@
...
+  }
+  else if (num % 10 === 3) {
+    return num + 'rd';
+  }
+  else {
+    return num + 'th';
+  }
+}
+
+module.exports = getOrdinalNumber;
```
cjyuan, line 20: No Jest test script?

Sprint-3/revise/implement/is-prime.js
```diff
@@ -0,0 +1,9 @@
+function isPrime(num) {
+  if (num <=1) return false;
+  for(let i = 2; i < num; i++) {
```
cjyuan, line 3: You can possibly improve the performance of the code in the following manners:
- Return `true` if `num` is 2
- Return `false` if `num` is an even number (we have already checked 2)

- Use a loop to check if `num` can be fully divided by an odd number >= 3 (but <= `Math.sqrt(num)`)
- In the loop, avoid calling `Math.sqrt(num)` repeatedly by first assigning the value of `Math.sqrt(num)` to a variable once, and then refer to the variable in the condition of the loop.

Sprint-3/revise/implement/password-validator.test.js
```diff
@@ -14,3 +14,9 @@ To be valid, a password must:
 
 You must breakdown this problem in order to solve it. Find one test case first and get that working
 */
+const isValidPass = require('./password-validator');
+
+test("Is prime number", function(){
```
cjyuan, line 19: Are we testing prime number ?

Sprint-3/revise/implement/password-validator.test.js
```diff
@@ -14,3 +14,9 @@ To be valid, a password must:
...
+  expect(isValidPass('Calohh35#')).toBe(true);
+  expect(isValidPass('alohh35#')).toBe(false);
+});
```
cjyuan, line 22: These tests does not check if the function can correctly return `false` for each specific type of invalid passwords.
For example, the test only check if the function can return false when a password does not contain any uppercase letter, but the test does not check if the function can return false when a password is missing a lowercase.

Can you include more test cases to see if the function can return false for each type of invalid passwords? They could possibly help you **detect a bug** in your code.

Sprint-3/revise/implement/repeat.js
```diff
@@ -0,0 +1,5 @@
+function repeat(str, count) {
+  return str.repeat(count);
```
cjyuan, line 2: According to the spec, what should happen if `count` is a negative number?

Sprint-3/implement/get-card-value.js
```diff
@@ -29,3 +29,26 @@
...
+  else if (cardValue === 'A') {
+    return 11;
+  }
+}
+
+console.assert(getCardValue('2♠') === 2, 'The given value of 2♠ returns 2');
+console.assert(getCardValue('A♠') === 11, 'The given value of "A" returns 11');
+console.assert(getCardValue('8♠') === 8, 'The given value of 8♠ returns 8');
+console.assert(getCardValue('J♠') === 10, 'The given value of J returns 10');
+console.assert(getCardValue('Q♠') === 10, 'The given value of 2♠ returns 10');
```
cjyuan, line 54: What do you expect from the following function calls?
Does your function return the value you expected?

```
getCardValue("0Q♠");
getCardValue("010♠");
getCardValue("02♠");
getCardValue("0x02♠");
getCardValue("2.1♠")
```

Sprint-3/implement/get-card-value.js
```diff
@@ -29,3 +29,26 @@
...
+      return Number(cardValue);
+    }
+  }
+  else if (/^[JQK]$/.test(cardValue)) {
+    return 10;
+  }
+  else if (cardValue === 'A') {
+    return 11;
+  }
+}
```
cjyuan, line 48: I think you missed this description:
```
```

cjyuan (commented):
I left my comments in the code. If you need assistance on using Jest, please let me know or seek assistance from a volunteers.

gai93003:
Thanks alot for the review, i will work on all the areas you suggested and
i will get back to you. I would appreciate if you could walk me through
jest and the use of assertions in node js whenever you have time.

Sprint-3/implement/get-card-value.js
gai93003: The function will throw an error 'Invalid card rank since' since the above numbers doesn't match the numbers between 2 and 10. in a blackjack.

Sprint-3/implement/is-proper-fraction.js
```diff
@@ -32,3 +32,44 @@
...
+    return false;
+  }
+}
+
+console.assert(isProperFraction(3, 0) === false, `No target output: Error (Denominator cannot be zero), Explanation: The function should throw an error when the denominator is zero, as it's not a valid fraction.`);
+
+// console.assert(isProperFraction(-4, 7), `target output: false, Explanation: The fraction  is not a proper fraction because the numerator is equal to the denominator. The function should return false.`);
+
+console.assert(isProperFraction(-4, 7) === true, `target output: false, Explanation: The fraction  is a proper fraction because the absolute value of the numerator (4) is less than the denominator (7). The function should return true.`);
+
```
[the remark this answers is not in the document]
gai93003: I have added the absolute method to the function.

Sprint-3/implement/is-valid-triangle.js
```diff
@@ -33,3 +33,17 @@
 // Then it should return true because the input forms a valid triangle.
 
 // This specification outlines the behavior of the isValidTriangle function for different input scenarios, ensuring it properly checks for invalid side lengths and whether they form a valid triangle according to the Triangle Inequality Theorem.
+
+function isValidTriangle(a, b, c) {
+  if (a + b <= c ||  a + c <= b || b + c <= a) {
+    return false;
+  }
+  else if (a <= 0 || b <= 0 || c <= 0) {
```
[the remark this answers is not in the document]
gai93003: I was trying to fulfill the acceptance criteria 
// scenario: invalid triangle
// Check for Valid Input:
// Given the sides a, b, and c,
// When any of the sides are less than or equal to zero,
// Then it should return false because a triangle cannot have zero or negative side lengths.

Sprint-3/implement/rotate-char.js
gai93003: console.log(rotateCharacter("x", 3)); - returns 'A'
console.log(rotateCharacter("Y", 1)); - returns a

I will look into learning more the operator. Thanks for the recommendation.

Sprint-3/revise/implement/get-ordinal-number.js
gai93003: I think i might have misplaced it in a different branch with the other files

Sprint-3/revise/implement/is-prime.js
gai93003: Alright, let me try implement the suggestions.

Sprint-3/revise/implement/password-validator.test.js
gai93003: Not really, i failed to change the description on the test.

Sprint-3/revise/implement/password-validator.test.js
gai93003: I missed out the digit checks in the regular expressions.

PR #209, opened 2024-11-29
CYF London | Fatma Arslantas | Module-Structuring-and-Testing-Data | Week 3

AFatmaa:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/revise/implement/get-ordinal-number.js
+++ b/Sprint-3/revise/implement/get-ordinal-number.js
@@ -4,19 +4,27 @@ function getOrdinalNumber(num) {
     if (num <= 0) return `${num}`;
 
     // Define the suffixes for ordinal numbers.
-    const suffixes = ["th", "st", "nd", "rd"];
+    const suffixes = ["th", "st", "nd", "rd"];  // Array holding ordinal suffixes for numbers
 
     // Extract the last two digits to handle special cases (e.g., 11th, 12th, 13th).
     const value = num % 100;
 
-    // For numbers like 11, 12, 13, we return "th" directly to handle the special cases.
+    // For numbers ending in 11, 12, or 13, the suffix is always "th" regardless of the last digit.
+    // This is because these numbers are exceptions to the general rule for ordinal suffixes.
     if (value >= 11 && value <= 13) {
         return `${num}th`;
     }
 
+    // This expression determines the correct suffix for the number 'num' based on its last digit.
+    // 1. `value % 10` gets the last digit of the number.
+    // 2. `suffixes[value % 10]` tries to access the appropriate suffix in the array.
+    //    - For example, if `value % 10` is 1, it selects "st" from the array.
+    //    - If `value % 10` is 5, there’s no corresponding index in the array, so it returns `undefined`.
+    // 3. `|| suffixes[0]` ensures that if the result is `undefined`, the default "th" is used.
+    //    - This handles cases where the last digit is outside the array bounds, like 5, 6, etc.
     // For all other numbers, apply the usual rule to add the correct suffix.
     // This handles cases like 1st, 2nd, 3rd, etc.
-    return `${num}${suffixes[value % 10] || suffixes[0]}`;
+    return `${num}${suffixes[value % 10] || suffixes[0]}`; // Concatenate the number with the correct suffix
 }
 
 console.log(getOrdinalNumber(-1));
@@ -30,5 +38,6 @@ console.log(getOrdinalNumber(21));
 console.log(getOrdinalNumber(101));
 console.log(getOrdinalNumber(111));
 console.log(getOrdinalNumber(42));
+console.log(getOrdinalNumber(15));
 
 module.exports = getOrdinalNumber;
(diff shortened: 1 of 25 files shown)


Sprint-3/revise/implement/get-ordinal-number.js
```diff
@@ -0,0 +1,29 @@
...
+
+    // Define the suffixes for ordinal numbers.
+    const suffixes = ["th", "st", "nd", "rd"];
+
+    // Extract the last two digits to handle special cases (e.g., 11th, 12th, 13th).
+    const value = num % 100;
+
+    // Return the number with the appropriate suffix.
+    // Use modulus to determine the correct suffix or fall back to "th".
+    return `${num}${suffixes[(value - 20) % 10] || suffixes[value] || suffixes[0]}`;
```
cjyuan, line 14: Can you elaborate how this expression yield the suffix value when `value` is 11?
`suffixes[(value - 20) % 10] || suffixes[value] || suffixes[0]`

Sprint-3/revise/implement/is-prime.js
```diff
@@ -0,0 +1,21 @@
+
+function isPrime(num) {
+    // Check if the input is less than 2, which is not prime
+    if (num < 2) return false;
+
+    // Loop from 2 to the square root of the number
```
cjyuan, line 6: You can possibly improve the performance of the code in the following manners:
- Check if num is 2, and check only odd numbers >= 3 in the loop
- Avoid calling `Math.sqrt(num)` repeatedly by assigning the value of `Math.sqrt(num)` to a variable once, and then refer to the variable in the condition of the loop.
  - Note: The condition is checked at the start of every iteration.

Sprint-3/revise/implement/is-prime.test.js
```diff
@@ -1,3 +1,43 @@
...
+    });
+
+    test('should handle larger prime numbers', () => {
+        expect(isPrime(101)).toBe(true);
+        expect(isPrime(103)).toBe(true);
+        expect(isPrime(107)).toBe(true);
+    });
+
+    test('should handle larger non-prime numbers', () => {
+        expect(isPrime(100)).toBe(false);
```
cjyuan, line 39: Could also test some odd composite numbers, and specify the parameter as `101 * 13` (a composite number as a product of two prime numbers).

Sprint-3/revise/implement/password-validator.test.js
```diff
@@ -14,3 +14,41 @@ To be valid, a password must:
...
+
+describe('isPasswordValid', () => {
+  const previousPasswords = ["Passw0rd!", "Admin123$", "Hello*World1"];
+
+  test('should return true for a valid password', () => {
+    expect(isPasswordValid("New*Pass1", previousPasswords)).toBe(true);
+  });
+
+  test('should return false for a password that is too short', () => {
+    expect(isPasswordValid("sh1*", previousPasswords)).toBe(false);
```
cjyuan, line 28: "Sh1*" would be a better test value because when the function return false, it would imply "Sh1*" passes all other checks except the password length check.

AFatmaa:
Hi @cjyuan, 

Thank you for your review. I'll check and fix them! 🙌😊

AFatmaa:
Hi @cjyuan,

Thank you for taking the time to review my PR and for your detailed comments and suggestions. I've carefully examined each of them, made the necessary adjustments, and committed the changes. Your feedback was helpful and I truly appreciate your guidance and feedback!

Please let me know if there’s anything else that needs improvement. 🙌

cjyuan:
Code is correct.

Can you explain how this expression work when `value` is 15?
```
    const suffixes = ["th", "st", "nd", "rd"];
    return `${num}${suffixes[value % 10] || suffixes[0]}`;
```

AFatmaa:
Hi @cjyuan,
If I understand your question correctly, here’s how I would explain it:

Can you explain how this expression work when value is 15?

    const suffixes = ["th", "st", "nd", "rd"];
    return `${num}${suffixes[value % 10] || suffixes[0]}`;
    
1. value % 10
- This expression calculates the last digit of the number value.
- For example: If value = 15, then 15 % 10 equals 5.

2. suffixes[value % 10]
- The suffixes array is defined as: const suffixes = ["th", "st", "nd", "rd"];
- The elements in this array are:
    suffixes[0] → "th"
    suffixes[1] → "st"
    suffixes[2] → "nd"
    suffixes[3] → "rd"
- The expression suffixes[value % 10] attempts to access the array at the index of the last digit of value.
- If value % 10 is outside the array’s bounds (e.g., there is no suffixes[5]), the result is undefined.

3. || suffixes[0]
- The || operator in JavaScript works as a fallback mechanism.
- If the left-hand side is a "falsy" value (like undefined), it evaluates the right-hand side instead.
- Here, if suffixes[value % 10] is undefined, the code will use suffixes[0], which is "th".

Example: value = 15
1. value % 10:  15 % 10 → 5
2. suffixes[5]: Since suffixes do not have an element at index 5, this evaluates to undefined.
3. || suffixes[0]: undefined || suffixes[0] → "th"
4. Final result: The function returns "15th".

The || suffixes[0] part ensures that the default "th" suffix is used if there’s no valid suffix at the index value % 10. For example, when the last digit is 5 or 6, "th" is returned as a fallback.

Thank you for taking the time to ask detailed questions and help me learn. I really appreciate it! 😊

PR #213, opened 2024-12-01
WM4 | Fatima Safana |Module 2/Sprint1 Exercises | Week4

Fatimasfn:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/errors/3.js
+++ b/Sprint-1/errors/3.js
@@ -1,5 +1,6 @@
-const cardNumber = 4533787178994213;
+const cardNumber = "4533787178994213";
 const last4Digits = cardNumber.slice(-4);
+console.log(last4Digits);
 
 // The last4Digits variable should store the last 4 digits of cardNumber
 // However, the code isn't working
@@ -7,3 +8,6 @@ const last4Digits = cardNumber.slice(-4);
 // Then run the code and see what error it gives.
 // Consider: Why does it give this error? Is this what I predicted? If not, what's different?
 // Then try updating the expression last4Digits is assigned to, in order to get the correct value
+//Answers:
+/*The code should print out 4213, because .slice() is a string type function, thus 
+value needs to be "string" */
--- a/Sprint-1/errors/4.js
+++ b/Sprint-1/errors/4.js
@@ -1,2 +1,3 @@
-const 12HourClockTime = "20:53";
-const 24hourClockTime = "08:53";
+const twelveHourClockTime = "20:53";
+const twentyFourHourClockTime = "08:53";
+//variable names cannot start with numbers.
--- a/Sprint-1/exercises/decimal.js
+++ b/Sprint-1/exercises/decimal.js
@@ -7,3 +7,13 @@ const num = 56.5678;
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
+
+let wholeNumberPart = Math.round(num);
+    console.log(wholeNumberPart)
+
+let decimalPart = num - Math.floor(num);
+    console.log(decimalPart);
+
+let roundedNum = Math.round(num);
+     console.log(roundedNum);
+
--- a/Sprint-1/exercises/paths.js
+++ b/Sprint-1/exercises/paths.js
@@ -11,8 +11,16 @@
 
 const filePath = "/Users/mitch/cyf/Module-JS1/week-1/interpret/file.txt";
 const lastSlashIndex = filePath.lastIndexOf("/");
+    console.log(lastSlashIndex);
 const base = filePath.slice(lastSlashIndex + 1);
+    console.log(base);
 console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
+
+const dirPart = filePath.substring(45, -1);
+    console.log(`The directory path of ${dirPart}`);
+
+const extPart = dirPart + "ext.txt";
+    console.log(`The ext path is ${extPart}`);
--- a/Sprint-1/exercises/random.js
+++ b/Sprint-1/exercises/random.js
@@ -7,3 +7,7 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+console.log(num);
+
+// The program creates a random whole number between 1 and 100, max and min inclusive.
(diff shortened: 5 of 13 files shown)


Sprint-1/errors/3.js
```diff
@@ -1,9 +1,13 @@
-const cardNumber = 4533787178994213;
+const cardNumber = "4533787178994213";
```
cjyuan, line 1: If you cannot modify this statement `const cardNumber = 4533787178994213;`
(that is, keep the variable's value unchanged), 
how would you modify the code so that it can still extract the last 4 digits from its value.

Sprint-1/errors/4.js
```diff
@@ -1,2 +1,3 @@
-const 12HourClockTime = "20:53";
-const 24hourClockTime = "08:53";
\ No newline at end of file
+const twelveHourClockTime = "20:53";
```
cjyuan, line 1: Have you also noticed the variable names do not quite match the values assigned to the variable?

Sprint-1/exercises/decimal.js
```diff
@@ -7,3 +7,13 @@ const num = 56.5678;
 // Create a variable called roundedNum and assign to it an expression that evaluates to 57 ( num rounded to the nearest whole number )
 
 // Log your variables to the console to check your answers
+
+let wholeNumberPart = Math.round(num);
+    console.log(wholeNumberPart)
+
+let decimalPart = num - Math.floor(num);
+    console.log(decimalPart);
```
cjyuan, line 15: If you run the program, you may notice that `console.log()` outputs the value of `decimalPart` 
with more than 4 decimal places (which is normal because of floating point arithmetic). 

If you were asked to print the value of `decimalPart` to exactly 4 decimal places,
how would you modify the code?

Sprint-1/exercises/paths.js
```diff
@@ -11,8 +11,16 @@
...
 const lastSlashIndex = filePath.lastIndexOf("/");
+    console.log(lastSlashIndex);
 const base = filePath.slice(lastSlashIndex + 1);
+    console.log(base);
 console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
+
+const dirPart = filePath.substring(45, -1);
```
cjyuan, line 22: Your code should be written to to work for any valid value of `filePath`.
For example, `"/path1/path2/package.json"`.

Sprint-1/exercises/paths.js
```diff
@@ -11,8 +11,16 @@
...
+    console.log(`The directory path of ${dirPart}`);
+
+const extPart = dirPart + "ext.txt";
```
cjyuan, line 25: The "ext part" refers to the file extension of the file represented by the file path, which is the string after the dot in the base (including the dot).

So the file extension of "package.json" would be ".json".

Sprint-1/exercises/random.js
```diff
@@ -7,3 +7,7 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+console.log(num);
+
+// The program creates a random whole number between 1 and 100, max and min inclusive.
```
cjyuan, line 13: This exercise was expecting you to **break down**
the expression `Math.floor(Math.random() * (maximum - minimum + 1)) + minimum`, and explain
why such expression can correctly produce a random integer in the range [1, 100].

Also, to test your understanding, how would you write an expression 
(without using any variable in the expression) that yields a random 
integer between -5 and 3 (including -5 and 3)?

Sprint-1/interpret/to-pounds.js
```diff
@@ -24,4 +29,16 @@ console.log(`£${pounds}.${pence}`);
 // Try and describe the purpose / rationale behind each step
 
 // To begin, we can start with
-// 1. const penceString = "399p": initialises a string variable with the value "399p"
+// 1. const penceString = "399p": initializes a string variable with the value "399p"
+// 3. initializes a variable that removes the "p" at the end, it uses method .substring
+// which returns the part of the string corresponding to the index and end index
+//-1 excluding the p
+//9. uses the method padstart to add "0" to the start of string penceStringWithoutTrailingP,
```
cjyuan, line 36: After line 9, what would `paddedPenceNumberString` be if `penceString` was "1234p" initially?
After line 9, what would `paddedPenceNumberString` be if `penceString` was "4p" initially?

Sprint-1/interpret/to-pounds.js
```diff
@@ -24,4 +29,16 @@ console.log(`£${pounds}.${pence}`);
...
+// it returns the same string unchanged because the target length is equal to the current length.
+// this effectively multiplies any value input as penceStringWithoutTrailingP by 100
+//12. uses the substring method to return the first value of the string, this is to get
+// the pound. it also used length -2 to make sure its in a valid range so the program
+// can go through the steps with any value input.
```
cjyuan, line 41: What do you mean by "first value of the string"? 
`length - 2` is a valid range but  "making sure its in a valid range" is not quite the reason why `length - 2` 
is used in the code.

Sprint-1/interpret/to-pounds.js
```diff
@@ -24,4 +29,16 @@ console.log(`£${pounds}.${pence}`);
...
+//18.the code gets the last two digits from the paddedPenceNumberString and ensures that the result
```
cjyuan, line 42: Can we expect this program to work as intended even if we deleted `.padEnd(2, "0")` from the code?

cjyuan (commented):
Code is good. I just left some suggestions and questions to challenge you in the code.

Sprint-1/errors/3.js
Fatimasfn: the type of variable needs to string for .slice() method to work.

Sprint-1/exercises/decimal.js
Fatimasfn: let decimalPart = (num - Math.floor(num)).toFixed(4);
    console.log(decimalPart);

Sprint-1/interpret/to-pounds.js
Fatimasfn: It will return 1234
It will return 004

Sprint-1/interpret/to-pounds.js
Fatimasfn: no, it will not work for single digit or double digits input

cjyuan: Well, can you find a value of `penceString` to show that the function won't work properly without `.padEnd(2, "0")`?

PR #219, opened 2024-12-07
CYF-ITP-South Africa | Rashaad Ebrahim | Module-Structuring-and-Testing-Data | Week 2

Rashaad-Ebrahim:
## Learners, PR Template

Self checklist

## Changelist

Sprint 2 exercises complete

## Questions

N/A

### DIFF
--- a/Sprint-2/debug/2.js
+++ b/Sprint-2/debug/2.js
@@ -1,8 +1,8 @@
 // Predict and explain first...
 
-const num = 103;
+// const num = 103;
 
-function getLastDigit() {
+function getLastDigit(num) {
   return num.toString().slice(-1);
 }
 
@@ -12,3 +12,6 @@ console.log(`The last digit of 806 is ${getLastDigit(806)}`);
 
 // This program should tell the user the last digit of each number.
 // Explain why getLastDigit is not working properly - correct the problem
+
+// num is declared as a constant variable on ln 3 and all results from then on are based on this value.
+// I have changed the function by having num as an argument in the function.
--- a/Sprint-2/extend/format-time.js
+++ b/Sprint-2/extend/format-time.js
@@ -3,22 +3,77 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
+  const minutes = time.slice(time.indexOf(":") + 1, time.indexOf(":") + 3);
+  // console.log(minutes);
+  if (hours === 0) {
+    return `${hours + 12}:${minutes} am`;
+  }
   if (hours > 12) {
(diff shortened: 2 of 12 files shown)


Sprint-2/debug/2.js
```diff
@@ -12,3 +12,6 @@ console.log(`The last digit of 806 is ${getLastDigit(806)}`);
 
 // This program should tell the user the last digit of each number.
 // Explain why getLastDigit is not working properly - correct the problem
+
+// num is declared as a constant variable on ln 3 and all results from then on are based on this value.
+// I have changed the function by having num as an argument in the function.
```
cjyuan, line 17: The change is good. 

With the way you changed the function, we can also keep `const num=103;` untouched. If you are interested in understanding why, may I suggest looking up (or ask ChatGPT)  "How local variable shadow global variable in JS".

Sprint-2/extend/format-time.js
```diff
@@ -3,22 +3,77 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
+  const minutes = time.slice(time.indexOf(":") + 1, time.indexOf(":") + 3);
+  // console.log(minutes);
+  if (hours === 0) {
+    return `${hours + 12}:${minutes} am`;
```
cjyuan, line 9: Could also be expressed as 
```
return `12:${minutes} am`;
```

Sprint-2/extend/format-time.js
```diff
@@ -3,22 +3,77 @@
...
+  {
+    currentOutput: formatAs12HourClock("4:05"),
+    targetOutput: "4:05 am",
+  },
+  {
+    currentOutput: formatAs12HourClock("13:05"),
+    targetOutput: "1:05 pm",
+  },
+];
+
```
cjyuan, line 57: What do you expect `formatAs12HourClock("12:05")` to return? And does your function return the expected value?

cjyuan (commented):
Code is good. There is 1 bug which you should try to fix.

Sprint-2/debug/2.js
Rashaad-Ebrahim: Yeah, I've got a good understanding of scope now.

Sprint-2/extend/format-time.js
Rashaad-Ebrahim: Ok, I noticed that anytime the hour is 12, it should be treated as an edge case. I've added an if statement to deal specifically with those cases.

`  if (Number(hours) === 12) {
    return `${hours}:${minutes} pm`;
  }`

PR #224, opened 2024-12-09
northwest |Rahwa Zeslus |module-structure-and-Testing-Data | WEEK3

RahwaZeslusHaile:
northwest |Rahwa Zeslus |module-structure-and-Testing-Data | WEEK3

### DIFF
--- a/Sprint-3/implement/get-angle-type.js
+++ b/Sprint-3/implement/get-angle-type.js
@@ -25,3 +25,26 @@
 // Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
+function getAngleType(angleMeasure){
+   
+   if (angleMeasure === 90){
+      return "Right_Angle";
+    }
+   else if (angleMeasure < 90) { 
+        return "Acute angle"; 
+      } 
+   else if (angleMeasure > 90 && angleMeasure < 180) { 
+      return "Obtuse angle"; 
+    } 
+   else if (angleMeasure===180) { 
+        return "Straight angle"; 
+      }  
+   else if (angleMeasure > 180) { 
+        return "Reflex angle"; 
+      }  
+}
+console.log(getAngleType(90))
+console.log(getAngleType(110))
+console.log(getAngleType(10))
+console.log(getAngleType(180))
+console.log(getAngleType(200))
--- a/Sprint-3/implement/get-angle-type.js
+++ b/Sprint-3/implement/get-angle-type.js
@@ -25,26 +25,40 @@
 // Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
-function getAngleType(angleMeasure){
-   
-   if (angleMeasure === 90){
-      return "Right_Angle";
-    }
-   else if (angleMeasure < 90) { 
-        return "Acute angle"; 
-      } 
-   else if (angleMeasure > 90 && angleMeasure < 180) { 
-      return "Obtuse angle"; 
-    } 
-   else if (angleMeasure===180) { 
-        return "Straight angle"; 
-      }  
-   else if (angleMeasure > 180) { 
-        return "Reflex angle"; 
-      }  
+
+
+
+function getAngleType(angleMeasure) {
+  if (typeof angleMeasure !== "number") {
+       return "Invalid angle";
+  }
+
+  angleMeasure = angleMeasure % 360;
+  if (angleMeasure < 0) {
+      angleMeasure += 360;
+  }
+
+  if (angleMeasure === 0 || angleMeasure === 360) {
+      return "Full angle";
+  } else if (angleMeasure === 90) {
+      return "Right angle";
+  } else if (angleMeasure > 0 && angleMeasure < 90) {
+      return "Acute angle";
+  } else if (angleMeasure > 90 && angleMeasure < 180) {
+      return "Obtuse angle";
+  } else if (angleMeasure === 180) {
+      return "Straight angle";
+  } else if (angleMeasure > 180 && angleMeasure < 360) {
+      return "Reflex angle";
+  }
 }
-console.log(getAngleType(90))
-console.log(getAngleType(110))
-console.log(getAngleType(10))
-console.log(getAngleType(180))
-console.log(getAngleType(200))
+
+console.log(getAngleType(1000));  
+console.log(getAngleType(-1000)); 
+console.log(getAngleType(360));   
+console.log(getAngleType(-450));  
+console.log(getAngleType(90));    
+console.log(getAngleType(110));   
+console.log(getAngleType(10));    
+console.log(getAngleType(180));   
+console.log(getAngleType(200));   
--- a/Sprint-3/implement/get-card-value.js
+++ b/Sprint-3/implement/get-card-value.js
@@ -29,3 +29,27 @@
 // Given a card with an invalid rank (neither a number nor a recognized face card),
 // When the function is called with such a card,
 // Then it should throw an error indicating "Invalid card rank."
+
+
+function getCardValue(card) {
+    const rank = card.slice(0, -1);
+    // to exclude the last character
+
+    // Check for Number Cards (2-10)
+    if (!isNaN(rank) && Number(rank) >= 2 && Number(rank) <= 10) {
+        return Number(rank);
+    }
+
+    // Check for Face Cards (J, Q, K) or '10'
+    if (['J', 'Q', 'K'].includes(rank) || rank === '10') {
+        return 10;
+    }
+
+    // Check for Ace (A)
+    if (rank === 'A') {
+        return 11;
+    }
+
+    
+    throw new Error("Invalid card rank.");
+}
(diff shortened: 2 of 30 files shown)


codrex (changes requested):

RahwaZeslusHaile:
thank you for your constructive reviews, I updated with the corrected one .

RahwaZeslusHaile:
Please take a look when you have the chance and let me know if there are any improvements or issues you see. Thanks in advance

cjyuan:
@RahwaZeslusHaile I think you can tag the reviewer (like what I did here with your ID) to let the reviewer know you would like a follow-up review from the reviewer.

Sprint-3/implement/is-valid-triangle.js
```diff
@@ -33,3 +33,24 @@
...
+        return true;
+    }
+// scenario: invalid triangle
+// Check for Valid Input:
+// Given the sides a, b, and c,
+// When any of the sides are less than or equal to zero,
+// Then it should return false because a triangle cannot have zero or negative side lengths.
+    if(a <= 0 || b <= 0 || c <= 0){
+        return false;
+    }
```
cjyuan, line 53: If any of `a`, `b`, and `c` is less than or equal to zero, then the condition at line 43 will always be false.
Is there any need to further check if `a`, `b`, and `c` is less than or equal to zero at line 51?

I will not go into details why in some programming languages (but not JavaScript) we need also to ensure a, b, c are positives.

Sprint-3/implement/rotate-char.js
```diff
@@ -15,14 +15,36 @@
...
 // Given a lowercase letter character and a positive integer shift,
 // When the function is called with these inputs,
+
 // Then it should rotate the lowercase letter by shift positions within the lowercase alphabet, wrapping around if necessary, and return the rotated lowercase letter as a string.
+function rotateCharacter(char, shift) {
+    // Check if the character is a lowercase letter
+    if (char >= "a" && char <= "z") {
+        // Calculate new character with wraparound
+        const charCode = char.charCodeAt(0);
+        const rotatedCode = ((charCode - 97 + shift) % 26) + 97; // 'a' starts at ASCII 97
```
cjyuan, line 25: This approach is good. 
If `shift` were allowed to be negative so that `charCode - 97 + shift` can become a negative number, how would you modify your code to deal with negative `shift`?

Sprint-3/revise/implement/card-validator.md
```diff
@@ -33,3 +33,33 @@ These are the requirements your project needs to fulfill:
...
+    for (let i = 0; i < cardNumber.length; i++) {
+        sum += parseInt(cardNumber.charAt(i), 10);
+    }
+
+    if (sum <= 16) {
+        return false;
+    }
+
+    return true;
+}
```
cjyuan, line 65: Have you fully tested this function? It missed checking something in the given `cardNumber`.

Sprint-3/revise/implement/is-prime.js
```diff
@@ -0,0 +1,28 @@
+//it is positive number
+// a number that can only be divided by itself and 1 without remainders
+// all are odd except 2
+
+
+function isPrimeNumber(number) {
+    if (number <= 1) {
+        throw new Error('There is no negative prime number or zero as a prime number.');
```
cjyuan, line 8: How are these non-prime numbers (e.g., -100, 1, 0) any different from other non-prime numbers (e.g., 10, 15, 100)?

Sprint-3/revise/implement/is-prime.js
```diff
@@ -0,0 +1,28 @@
...
+
+    if (number === 2) {
+        return true;
+    }
+
+    if (number % 2 === 0) {
+        return false;
+    }
+
+    for (let index = 3; index <= Math.sqrt(number); index += 2) {
```
cjyuan, line 19: You can possibly improve the performance of the code in the following manner:
- Avoid calling `Math.sqrt(num)` repeatedly by assigning the value of `Math.sqrt(num)` to a variable once, and then refer to the variable in the condition of the loop.
  - Note: The condition is checked at the start of every iteration.

Sprint-3/revise/implement/password-validator.test.js
```diff
@@ -14,3 +14,41 @@ To be valid, a password must:
 
 You must breakdown this problem in order to solve it. Find one test case first and get that working
 */
+const isValidPassword = require('./password-validator');
+
+describe('isValidPassword', () => {
+    const previousPasswords = ['Password1!', 'Admin123#'];
+ 
+      it('should return false for passwords with less than 5 characters', () => {
+        expect(isValidPassword('P1!', previousPasswords)).toBe(false);
```
cjyuan, line 23: The function could have also returned `false` because `P1!` does not contain any lowercase letter. How can we be 100% sure the function can correctly check passwords shorter than 5 characters?

Sprint-3/revise/implement/repeat.js
```diff
@@ -0,0 +1,14 @@
...
+    if (count < 0) {
+        throw new Error('Count cannot be negative');
+    }
+    if (count === 0) {
+        return '';
+    }
+    if (count === 1) {
+        return str;
+    }
+    return str.repeat(count);
```
cjyuan, line 11: These statements could be greatly shorten if you know how `String.repeat()` works.

cjyuan (commented):
The code is quite solid.
`card-validator.md` has a bug that needs fixing.

Sprint-3/implement/get-angle-type.js
```diff
@@ -25,3 +25,26 @@
...
+   else if (angleMeasure > 90 && angleMeasure < 180) { 
+      return "Obtuse angle"; 
+    } 
+   else if (angleMeasure===180) { 
+        return "Straight angle"; 
+      }  
+   else if (angleMeasure > 180) { 
+        return "Reflex angle"; 
+      }  
+}
```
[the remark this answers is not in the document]
RahwaZeslusHaile: Yes I understand ,I didn't see that way. So what I have done in this new code is try to change the degrees to to an equivalent angle within the standard range of 
0 degree to 360 degree

Sprint-3/implement/get-card-value.js
```diff
@@ -29,3 +29,27 @@
...
+
+
+function getCardValue(card) {
+    const rank = card.slice(0, -1);
+    // to exclude the last character
+
+    // Check for Number Cards (2-10)
+    if (!isNaN(rank) && Number(rank) >= 2 && Number(rank) <= 10) {
+        return Number(rank);
+    }
```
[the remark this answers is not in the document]
RahwaZeslusHaile: 1,getCardValue("010♠");  
!isNaN(rank) → true because "010" is a numeric string.
Number(rank) → 10 (leading zeros are ignored by Number()).
Number(rank) >= 2 && Number(rank) <= 10 → true.

2,getCardValue("02♠");  

!isNaN(rank) → true because "02" is a numeric string.
Number(rank) → 2 (leading zeros are ignored by Number()).
Number(rank) >= 2 && Number(rank) <= 10 → true.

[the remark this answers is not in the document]
cjyuan: At the end, do you want the function to recognise these card values?

Sprint-3/implement/is-valid-triangle.js
RahwaZeslusHaile: Thank you for your review! You're absolutely right in your observation. Since the check for whether any side is less than or equal to zero is already handled in the first condition (a <= 0 || b <= 0 || c <= 0) at line 43, the second check at line 51 is unnecessary.

This means the function can be simplified by removing that second check, making the logic more concise and clearer.

Sprint-3/revise/implement/card-validator.md
RahwaZeslusHaile: Yes! You are right—it misses checking the card in a lot of ways. I updated the code to have more comprehensive validations, including checks for:

1,Ensuring the card contains only numeric characters.
2,Verifying the card number length is exactly 16 digits.
3,Confirming the card has at least two unique digits.
4,Applying the Luhn algorithm for number validation.
5,Validating that the last digit (check digit) is even

Sprint-3/revise/implement/is-prime.js
RahwaZeslusHaile: you’re absolutely right that treating numbers less than or equal to 1 differently by throwing an error while returning false for other non-prime numbers (e.g., 10, 15) introduces unnecessary complexity. Thank you

Sprint-3/revise/implement/is-prime.js
RahwaZeslusHaile: Thank you for pointing out the potential improvement in performance by avoiding the repeated calculation of Math.sqrt(number) in the loop condition.

Sprint-3/revise/implement/password-validator.test.js
RahwaZeslusHaile: Thank you for pointing out that ,I tried to include requirement to make the password valid(the lower case letter, uppercase letter, a number,and one-alphanumeric symbol) unless the length is less than 5. So the test only checks that the length is less than five and returns false.

RahwaZeslusHaile:
I’ve made corrections to improve the overall functionality, readability, and efficiency of the code. 

Can you please review the updates and provide feedback?

PR #230, opened 2024-12-16
LONDON | Pooriya Ketabi | Module-Structuring-and-Testing-Data | Sprint 3

PooriyaKTB:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/CYF/ordinal-testing-example/get-ordinal-number.test.js
+++ b/CYF/ordinal-testing-example/get-ordinal-number.test.js
@@ -1,0 +1,9 @@
+function getOrdinalNumber() {
+    return "1st";
+}
+
+test("converts 1 to an ordinal number", function() {
+    expect(getOrdinalNumber(1)).toEqual("1st");
+    expect(getOrdinalNumber(11)).toEqual("11th");
+    expect(getOrdinalNumber(21)).toEqual("21st");
+});
--- a/Sprint-3/implement/get-card-value.js
+++ b/Sprint-3/implement/get-card-value.js
@@ -29,3 +29,22 @@
 // Given a card with an invalid rank (neither a number nor a recognized face card),
 // When the function is called with such a card,
 // Then it should throw an error indicating "Invalid card rank."
+
+function getCardValue(card) {
+  let rank = card.slice(0, -1);
+  if (parseInt(rank) >= 2 && parseInt(rank) <= 10) {
+    return parseInt(rank);
+  }
+  switch (rank) {
+    case "J":
+    case "Q":
+    case "K":
+      return 10;
+    case "A":
+      return 11;
+    default:
+      throw new Error("Invalid card rank");
+  }
+}
+
+module.exports = getCardValue;
--- a/Sprint-3/implement/get-card-value.test.js
+++ b/Sprint-3/implement/get-card-value.test.js
@@ -1,0 +1,20 @@
+const getCardValue = require("./get-card-value");
+test("should return the correct value for number cards (2-10)", () => {
+  expect(getCardValue("2♠")).toBe(2);
+  expect(getCardValue("9♥")).toBe(9);
+  expect(getCardValue("10♦")).toBe(10);
+});
+
+test("should return 10 for face cards (J, Q, K)", () => {
+  expect(getCardValue("J♠")).toBe(10);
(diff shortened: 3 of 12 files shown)


Sprint-3/implement/get-card-value.js
```diff
@@ -29,3 +29,22 @@
 // Given a card with an invalid rank (neither a number nor a recognized face card),
 // When the function is called with such a card,
 // Then it should throw an error indicating "Invalid card rank."
+
+function getCardValue(card) {
+  let rank = card.slice(0, -1);
+  if (parseInt(rank) >= 2 && parseInt(rank) <= 10) {
+    return parseInt(rank);
```
illicitonion, line 36: You're calling `parseInt` on the same input here three times - can you think how you can avoid needing to repeat this? Why is doing the same thing three times not ideal?

Sprint-3/implement/get-card-value.test.js
illicitonion, line 1: These are some really nice tests, good job!

Sprint-3/implement/is-proper-fraction.js
```diff
@@ -32,3 +32,46 @@
...
+  if (denominator === 0) {
+    throw new Error("Denominator cannot be zero");
+  }
+  return Math.abs(numerator) < Math.abs(denominator);
+}
+
+// ---------------------------------------- Tests: ----------------------------------------
+// -------- Proper Fraction --------
+console.assert(
+  isProperFraction(2, 3) === true,
```
illicitonion, line 46: We actually don't need to put `=== true` or `=== false` in assertions. What we need is an expression which evaluates to a boolean (true or false), and `isProperFraction` already returns a boolean, so we can remove the `=== true` (and write `!isProperFraction(...)` where we want to check something is false.

Almost always, writing `something === true` or `something === false` should be replaced by just `something` or `!something` :)

CYF/ordinal-testing-example/get-ordinal-number.test.js
illicitonion, line 1: Ideally this directory wouldn't be included in your coursework PR, as this was just your local work during the prep, rather than the exercises you're submitting for review.

illicitonion (commented):
This is looking really good!

Don't forget to do the playing computer exercise in the `investigate` folder.

Sprint-3/implement/get-card-value.js
PooriyaKTB: It's not ideal, because of performance and readability (debugging). Therefore it's better to store it in a variable and then use the variable instead.
Thank you for mentioning it.

Sprint-3/implement/get-card-value.test.js
PooriyaKTB: Thank you. I'm glad you found it nice.

Sprint-3/implement/is-proper-fraction.js
PooriyaKTB: I truly appreciate your detailed explanations. Your guidance has been really helpful, and I’ve made the necessary changes accordingly.

CYF/ordinal-testing-example/get-ordinal-number.test.js
PooriyaKTB: Thank you for pointing that out.

Sprint-3/implement/get-card-value.js
```diff
@@ -29,3 +29,23 @@
 // Given a card with an invalid rank (neither a number nor a recognized face card),
 // When the function is called with such a card,
 // Then it should throw an error indicating "Invalid card rank."
+
+function getCardValue(card) {
+  let rank = card.slice(0, -1);
+  const parseIntRank = parseInt(rank);
```
illicitonion, line 35: It's useful to differentiate variable names when you have two things which have the same (but for some reason different, e.g. a different type type) data in them.

I probably _wouldn't_ use `parseIntRank` as a variable name - `parse` is a verb, and verbs suggest _doing_ (i.e. function names).

What do you think of each of these variable name pairs?

`rankString` and `rankNumber`

`rankAsString` and `rankAsNumber`

`stringRank` and `numericRank`

`rankString` and `rank`

illicitonion (commented):
Looking really good!

PR #231, opened 2024-12-16
West Midlands | Alireza Seifi Shalamzari | Module-Structuring-and-Testing-Data | Week 3

Alireza1620:
Self checklist

## Changelist

Briefly explain your PR.

## Notes
For the file Rotate Char, it was too hard to do!


Sprint-3/implement/get-angle-type.js
```diff
@@ -25,3 +25,50 @@
...
+}
+// Test the function with different angles
+console.log(getAngleType(90)); // Should print "Right angle"
+console.log(getAngleType(45)); // Should print "Acute angle"
+console.log(getAngleType(135)); // Should print "Obtuse angle"
+console.log(getAngleType(180)); // Should print "Straight angle"
+console.log(getAngleType(225)); // Should print "Reflex angle"
+console.log(getAngleType(360)); // Should print "Reflex angle"
+console.log(getAngleType(0)); // Should print "Invalid angle"
+console.log(getAngleType(-45)); // Should print "Invalid angle"
```
cjyuan, line 62: Does these two show "Invalid angle"?

It is not easy to spot the error from just `console.log()` output. That's one reason why a test framework like Jest can help us identify cases when a return value does not match an expected value.

Sprint-3/implement/get-card-value.js
```diff
@@ -29,3 +29,49 @@
 // Given a card with an invalid rank (neither a number nor a recognized face card),
 // When the function is called with such a card,
 // Then it should throw an error indicating "Invalid card rank."
+
+function getCardValue(card) {
+    // Remove the suit from the card string
+    const rank = card.slice(0, -1);
+    
+    // Handle Number Cards (2-9)
+    if (rank >= "2" && rank <= "9") {
```
cjyuan, line 38: #### Todo

This check is not solid enough. 
You can test your function by calling `getCardValue("23♠")` or `getCardValue("899♠")` to see what they return.

Sprint-3/implement/is-proper-fraction.js
```diff
@@ -32,3 +32,21 @@
...
+    }
+    // Check if the numerator is less than the denominator and return true if it is
+    // Otherwise, return false
+    return numerator < denominator;
+}
+
+//Testing the function
+console.log(isProperFraction(2, 3)); //should return True
+console.log(isProperFraction(5, 2)); //should return False
+console.log(isProperFraction(-4, 7)); //should return True
```
cjyuan, line 49: In mathematics, -4/7 == 4/-7, and -4/-7 == 4/7.
So, ideally `isProperFraction()` should recognise all of them as proper fractions.

Hint: If you compute the absolute value of both parameters inside the function first, the code can become much simpler.

cjyuan (commented):
The code in the 4 .js files you implemented are quite solid.

You also prepared many good test cases but you were not using `console.assert()` or Jest to specify the tests.

There are a few more exercises in folder `revised` not yet included in this branch.

Do you need help setting up Jest or using it to specify unit tests? You can seek help from a volunteer in a workshop on Saturday or DM me on Slack to make an appointment. You can also do the same if you need some guidance in these exercises.

Sprint-3/implement/get-card-value.js
Alireza1620: I did work on this and I found a solution!
I added a set of valid ranks so only the valid ones can return an actual value

Sprint-3/implement/is-proper-fraction.js
Alireza1620: I added a part which first make absolute numbers then check them

Sprint-3/implement/get-angle-type.js
Alireza1620: I changed the testing method to Jest

PR #253, opened 2025-02-04
ITP-JAN-25 | London | Ihor Bolzhelarskyi | Module-Structuring-and-Testing-Data |Sprint 1| Week 4 

IhorBolzhelarskyi:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -17,7 +17,9 @@ console.log(`The base part of ${filePath} is ${base}`);
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
 
-const dir = ;
-const ext = ;
+const dir = filePath.slice(0, lastSlashIndex);
+const ext = filePath.slice(-4);
+console.log(`The base part of ${filePath} is ${dir}`);
+console.log(`The base part of ${filePath} is ${ext}`);
 
 // https://www.google.com/search?q=slice+mdn
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -16,9 +16,9 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
-
+const lastDotIndex = filePath.lastIndexOf(".");
 const dir = filePath.slice(0, lastSlashIndex);
-const ext = filePath.slice(-4);
+const ext = filePath.slice(lastDotIndex);
 console.log(`The base part of ${filePath} is ${dir}`);
 console.log(`The base part of ${filePath} is ${ext}`);
 
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -2,8 +2,15 @@ const minimum = 1;
 const maximum = 100;
 
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
+console.log(num);
 
 // In this exercise, you will need to work out what num represents?
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+// num is generating a random integer number between two values minimum =1 and maximum =100.
+// Math.random() static method returns a floating-point, pseudo-random number that's greater than or equal to 0 and less than 1;
+// (maximum - minimum + 1) (100 - 1 + 1 =100) in our case, Math.random() * 100 generate a random decimal number between 0 - 99.999;
+// The Math.floor() static method always rounds down and returns whole number, in our case 0-99;
+// + minimum in our case + 1, increase random range by 1 to make it between 1 - 100;
--- a/Sprint-1/2-mandatory-errors/3.js
+++ b/Sprint-1/2-mandatory-errors/3.js
@@ -1,5 +1,6 @@
 const cardNumber = 4533787178994213;
-const last4Digits = cardNumber.slice(-4);
+const last4Digits = +cardNumber.toString().slice(-4);
+console.log(last4Digits);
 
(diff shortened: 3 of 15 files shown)


Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -17,7 +17,9 @@ console.log(`The base part of ${filePath} is ${base}`);
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
 
-const dir = ;
-const ext = ;
+const dir = filePath.slice(0, lastSlashIndex);
+const ext = filePath.slice(-4);
```
illicitonion, line 21: This solves the problem for this exact string, but it isn't a very _general_ solution.

Not all file extensions are 3 characters long. How could you write this instead, so that it would also work if the filename was `file.jpeg` or `file.tar.gz`?

(Tip: You've done something called "hard-coding" a value here - you've put in one specific value, rather than calculating the answer that would work for _any_ string).

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -2,8 +2,15 @@ const minimum = 1;
...
 // In this exercise, you will need to work out what num represents?
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+// num is generating a random integer number between two values minimum =1 and maximum =100.
+// Math.random() static method returns a floating-point, pseudo-random number that's greater than or equal to 0 and less than 1;
+// (maximum - minimum + 1) (100 - 1 + 1 =100) in our case, Math.random() * 100 generate a random decimal number between 0 - 99.999;
+// The Math.floor() static method always rounds down and returns whole number, in our case 0-99;
+// + minimum in our case + 1, increase random range by 1 to make it between 1 - 100;
```
illicitonion, line 16: This is a really clear explanation of what's going on, and deals well both with _what_ is being done, as well as _how_ it's being done.

Sprint-1/2-mandatory-errors/3.js
```diff
@@ -1,9 +1,15 @@
...
 // However, the code isn't working
 // Before running the code, make and explain a prediction about why the code won't work
 // Then run the code and see what error it gives.
 // Consider: Why does it give this error? Is this what I predicted? If not, what's different?
 // Then try updating the expression last4Digits is assigned to, in order to get the correct value
+
+// TypeError: cardNumber.slice is not a function
+// We can t use slice() method to numbers, only strings or arrays
+// At the beginning we need to change the value of our variable to string by tostring() method;
+// Then by adding + we convert the extracted string back into a number.
```
illicitonion, line 15: This looks good and makes sense. I'm curious, if the card number ended in `0012` what would you expect to be logged, and what do you actually see logged?

Sprint-1/2-mandatory-errors/4.js
```diff
@@ -1,2 +1,8 @@
...
-const 24hourClockTime = "08:53";
\ No newline at end of file
+// const 12HourClockTime = "20:53";
+// const 24hourClockTime = "08:53";
+
+const twelveHourClockTime = "08:53";
+const twentyFourHourClockTime = "20:53";
+
+// The name of variable in Java Script can't start with number
+// And also there is a time mistake
```
illicitonion, line 8: Good catch!

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -25,3 +25,16 @@ console.log(`£${pounds}.${pence}`);
 
 // To begin, we can start with
 // 1. const penceString = "399p": initialises a string variable with the value "399p"
+// 3.const penceStringWithoutTrailingP = penceString.substring(
+//   0,
+//   penceString.length - 1
+// ); removing last character "p" from string
+// 8. const paddedPenceNumberString = penceStringWithoutTrailingP.padStart(3, "0") Making sure the number has 3 digits.
```
illicitonion, line 32: Are we making sure it has _exactly_ 3 digits?

Sprint-1/4-stretch-explore/objects.md
```diff
@@ -6,11 +6,20 @@ Open the Chrome devtools Console, type in `console.log` and then hit enter
...
 Now enter just `console` in the Console, what output do you get back?
 
+console {debug: ƒ, error: ƒ, info: ƒ, log: ƒ, warn: ƒ, …}
+
 Try also entering `typeof console`
 
 Answer the following questions:
 
-What does `console` store?
+What does `console` store? It stores methods wich we can use for debugging purpose.
```
illicitonion, line 19: This is correct! Which methods exactly does it store? What are their names?

illicitonion (commented):
This is looking really good - good job!

I left a few comments to think about and respond to

Sprint-1/4-stretch-explore/objects.md
IhorBolzhelarskyi: For example console.log and console.assert, that we have already used in our exercises.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
IhorBolzhelarskyi: No, at least '3' and it is not a number actually it is a string

Sprint-1/2-mandatory-errors/3.js
IhorBolzhelarskyi: Thank you, I hadn't noticed that. I expected to get '0012', but I am getting '12' because the string was transformed into a number by using +. I will leave the result as a string.

Sprint-1/1-key-exercises/4-random.js
IhorBolzhelarskyi: Thank you

Sprint-1/1-key-exercises/3-paths.js
IhorBolzhelarskyi: Thank you, not sure how to corectly fix that, now it cut should slice everything from last dot.

illicitonion:
This looks great, thanks for following up!

PR #258, opened 2025-02-06
ITP_J25_LDN| EYUEL_ABRAHAM | Module-Structuring-and-Testing-Data  Coursework/sprint 2 | WEEK-2 

eyuell21:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.
Completed the exercises given for sprint two.

## Questions

### DIFF
--- a/Sprint-2/1-key-errors/1.js
+++ b/Sprint-2/1-key-errors/1.js
@@ -19,8 +19,8 @@
 // =============> write your new code here
 function convertToPercentage(decimalNumber) {
   const percentage = `${decimalNumber * 100}%`;
-
   return percentage;
 }
-decimalNumber = 0.5
-console.log(convertToPercentage(decimalNumber));
+
+let decimalNumber1 = 0.5;
+console.log(convertToPercentage(decimalNumber1));
--- a/Sprint-2/1-key-errors/1.js
+++ b/Sprint-2/1-key-errors/1.js
@@ -1,20 +1,26 @@
 // Predict and explain first...
-
+// changes decimal numbers to percentage
 // Why will an error occur when this program runs?
 // =============> write your prediction here
 
 // Try playing computer with the example to work out what is going on
 
+// function convertToPercentage(decimalNumber) {
+//   const percentage = `${decimalNumber * 100}%`;
+
+//   return percentage;
+// }
+
+// console.log(decimalNumber);
+
+// =============> write your explanation here
+// trying to redeclare the variable called decimalNumber twice in the function.
+// Finally, correct the code to fix the problem
+// =============> write your new code here
 function convertToPercentage(decimalNumber) {
-  const decimalNumber = 0.5;
   const percentage = `${decimalNumber * 100}%`;
 
   return percentage;
 }
-
-console.log(decimalNumber);
-
-// =============> write your explanation here
-
-// Finally, correct the code to fix the problem
-// =============> write your new code here
+decimalNumber = 0.5
+console.log(convertToPercentage(decimalNumber));
--- a/Sprint-2/2-mandatory-debug/1.js
+++ b/Sprint-2/2-mandatory-debug/1.js
@@ -1,13 +1,18 @@
 // Predict and explain first...
-//  =============> write your prediction here
+//  =============> write your prediction here : gives the sum of two numbers a and b. but i will not work because a+b is written under the return statement.
 
+// function sum(a, b) {
+//   return;
+//   a + b;
+// }
+
+// console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
+
+// =============> write your explanation here: The sum of 10 and 32 is undefined because a+b is written under the return statement.
+// Finally, correct the code to fix the problem
+//  =============> write your new code here
 function sum(a, b) {
-  return;
-  a + b;
+  return a + b;
 }
 
 console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
-
-// =============> write your explanation here
-// Finally, correct the code to fix the problem
-//  =============> write your new code here
--- a/Sprint-2/2-mandatory-debug/1.js
+++ b/Sprint-2/2-mandatory-debug/1.js
@@ -8,7 +8,7 @@
 
 // console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
 
-// =============> write your explanation here: The sum of 10 and 32 is undefined because a+b is written under the return statement.
+// =============> write your explanation here: The sum of 10 and 32 is undefined because a+b is written under the return statement.(return; will always return undefined. It is the same as writing return undefined)
 // Finally, correct the code to fix the problem
 //  =============> write your new code here
 function sum(a, b) {
--- a/Sprint-2/2-mandatory-debug/2.js
+++ b/Sprint-2/2-mandatory-debug/2.js
@@ -28,4 +28,6 @@ console.log(`The last digit of 105 is ${getLastDigit(105)}`);
 console.log(`The last digit of 806 is ${getLastDigit(806)}`);
 
 // This program should tell the user the last digit of each number.
-// Explain why getLastDigit is not working properly - correct the problem: it's because now num is passed as a parameter.
+// Explain why getLastDigit is not working properly - correct the problem:  The number wasn't passed into the function as a parameter.
+
+
--- a/Sprint-2/2-mandatory-debug/2.js
+++ b/Sprint-2/2-mandatory-debug/2.js
@@ -1,11 +1,25 @@
 // Predict and explain first...
-
-// Predict the output of the following code:
+// extracts only the last digit of a number.
+// Predict the output of the following code: 3
 // =============> Write your prediction here
 
-const num = 103;
+// let num = 103;
 
-function getLastDigit() {
+// function getLastDigit(num) {
+//   return num.toString().slice(-1);
+// }
+
+// console.log(`The last digit of 42 is ${getLastDigit(42)}`);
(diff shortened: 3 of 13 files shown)


Sprint-2/1-key-errors/1.js
```diff
@@ -1,20 +1,26 @@
...
-}
+//   return percentage;
+// }
 
-console.log(decimalNumber);
+// console.log(decimalNumber);
 
 // =============> write your explanation here
-
+// trying to redeclare the variable called decimalNumber twice in the function.
```
BenSymons, line 17: also- the code was not invoking the `convertToPercentage` function which you corrected

Sprint-2/2-mandatory-debug/1.js
```diff
@@ -1,13 +1,18 @@
...
+// function sum(a, b) {
+//   return;
+//   a + b;
+// }
 
-console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
+// console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
 
-// =============> write your explanation here
+// =============> write your explanation here: The sum of 10 and 32 is undefined because a+b is written under the return statement.
```
BenSymons, line 11: it is actually because that is what the function returns. `return;` will always return undefined. It is the same as writing `return undefined`

Sprint-2/2-mandatory-debug/2.js
```diff
@@ -1,24 +1,31 @@
...
+// function getLastDigit(num) {
+//   return num.toString().slice(-1);
+// }
+
+// console.log(`The last digit of 42 is ${getLastDigit(42)}`);
+// console.log(`The last digit of 105 is ${getLastDigit(105)}`);
+// console.log(`The last digit of 806 is ${getLastDigit(806)}`);
 
-function getLastDigit() {
+// Now run the code and compare the output to your prediction
```
BenSymons, line 16: You've corrected the mistake but your explanation isn't quite right. The problem was the number wasn't passed into the function as a parameter

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,8 @@
 // You will need to come up with an appropriate name for the function
 // Use the MDN string documentation to help you find a solution
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
+
```
BenSymons, line 17: fun fact- this is sometimes called screaming snake case!

Sprint-2/3-mandatory-implement/3-to-pounds.js
```diff
@@ -4,3 +4,16 @@
...
 
 // You should call this function a number of times to check it works for different inputs
+
+function changeToPounds(penceString){
+    let penceStringWithoutTrailingP = penceString.substring(0,penceString.length - 1);
+    let paddedPenceNumberString = penceStringWithoutTrailingP.padStart(3, "0");
+    let pounds = paddedPenceNumberString.substring(0,paddedPenceNumberString.length - 2);
+    const pence = paddedPenceNumberString.substring(paddedPenceNumberString.length - 2).padEnd(2, "0");
+    return `£${pounds}.${pence}`
+}
```
fernandamelov, line 14: Great job so far !

try and test this one here:

function changeToPounds(penceString){
return £${(parseInt(penceString) / 100).toFixed(2)};
}

This is a robuster method to convert to pounds, and easier to understand . You don't need to worry about where de letter p will appear in the string, when using parseInt , it removes all letters elements from the value and takes just the numbers.

And then , just divide it to 100 to transform pence in pounds .

fernandamelov (changes requested):
Great job so far !

try and test this one here for 3-mandatory-implement/3-to-pounds.js:

function changeToPounds(penceString){
return £${(parseInt(penceString) / 100).toFixed(2)};
}

This is a robuster method to convert to pounds, and easier to understand . You don't need to worry about where de letter p will appear in the string, when using parseInt , it removes all letters elements from the value and takes just the numbers.

And then , just divide it to 100 to transform pence in pounds .

Sprint-2/3-mandatory-implement/3-to-pounds.js
```diff
@@ -4,3 +4,16 @@
 // You will need to declare a function called toPounds with an appropriately named parameter.
 
 // You should call this function a number of times to check it works for different inputs
+
+function changeToPounds(penceString){
+    let penceStringWithoutTrailingP = penceString.substring(0,penceString.length - 1);
```
BenSymons, line 9: `string.slice()` is a little easier for this

```
let penceStringWithoutTrailingP = penceString.slice(0, -1);
```

Sprint-2/4-mandatory-interpret/time-format.js
```diff
@@ -18,17 +19,17 @@ function formatTimeDisplay(seconds) {
 
 // a) When formatTimeDisplay is called how many times will pad be called?
 // =============> write your answer here
-
+// three times
```
BenSymons, line 22: is there a way you can check this?

Sprint-2/4-mandatory-interpret/time-format.js
```diff
@@ -18,17 +19,17 @@ function formatTimeDisplay(seconds) {
 
 // a) When formatTimeDisplay is called how many times will pad be called?
 // =============> write your answer here
-
+// three times 
 // Call formatTimeDisplay with an input of 61, now answer the following:
 
 // b) What is the value assigned to num when pad is called for the first time?
```
BenSymons, line 25: this depends on the number passed in at the start. If it's big enough it will be more than 0

Sprint-3/1-key-implement/1-get-angle-type.js
```diff
@@ -8,7 +8,13 @@
 // Then, write the next test! :) Go through this process until all the cases are implemented
 
 function getAngleType(angle) {
-   if (angle === 90) return "Right angle";
+  if (angle === 90) return "Right angle";
```
BenSymons, line 11: none of these if statements need to be `else if` or `else` because you are returning a value from the function. Whenever you return something from a function, the execution stops so none of the code below is executed. 

Also, you know that an angle is < 90 because the code carried on past that if statment so you only need to do:

```
if(angle < 90) return "Acute angle";
if(angle < 180) return "Obtuse angle";
if(angle < 360) return "Reflex angle";
```
etc

Feel free to message me if this doesn't make sense

Sprint-3/1-key-implement/1-get-angle-type.js
```diff
@@ -8,7 +8,13 @@
...
+  else if (angle < 90) return "Acute angle"
+  else if (angle > 90 && angle < 180) return "Obtuse angle"
+  else if (angle === 180) return "Straight angle"
+  else if (angle > 180 && angle < 360) return "Reflex angle"
+  else if(angle === 360)  return "Full angle"
+  else return "please insert a valid angle (0-356)"
```
BenSymons, line 17: angles up to 360 should be valid

Sprint-3/1-key-implement/2-is-proper-fraction.js
```diff
@@ -8,7 +8,10 @@
 // write one test at a time, and make it pass, build your solution up methodically
 
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
+    if (Math.abs(numerator) < Math.abs(denominator) && numerator !=0 && denominator != 0) return true;
```
BenSymons, line 11: Careful. the `Math.abs()` method just takes any negative number and turns it into a positive which would break this function in certain scenarios.

For example: `console.log(isProperFraction(-5, 3));` would incorrectly return `false` when it is a proper fraction.

Do you know how you could fix this bug? Feel free to message me if you need a hand

BenSymons (commented):
A few suggestions and some small mistakes here and there but for the most part, well done!

Sprint-2/2-mandatory-debug/1.js
eyuell21: Noted. Thanks.

Sprint-2/3-mandatory-implement/3-to-pounds.js
eyuell21: Done. This method is very simple and amazing. Thankyou!

Sprint-2/3-mandatory-implement/3-to-pounds.js
eyuell21: Okay. Thankyou!

Sprint-2/4-mandatory-interpret/time-format.js
eyuell21: We can use a global variable, which is increased each time the function is called.(got it from stackoverflow).

var myFuncCalls = 0;

function pad(num) {
  myFuncCalls++;
  console.log( "I have been called " + myFuncCalls + " times" );
  return num.toString().padStart(2, "0");
  
}

Sprint-2/4-mandatory-interpret/time-format.js
eyuell21: Noted, Thankyou for the review.

Sprint-3/1-key-implement/1-get-angle-type.js
eyuell21: This commit and the one after this were not supposed to be on this Pull request as they are from sprint three they are here by accident. I'll have to redo them with other the sprint 3 question on a new branch and a new PR for sprint 3.

halilibrahimcelik:
Well done you can close this PR now

PR #274, opened 2025-02-11
ITP-JAN-25 | London | Ihor Bolzhelarskyi | Module-Structuring-and-Testing-Data |Sprint 2| Week 4

IhorBolzhelarskyi:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -1,5 +1,5 @@
 // Predict and explain first...
-//  =============> write your prediction here
+//  =============> write your prediction here  Name of parameter of our function is the same like a name of variable which we are going to return, so we will get an error
 
 // call the function capitalise with a string input
 // interpret the error message and figure out why an error is occurring
@@ -9,5 +9,13 @@ function capitalise(str) {
   return str;
 }
 
-// =============> write your explanation here
+console.log(capitalise("hanan"));
+// =============> write your explanation here we have to change a name of variable or parameter
 // =============> write your new code here
+
+function capitalise(str) {
+  let string = `${str[0].toUpperCase()}${str.slice(1)}`;
+  return string;
+}
+
+console.log(capitalise("hanan"));
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -15,5 +15,10 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
+  const result = +(weight / (height * height)).toFixed(1);
+  return result;
+
+  // return the BMI of someone based off their weight and height
 }
+
+console.log(calculateBMI(22, 1.53));
--- a/Sprint-2/5-stretch-extend/format-time.js
+++ b/Sprint-2/5-stretch-extend/format-time.js
@@ -4,22 +4,34 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
-  if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+  const minutes = Number(time.slice(3, 5));
+  if (hours == 0) {
+    return `${hours + 12}:${minutes.toString().padStart(2, "0")} am`;
+  } else if (hours == 12) {
+    return `${hours}:${minutes.toString().padStart(2, "0")} pm`;
+  } else if (hours > 12) {
+    return `${hours - 12}:${minutes.toString().padStart(2, "0")} pm`;
   }
-  return `${time} am`;
+  return `${hours}:${minutes.toString().padStart(2, "0")} am`;
 }
 
-const currentOutput = formatAs12HourClock("08:00");
-const targetOutput = "08:00 am";
+const currentOutput = formatAs12HourClock("08:07");
+const targetOutput = "8:07 am";
 console.assert(
   currentOutput === targetOutput,
   `current output: ${currentOutput}, target output: ${targetOutput}`
 );
 
-const currentOutput2 = formatAs12HourClock("23:00");
-const targetOutput2 = "11:00 pm";
+const currentOutput2 = formatAs12HourClock("00:00");
+const targetOutput2 = "12:00 am";
 console.assert(
   currentOutput2 === targetOutput2,
   `current output: ${currentOutput2}, target output: ${targetOutput2}`
+);
+
+const currentOutput3 = formatAs12HourClock("12:09");
+const targetOutput3 = "12:09 pm";
+console.assert(
+  currentOutput3 === targetOutput3,
+  `current output: ${currentOutput3}, target output: ${targetOutput3}`
 );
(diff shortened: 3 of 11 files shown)


Sprint-2/1-key-errors/0.js
```diff
@@ -9,5 +9,13 @@ function capitalise(str) {
   return str;
 }
 
-// =============> write your explanation here
+console.log(capitalise("hanan"));
+// =============> write your explanation here we have to change a name of variable or parameter
 // =============> write your new code here
+
+function capitalise(str) {
+  let string = `${str[0].toUpperCase()}${str.slice(1)}`;
```
illicitonion, line 17: I don't think it's super clear what the difference is between a variable named `str` and a variable named `string`.

Can you think of how we could make it more clear what the difference is between these variables?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,10 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+  const result = +(weight / (height * height)).toFixed(1);
+  return result;
```
illicitonion, line 19: This works, but I've noticed you're making a variable here that you're only using once.

That can be useful, because the name could be clear for what it is, but I'm not sure the name "result" is very clarifying.

You could instead simplify this down into one line:

```suggestion
  return +(weight / (height * height)).toFixed(1);
```

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,22 +4,34 @@
...
   const hours = Number(time.slice(0, 2));
-  if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+  const minutes = Number(time.slice(3, 5));
+  if (hours == 0) {
+    return `${hours + 12}:${minutes.toString().padStart(2, "0")} am`;
+  } else if (hours == 12) {
+    return `${hours}:${minutes.toString().padStart(2, "0")} pm`;
+  } else if (hours > 12) {
+    return `${hours - 12}:${minutes.toString().padStart(2, "0")} pm`;
```
illicitonion, line 13: This works, but there's a lot of repeated code (e.g. that we're deciding to use `:`s between parts, the padding, etc).

Can you think how to write this where we only call `padStart` once, we only add a `:` once, etc?

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,22 +4,34 @@
...
   }
-  return `${time} am`;
+  return `${hours}:${minutes.toString().padStart(2, "0")} am`;
 }
 
-const currentOutput = formatAs12HourClock("08:00");
-const targetOutput = "08:00 am";
+const currentOutput = formatAs12HourClock("08:07");
+const targetOutput = "8:07 am";
```
illicitonion, line 19: Good job finding a lot of edge cases for this code!

illicitonion (commented):
This is generally looking really good, congrats!

Please do updated the code with the suggestions I've made and re-add the "Needs Review" label :)

Sprint-2/1-key-errors/0.js
IhorBolzhelarskyi: The variable str is a parameter of function capitalise(str) (input of this function). It holds a original string passed into the function. The variable string is lacal variable in this function, that stores modified version of parameter str, wich we are going to return (output of this function).

Sprint-2/5-stretch-extend/format-time.js
IhorBolzhelarskyi: We can assign the minutes variable the value  `:${Number(time.slice(3, 5)).toString().padStart(2, "0")}`

Sprint-2/1-key-errors/0.js
illicitonion: I can work that out by reading the function, but I don't think it's very clear at a glance.

Imagine if this function was much longer, and you used `string` at some point inside it - would it be obvious whether it was the original string or the fixed string?

In general if you have two similar variables, I'd recommend changing _both_ of their names to make clear what each  one is.

Sprint-2/5-stretch-extend/format-time.js
illicitonion: That's definitely simpler, thank you!

But I think it's still not quite clear what we _always_ do, and what we _sometimes_ do.

I would argue we _always_:

Put hours, then a colon, then minutes, then a space, then an indicator of morning/evening.

So I'd suggest you end up with a single `return` statement which looks like:

```javascript
return ${hours}:${minutes} ${ampm};
```

and set up the code before this in your `if`s so that this works.

It's really useful when we have conditional code to know what is always the same, and what is different between each case. Right now, because you have really different things in each `return`, we need to read them all carefully to look for small differences.

Sprint-2/1-key-errors/0.js
IhorBolzhelarskyi: Yes,i got this, their names are almost the same
Thank you

Sprint-2/5-stretch-extend/format-time.js
IhorBolzhelarskyi: I think I found a solution to simplify the function

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -3,23 +3,34 @@
...
-    return `${hours - 12}:00 pm`;
-  }
-  return `${time} am`;
+  let hours = Number(time.slice(0, 2));
+  const minutes = time.slice(3, 5);
+  const ampm = hours >= 12 ? `pm` : `am`;
+
+  if (hours === 0) {
+    hours = 12;
+  } else if (hours > 12) hours = hours - 12;
```
illicitonion, line 12: I'd recommend always using `{}`s around the body of `if`/`else` statements. As well as consistency (which is good), you can read more about some of the problems not using `{}`s causes at https://www.blackduck.com/blog/understanding-apple-goto-fail-vulnerability-2.html

Sprint-2/5-stretch-extend/format-time.js
illicitonion: Looks good! Good job!

Sprint-2/1-key-errors/0.js
illicitonion: Can you rename them to be more clear? :)

IhorBolzhelarskyi: Yes😅

Sprint-2/1-key-errors/0.js
```diff
@@ -14,8 +14,8 @@ console.log(capitalise("hanan"));
 // =============> write your new code here
 
 function capitalise(str) {
-  let string = `${str[0].toUpperCase()}${str.slice(1)}`;
-  return string;
+  let modifiedStr = `${str[0].toUpperCase()}${str.slice(1)}`;
```
illicitonion, line 17: I'd generally recommend updating _both_ variables - `originalStr` and `modifiedStr` - otherwise if someone just sees `str` they may still have the question "is this the old one or the new one" :)

IhorBolzhelarskyi: yes, that makes sense, thank you

PR #283, opened 2025-02-13
LONDON-JAN-25 | ANDREI FILIPPOV | Module-Structuring-and-Testing-Data | WEEK 2

Droid-An:
## Learners, PR Template

Self checklist

## Changelist

Completed Sprint 2 Coursework

## Questions

no questions

### DIFF
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -1,13 +1,22 @@
 // Predict and explain first...
 //  =============> write your prediction here
+// I don't see any error in this code 
 
 // call the function capitalise with a string input
 // interpret the error message and figure out why an error is occurring
 
-function capitalise(str) {
-  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
-  return str;
-}
+// function capitalise(str) {
+//   let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+//   return str;
+// }
+// console.log(capitalise('cat'))
 
 // =============> write your explanation here
+// we cannot use js reserved keywords as a function name
+
 // =============> write your new code here
+function capitalise(str) {
+  let capitalised_str = `${str[0].toUpperCase()}${str.slice(1)}`;
+  return capitalised_str;
+}
+console.log(capitalise('cat'))
--- a/Sprint-2/1-key-errors/1.js
+++ b/Sprint-2/1-key-errors/1.js
@@ -3,18 +3,31 @@
 // Why will an error occur when this program runs?
 // =============> write your prediction here
 
+// Error occurs because console.log is trying to show a variable, that wasn't declared globally 
+
 // Try playing computer with the example to work out what is going on
 
-function convertToPercentage(decimalNumber) {
+// function convertToPercentage(decimalNumber) {
+//   const decimalNumber = 0.5;
+//   const percentage = `${decimalNumber * 100}%`;
+
+//   return percentage;
+// }
+
+// console.log(decimalNumber);
+
+// =============> write your explanation here
+// after running the program I understood that program is not working because we declare the const 'decimalNumber' which was already declared with the function. 
+// Also the code is trying to show 'decimalNumber', which is local variable. What we want to show is a return of 'convertToPercentage' function 
+
+// Finally, correct the code to fix the problem
+// =============> write your new code here
+
+function convertToPercentage(decNumber) {
   const decimalNumber = 0.5;
   const percentage = `${decimalNumber * 100}%`;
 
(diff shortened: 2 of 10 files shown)


Sprint-2/1-key-errors/0.js
```diff
@@ -1,13 +1,22 @@
...
-  return str;
-}
+// function capitalise(str) {
+//   let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+//   return str;
+// }
+// console.log(capitalise('cat'))
 
 // =============> write your explanation here
+// we cannot use js reserved keywords as a function name
```
cjyuan, line 15: Is this the cause of the error? I don't think `capitalise` is a keyword in JS.

Sprint-2/1-key-errors/1.js
```diff
@@ -3,18 +3,31 @@
...
 // =============> write your new code here
+
+function convertToPercentage(decNumber) {
+  const decimalNumber = 0.5;
+  const percentage = `${decimalNumber * 100}%`;
+
+  return percentage;
+}
+
+console.log(convertToPercentage());
```
cjyuan, line 33: With the current implementation, the function will always return "50%".
How would you modify the function so that it can be easily reused to convert different numbers to their equivalent percentages?

Sprint-2/2-mandatory-debug/0.js
```diff
@@ -1,14 +1,23 @@
...
+// console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
 
 // =============> write your explanation here
 
+
 // Finally, correct the code to fix the problem
 //  =============> write your new code here
+function multiply(a, b) {
+  console.log(a * b);
+}
```
cjyuan, line 21: This function is the same as the original one.

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -16,4 +16,7 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    return (weight / (height*height)).toFixed(1)
```
cjyuan, line 19: Should this function return a number or a string?

What if someone wants to calculate the BMI correct to two decimal places?

cjyuan (commented):
Looks good mostly. I only have a few comments and suggestions.

Sprint-2/1-key-errors/0.js
Droid-An: Oh, seems like I mixed up a bit. In this case the problem was with the variable called "str"

Sprint-2/3-mandatory-implement/1-bmi.js
Droid-An: The function should return number, but since it only manipulates numbers, there's no necessity to convert type

cjyuan: Have you checked if `calculateBMI()` is returning a value of type number? 

```
// Both of these statements would output 100 in the console.
console.log(100);
console.log("100");
```

cjyuan (commented):
Just two more comments.

Feel free to respond via comments in your PR. That way I will get notified via email of your response.

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -16,7 +16,7 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
-    return (weight / (height*height)).toFixed(1)
+    return (weight / (height*height)).toFixed(2)
```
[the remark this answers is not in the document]
Droid-An: I checked the data type using the `typeof` operator, and it showed that my return value was a string. To fix this, I used the `Number()` function to convert the string into a number.

Sprint-2/1-key-errors/0.js
```diff
@@ -5,18 +5,18 @@
...
-// console.log(capitalise('cat'))
+function capitalise(str) {
+  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+  return str;
+}
+console.log(capitalise('cat'))
 
 // =============> write your explanation here
-// we cannot use js reserved keywords as a function name
+// we cannot use js reserved keywords as a variable name
```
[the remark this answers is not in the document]
Droid-An: I usually try to use proper terms, and at the time, I probably thought the name was correct. After taking a fresh look at the code, I realized that the problem wasn’t with the name itself, but with iteration of it.

cjyuan:
Changes look good. Well done!

PR #284, opened 2025-02-13
ITP25LDN | Eyuel Abraham | Module-Structuring-and-Testing-Data | Week3 

eyuell21:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.
Completed the exercises on this coursework.

## Questions

### DIFF
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -13,8 +13,8 @@ function getAngleType(angle) {
   if (angle > 90 && angle < 180) return "Obtuse angle"
   if (angle === 180) return "Straight angle"
   if (angle > 180 && angle < 360) return "Reflex angle"
-  if(angle === 360)  return "Full angle"
-  else return "please insert a valid angle (0-360)"
+  if (angle === 360)  return "Full angle"
+  return "please insert a valid angle (0-360)"
    // read to the end, complete line 36, then pass your test here
 }
 


Sprint-3/1-key-implement/1-get-angle-type.js
```diff
@@ -8,7 +8,13 @@
...
-   if (angle === 90) return "Right angle";
+  if (angle === 90) return "Right angle";
+  if (angle < 90) return "Acute angle"
+  if (angle > 90 && angle < 180) return "Obtuse angle"
+  if (angle === 180) return "Straight angle"
+  if (angle > 180 && angle < 360) return "Reflex angle"
+  if(angle === 360)  return "Full angle"
+  else return "please insert a valid angle (0-360)"
    // read to the end, complete line 36, then pass your test here
 }
```
halilibrahimcelik, line 19: it is   good to  handle your  logic you can also use else if to  combine your logic and in the last result always return relavant message that would fit context of your function

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
```diff
@@ -1,5 +1,27 @@
...
+    }
+    if (lastDigit == 2) {
+        return num + "nd";
+    }
+    if (lastDigit == 3) {
+        return num + "rd";
+    }
+
+    return num + "th";
 }
```
halilibrahimcelik, line 17: Your code is working, it's good but we can also enhance this logic by checking types of arguments as well for instance what if we pass getOrdinalNumber("22") strings. consider this as a challange

Sprint-3/4-stretch-investigate/password-validator.js
```diff
@@ -1,6 +1,25 @@
...
+//     return password.length < 5 ? false : true
+// }
+
+const passwordRequirements = /^(?=.*[A-Z])(?=.*[a-z])(?=.*\d)(?=.*[\W_]).{5,20}$/;
+
+let passwordsUsed = [];
+
 function passwordValidator(password) {
-    return password.length < 5 ? false : true
+    if(passwordRequirements.test(password) && !passwordsUsed.includes(password)){
```
halilibrahimcelik, line 10: Regex is an essential way to handle your this kind string validation, here is a challange, instead of relying on regex how can we make sure this passwordValidation

halilibrahimcelik (approved):

Sprint-3/4-stretch-investigate/password-validator.js
eyuell21: We can store our requirements in an array and we can compare each element of our given password with the array and return valid if it satisfies our requirements.

halilibrahimcelik: I would like to see in action, if it possible just write  your new logic  but this time without regex :) and commit your changes

PR #296, opened 2025-02-16
ITP-JAN-25 | Sulaiman Kamara | London Brixton Office | Module-Structuring-and-Testing-Data | Sprint 1, Issue 3

sul-devs:
Completed Sprint 1 Exercises.

### DIFF
--- a/Sprint-1/1-key-exercises/2-initials.js
+++ b/Sprint-1/1-key-exercises/2-initials.js
@@ -5,12 +5,16 @@ let lastName = "Johnson";
 // Declare a variable called initials that stores the first character of each string.
 // This should produce the string "CKJ", but you must not write the characters C, K, or J in the code of your solution.
 
+// https://www.google.com/search?q=get+first+character+of+string+mdn
+
+// MY SOLUTION START //
+
 // Get first character of each name, store it in a string, and remove spaces with trim().
 let initials = `${firstName.charAt(0)}${middleName.charAt(0)}${lastName.charAt(
   0
-)}`.trim(); 
+)}`.trim();
 
 // View output to check if 'initals' variable matches the expected output 'CKJ'.
 console.log(initials);
 
-// https://www.google.com/search?q=get+first+character+of+string+mdn
+// MY SOLUTION END //
--- a/Sprint-1/1-key-exercises/2-initials.js
+++ b/Sprint-1/1-key-exercises/2-initials.js
@@ -12,7 +12,7 @@ let lastName = "Johnson";
 // Get first character of each name, store it in a string, and remove spaces with trim().
 let initials = `${firstName.charAt(0)}${middleName.charAt(0)}${lastName.charAt(
   0
-)}`.trim();
+)}`;
 
 // View output to check if 'initals' variable matches the expected output 'CKJ'.
 console.log(initials);
--- a/Sprint-1/1-key-exercises/2-initials.js
+++ b/Sprint-1/1-key-exercises/2-initials.js
@@ -5,7 +5,12 @@ let lastName = "Johnson";
 // Declare a variable called initials that stores the first character of each string.
 // This should produce the string "CKJ", but you must not write the characters C, K, or J in the code of your solution.
 
-let initials = ``;
+// Get first character of each name, store it in a string, and remove spaces with trim().
+let initials = `${firstName.charAt(0)}${middleName.charAt(0)}${lastName.charAt(
+  0
+)}`.trim(); 
+
+// View output to check if 'initals' variable matches the expected output 'CKJ'.
+console.log(initials);
 
 // https://www.google.com/search?q=get+first+character+of+string+mdn
-
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -11,7 +11,7 @@
 
 const filePath = "/Users/mitch/cyf/Module-JS1/week-1/interpret/file.txt";
 const lastSlashIndex = filePath.lastIndexOf("/");
-const indexOfDot = filePath.lastIndexOf("."); // Get index of '.'
+const indexOfDot = filePath.lastIndexOf("."); // Get last index of '.'
 const base = filePath.slice(lastSlashIndex + 1);
 const dir = filePath.slice(0, lastSlashIndex); // 1. Create a variable to store the dir part of the filePath variable
 const ext = filePath.slice(indexOfDot); // 2. Create a variable to store the ext part of the variable
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -11,13 +11,14 @@
 
 const filePath = "/Users/mitch/cyf/Module-JS1/week-1/interpret/file.txt";
 const lastSlashIndex = filePath.lastIndexOf("/");
+const indexOfDot = filePath.indexOf("."); // Get index of '.'
 const base = filePath.slice(lastSlashIndex + 1);
-console.log(`The base part of ${filePath} is ${base}`);
+const dir = filePath.slice(0, lastSlashIndex); // 1. Create a variable to store the dir part of the filePath variable
+const ext = filePath.slice(indexOfDot); // 2. Create a variable to store the ext part of the variable
 
-// Create a variable to store the dir part of the filePath variable
-// Create a variable to store the ext part of the variable
+// Output
+console.log(`1. Dir - "${dir}"`);
+console.log(`2. Ext - "${ext}"`);
+console.log(`3. The base part of "${filePath}" is "${base}"`);
 
-const dir = ;
-const ext = ;
-
-// https://www.google.com/search?q=slice+mdn
+// Resource - https://www.google.com/search?q=slice+mdn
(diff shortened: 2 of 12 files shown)


Sprint-1/1-key-exercises/2-initials.js
```diff
@@ -5,7 +5,16 @@ let lastName = "Johnson";
...
-let initials = ``;
-
 // https://www.google.com/search?q=get+first+character+of+string+mdn
 
+// MY SOLUTION START //
+
+// Get first character of each name, store it in a string, and remove spaces with trim().
+let initials = `${firstName.charAt(0)}${middleName.charAt(0)}${lastName.charAt(
+  0
+)}`.trim();
```
cjyuan, line 15: What's the purpose of using `.trim()` when constructing this string?

Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -11,13 +11,14 @@
 
 const filePath = "/Users/mitch/cyf/Module-JS1/week-1/interpret/file.txt";
 const lastSlashIndex = filePath.lastIndexOf("/");
+const indexOfDot = filePath.indexOf("."); // Get index of '.'
 const base = filePath.slice(lastSlashIndex + 1);
-console.log(`The base part of ${filePath} is ${base}`);
+const dir = filePath.slice(0, lastSlashIndex); // 1. Create a variable to store the dir part of the filePath variable
+const ext = filePath.slice(indexOfDot); // 2. Create a variable to store the ext part of the variable
```
cjyuan, line 17: Since a folder name can also contain '.', can you modify your code so that it will also work for a file path like `"/path1/path2.old/file.txt"`?

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -7,3 +7,13 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+// DOCUMENTATION
+// 'num' - generates a random number within range between 'minimum' and 'maximum'.
+// Math.floor() - rounds any number down to the nearest whole number.
+// Math.random() - generates a random number between 0 and 0.9 recurring (never 1).
+// '* (maximum - minimum + 1)' - Calculates range. '+ 1' means generated number can actually be <= range instead of just < range.
+// '+ minimum' - totals the minimum and maximum to get the total range.
```
cjyuan, line 16: The phrase "between X and Y" alone is not precise enough in program specification because 
it does not state clearly whether X and Y are included in the range.

One concise way to specify a range of values is to use the **interval notation**.
For example, you can describe the return value of `Math.random()` as "a random number in [0, 1)".

Sprint-1/2-mandatory-errors/3.js
```diff
@@ -1,4 +1,4 @@
-const cardNumber = 4533787178994213;
+const cardNumber = [4, 5, 3, 3, 7, 8, 7, 1, 7, 8, 9, 9, 4, 2, 1, 3];
```
cjyuan, line 1: If you cannot modify this statement `const cardNumber = 4533787178994213;`
(that is, keep the variable's value unchanged), 
how would you modify the code (through type conversion) so that it can still extract the last 4 digits from its value.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -13,10 +13,20 @@ console.log(`The percentage change is ${percentageChange}`);
 
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
 
+// 4 function calls, in lines 4, 5, 8, and 10.
```
cjyuan, line 16: There are 5 function calls. Can you identify them?

Note: `foo(bar())` would be considered as two function calls in which `bar()` is called first and its return value is passed as a parameter to `foo()`.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -13,10 +13,20 @@ console.log(`The percentage change is ${percentageChange}`);
...
 
+// In lines 4 and 5, 'priceDifference' and 'priceAfterOneYear' are reassigned.
+
 // d) Identify all the lines that are variable declarations
 
+// 1, 2, 7, 8 are all variable declarations, and 4 and 5 are variable redeclarations.
+
 // e) Describe what the expression Number(carPrice.replaceAll(",","")) is doing - what is the purpose of this expression?
+
+// Type conversion: converts values from string to number, in order to be used in calculations later on in the program.
```
cjyuan, line 32: This only describes what `Number(...)` does. What about `carPrice.replaceAll(",","")`?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -13,13 +13,25 @@ console.log(result);
...
 
+// Answer: It sees how many whole minutes go into the movieLength, and tells you what was left over.
+
 // d) Interpret line 4, what does the expression assigned to totalMinutes mean?
 
+// Answer: It gets the total length of the movie in seconds, divides it by 60 to get the total length of the movie in minutes.
+
 // e) What do you think the variable result represents? Can you think of a better name for this variable?
 
+// Answer: It represents the current time in the movie, so maybe 'timestamp' could be a better name.
```
cjyuan, line 33: The value assigned to `result` is not "the current time in the movie". So **timestamp** is not quite appropriate.
Can you think of another name?

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -25,3 +25,8 @@ console.log(`£${pounds}.${pence}`);
 
 // To begin, we can start with
 // 1. const penceString = "399p": initialises a string variable with the value "399p"
+// 2. penceStringWithoutTrailingP: separates the number from the 'p', so it's now "399"
+// 3. paddedPenceNumberString: adds a 0 to the start, so it's "0399"
```
cjyuan, line 29: Does `"399".padStart(3, "0")` yield "0399"?

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -25,3 +25,8 @@ console.log(`£${pounds}.${pence}`);
...
+// 4. pounds: extracts the first 2 characters to get the pounds, so it's "03"
+// 5. pence: gets all the characters after 2nd character, so it's "99". The 'padEnd(2)' makes sure there is always two digits for the pence part
```
cjyuan, line 31: These descriptions are not quite correct. What if the original value of `penceString` is "123456p"?

Also, can we expect this program to work as intended even if we deleted `.padEnd(2, "0")` from the code?

cjyuan (commented):
Look mostly good. I only have some suggestions.

Sprint-1/1-key-exercises/2-initials.js
sul-devs: > What's the purpose of using `.trim()` when constructing this string?

No reason, I was trying to do something else before that would've used the trim() function, but forgot to remove it. I've removed it now.

Sprint-1/1-key-exercises/3-paths.js
sul-devs: I changed the method used in 'indexOfDot' variable to use lastIndexOf() method instead of indexOf()

Sprint-1/1-key-exercises/4-random.js
sul-devs: I'm not really sure what to do here. Can you clarify what I should do a bit more please?

Sprint-1/2-mandatory-errors/3.js
sul-devs: In 'cardNumberString', I converted 'cardNumber' to a string, and used '.substring()' method to extract the last 4 digits.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
sul-devs: Updated it, I found 5 now.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
sul-devs: I've added an explanation now.

Sprint-1/3-mandatory-interpret/2-time-format.js
sul-devs: From my understanding, the 'result' represents the total movie time in hours, minutes, and seconds, so I thought 'duration' would be a good name to represent the 'result' variable.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
sul-devs: Does it yield "000399"? From my understanding, the first argument '3' is the number of times the second argument '0' should be added to the start, based on the example in this documentation (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart), but I wasn't too clear on how it works in the program.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
sul-devs: Need a bit of help on this. Can I book a pair programming session with you?

PR #304, opened 2025-02-18
ITP-May 2025 | Chi Mbah | Module-Structuring-and-Testing-Data | Week 5 | Sprint 2

Chi-Mb:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -15,5 +15,9 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
+
+    let BMI = weight / (height * height);
+    return BMI.toFixed(1);
 }
+
+console.log(calculateBMI(70, 1.73));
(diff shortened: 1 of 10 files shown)


Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,9 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+
+    let BMI = weight / (height * height);
+    return BMI.toFixed(1);
+}
```
cjyuan, line 21: What type of value do you expect the function to return? A number or a string?
Is your implementation matches your expectation?

cjyuan (commented):
Looks good!

Why not try also `Sprint-2/5-stretch-extend`?

Sprint-2/3-mandatory-implement/1-bmi.js
Chi-Mb: @cjyuan I expect the function to return a string since .toFixed(1) converts the number to one decimal place. My implementation doesn’t match my expectation because the BMI calculation is outside the function, making it incorrect.

cjyuan: I meant to ask, what type of value do you think the function should return, not what it will return according to how you implement it.

PR #305, opened 2025-02-18
ITP-JAN-25 | London | Ihor Bolzhelarskyi | Module-Structuring-and-Testing-Data | Sprint 3 | Week 5

IhorBolzhelarskyi:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/1-key-implement/3-get-card-value.js
+++ b/Sprint-3/1-key-implement/3-get-card-value.js
@@ -8,9 +8,30 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
-    if (rank === "A") return 11;
+  const nosuit = card.slice(0, -1);
+  if (nosuit === "A") {
+    return 11;
+  } else if (nosuit === "Q") {
+    return 10;
+  } else if (nosuit === "K") {
+    return 10;
+  } else if (nosuit === "J") {
+    return 10;
+  } else if (
+    nosuit === "2" ||
+    nosuit === "3" ||
+    nosuit === "4" ||
+    nosuit === "5" ||
+    nosuit === "6" ||
+    nosuit === "7" ||
+    nosuit === "8" ||
+    nosuit === "9"
+  ) {
+    return parseInt(nosuit);
+  } else {
+    return "Invalid card rank.";
+  }
 }
-
 // You need to write assertions for your function to check it works in different cases
 // we're going to use this helper function to make our assertions easier to read
 // if the actual output matches the target output, the test will pass
@@ -28,24 +49,32 @@ function assertEquals(actualOutput, targetOutput) {
 const aceofSpades = getCardValue("A♠");
 assertEquals(aceofSpades, 11);
 
-// Handle Number Cards (2-10):
-// Given a card with a rank between "2" and "9",
-// When the function is called with such a card,
-// Then it should return the numeric value corresponding to the rank (e.g., "5" should return 5).
+// // Handle Number Cards (2-10):
+// // Given a card with a rank between "2" and "9",
+// // When the function is called with such a card,
+// // Then it should return the numeric value corresponding to the rank (e.g., "5" should return 5).
 const fiveofHearts = getCardValue("5♥");
+assertEquals(fiveofHearts, 5);
+
 // ====> write your test here, and then add a line to pass the test in the function above
 
 // Handle Face Cards (J, Q, K):
 // Given a card with a rank of "10," "J," "Q," or "K",
 // When the function is called with such a card,
 // Then it should return the value 10, as these cards are worth 10 points each in blackjack.
+const jackofSpace = getCardValue("J♠");
+assertEquals(jackofSpace, 10);
 
 // Handle Ace (A):
 // Given a card with a rank of "A",
 // When the function is called with an Ace,
 // Then it should, by default, assume the Ace is worth 11 points, which is a common rule in blackjack.
+const aceOfHearts = getCardValue("A♥");
+assertEquals(aceOfHearts, 11);
 
 // Handle Invalid Cards:
 // Given a card with an invalid rank (neither a number nor a recognized face card),
 // When the function is called with such a card,
 // Then it should throw an error indicating "Invalid card rank."
+const invalidCard = getCardValue("1♥");
+assertEquals(invalidCard, "Invalid card rank.");
--- a/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
+++ b/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
@@ -5,7 +5,16 @@ test("should return true for a proper fraction", () => {
 });
 
 // Case 2: Identify Improper Fractions:
+test("should return false for improper fraction", () => {
+  expect(isProperFraction(10, 3)).toEqual(false);
+});
 
 // Case 3: Identify Negative Fractions:
+test("Should return true if absolute value numerator>denominator and false if absolute numerator>denominator (without minus) in this case return false", () => {
+  expect(isProperFraction(-4, 3)).toEqual(false);
+});
 
 // Case 4: Identify Equal Numerator and Denominator:
+test("should return false for Equal Numerator and Denominator", () => {
+  expect(isProperFraction(7, 7)).toEqual(false);
+});
--- a/Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
+++ b/Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
@@ -1,11 +1,23 @@
 const getCardValue = require("./3-get-card-value");
 
 test("should return 11 for Ace of Spades", () => {
-    const aceofSpades = getCardValue("A♠");
-    expect(aceofSpades).toEqual(11);
-    });
+  const aceofSpades = getCardValue("A♠");
+  expect(aceofSpades).toEqual(11);
+});
 
 // Case 2: Handle Number Cards (2-10):
+test("should return 2 for 2♠", () => {
+  expect(getCardValue("2♠")).toEqual(2);
+});
 // Case 3: Handle Face Cards (J, Q, K):
+test("should return 10 for J♠", () => {
+  expect(getCardValue("J♠")).toEqual(10);
+});
 // Case 4: Handle Ace (A):
+test("should return 11 for A", () => {
+  expect(getCardValue("A♥")).toEqual(11);
+});
 // Case 5: Handle Invalid Cards:
+test("should return Invalid card rank. for invalid card (for example F♥),", () => {
+  expect(getCardValue("F♥")).toEqual("Invalid card rank.");
+});
--- a/Sprint-3/3-mandatory-practice/implement/count.test.js
+++ b/Sprint-3/3-mandatory-practice/implement/count.test.js
@@ -22,3 +22,9 @@ test("should count multiple occurrences of a character", () => {
 // And a character char that does not exist within the case-sensitive str,
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of the char were found in the case-sensitive str.
+test("should return 0 if character does not exist in this string", () => {
+  const str = "Hello world";
+  const char = "m";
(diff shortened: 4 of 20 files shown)


Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
```diff
@@ -5,7 +5,16 @@ test("should return true for a proper fraction", () => {
 });
 
 // Case 2: Identify Improper Fractions:
+test("should return false for improper fraction", () => {
+  expect(isProperFraction(10, 3)).toEqual(false);
+});
 
 // Case 3: Identify Negative Fractions:
+test("Should return true if absolute value numerator>denominator and false if absolute numerator>denominator (without minus) in this case return false", () => {
```
cjyuan, line 13: 1. What is the difference between "absolute value numerator>denominator" and "absolute numerator>denominator (without minus)"? Why do you expect the former is true but the latter is false?

2. The description seems to suggest multiple different cases but there is only one test case specified.

3. What should be the return value of the following function calls?
```
isProperFraction(-2, -3)
isProperFraction(-2, 3)
isProperFraction(2, -3)
isProperFraction(-4, -3)
isProperFraction(4, -3)
```

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -1,11 +1,23 @@
...
-    expect(aceofSpades).toEqual(11);
-    });
+  const aceofSpades = getCardValue("A♠");
+  expect(aceofSpades).toEqual(11);
+});
 
 // Case 2: Handle Number Cards (2-10):
+test("should return 2 for 2♠", () => {
+  expect(getCardValue("2♠")).toEqual(2);
+});
```
cjyuan, line 11: If you test all ranks from 2-10, you may **find a bug in your code. :)** 

You can easily use a for-loop to test all values from 2-10.

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -1,11 +1,23 @@
...
 // Case 3: Handle Face Cards (J, Q, K):
+test("should return 10 for J♠", () => {
+  expect(getCardValue("J♠")).toEqual(10);
```
cjyuan, line 14: The function works for "J" does not necessary guarantee it works also for "Q" and "K".
So it is better to include also "Q" and "K" in the test cases.

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -1,11 +1,23 @@
...
+});
 // Case 4: Handle Ace (A):
+test("should return 11 for A", () => {
+  expect(getCardValue("A♥")).toEqual(11);
+});
 // Case 5: Handle Invalid Cards:
+test("should return Invalid card rank. for invalid card (for example F♥),", () => {
```
cjyuan, line 21: There is no need to specify "(for example F♥)," in the description. The test values shown in the test cases are the examples.

Sprint-3/3-mandatory-practice/implement/count.test.js
```diff
@@ -22,3 +22,9 @@ test("should count multiple occurrences of a character", () => {
 // And a character char that does not exist within the case-sensitive str,
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of the char were found in the case-sensitive str.
+test("should return 0 if character does not exist in this string", () => {
+  const str = "Hello world";
+  const char = "m";
+  const count = countChar(str, char);
+  expect(count).toEqual(0);
```
cjyuan, line 29: You don't have to follow the format given in the example.
You can also express the test as `expect(countChar("Hello world", "m")).toEqual(0)` if you think it is easier to read.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
```diff
@@ -1,5 +1,9 @@
 function getOrdinalNumber(num) {
- return "1st";
+  if (num === 1) return "1st";
+  else if (num === 2) return "2nd";
+  else if (num === 3) return "3rd";
+  else if (num >= 4 && num <= 10) return `${num}th`;
+  else return "This function takes number from 1-10!";
```
cjyuan, line 6: Why limit the function to only 1-10? Why not make the function works for all positive integers?

Sprint-3/3-mandatory-practice/implement/repeat.js
```diff
@@ -1,5 +1,7 @@
-function repeat() {
-    return "hellohellohello";
+function repeat(str, count) {
+  if (count === 0) return "";
+  else if (count < 0) return "Please enter positive number!";
+  else return str.repeat(count);
```
cjyuan, line 4: How would the caller distinguish the following two function calls?
1. `repeat("Please enter positive number!", 1)`
2. `repeat("", -1)`

Sprint-3/4-stretch-investigate/find.js
```diff
@@ -19,7 +19,7 @@ console.log(find("code your future", "z"));
...
 // Pay particular attention to the following:
 
-// a) How the index variable updates during the call to find
-// b) What is the if statement used to check
-// c) Why is index++ being used?
-// d) What is the condition index < str.length used for?
+// a) How the index variable updates during the call to find increases by 1 for each loop iteration, stop if str[index] === char
+// b) What is the if statement used to check  Character in parameter str order from [0] to the end to be equal specific character in second parameter char.
+// c) Why is index++ being used? to increment it so loop check each character in the string.
+// d) What is the condition index < str.length used for? true condition while this loop will run
```
cjyuan, line 25: Question (d) is asking, what does the condition `index < str.length` intend to check?

Sprint-3/4-stretch-investigate/password-validator.js
```diff
@@ -1,6 +1,25 @@
+const passwords = [`Qwerty12345$`, "Zxc322!", `Ggwp228*`];
 function passwordValidator(password) {
-    return password.length < 5 ? false : true
+  // prettier-ignore
+  if (passwords.includes(password))
+    return false
+  else if(password.length >= 5 && /[A-Z]/.test(password) && /[a-z]/.test(password)&& /[0-9]/.test(password) && /[!#$%.*&]/.test(password)) return true
+  else return false
```
cjyuan, line 7:
```
if (condition)
  return true
else
  return false
```
can be simplified as `return Boolean(condition);`

Sprint-3/4-stretch-investigate/password-validator.test.js
```diff
@@ -15,12 +15,35 @@ To be valid, a password must:
...
+});
+
+test("invalid password without symbol should return false", () => {
+  // Arrange
+  const password = "Hamburger3";
+  // Act
+  const result = isValidPassword(password);
+  // Assert
+  expect(result).toEqual(false);
+});
```
cjyuan, line 49: What about cases that violate the following requirements:
- Have at least 5 characters.
- Have at least one English uppercase letter (A-Z)
- Have at least one English lowercase letter (a-z)

cjyuan (commented):
Code is mostly good. 
I left you some suggestions.

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -28,24 +49,32 @@ function assertEquals(actualOutput, targetOutput) {
...
 // Then it should, by default, assume the Ace is worth 11 points, which is a common rule in blackjack.
+const aceOfHearts = getCardValue("A♥");
+assertEquals(aceOfHearts, 11);
 
 // Handle Invalid Cards:
 // Given a card with an invalid rank (neither a number nor a recognized face card),
 // When the function is called with such a card,
 // Then it should throw an error indicating "Invalid card rank."
+const invalidCard = getCardValue("1♥");
+assertEquals(invalidCard, "Invalid card rank.");
```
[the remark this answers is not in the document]
IhorBolzhelarskyi: Yes, i have just written tests for not a number conditions, also add throw and catch an error in my code

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
IhorBolzhelarskyi: I made i mistake in description, it should always return false if absolute value numerator>denominator and true if absolute value numerator<denominator. I added more tests

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
IhorBolzhelarskyi: yes, it doesnt contain case 10

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
IhorBolzhelarskyi: fixed that

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
IhorBolzhelarskyi: 👍

Sprint-3/3-mandatory-practice/implement/count.test.js
IhorBolzhelarskyi: okey

Sprint-3/3-mandatory-practice/implement/repeat.js
IhorBolzhelarskyi: We can throw new Error(`Please enter positive number`), instead of return string`Please enter positive number!`

Sprint-3/4-stretch-investigate/find.js
IhorBolzhelarskyi: This condition checks index is still less then str.length during loop run. If condition true loop is running if false loop stoped

Sprint-3/4-stretch-investigate/password-validator.js
IhorBolzhelarskyi: ok

Sprint-3/4-stretch-investigate/password-validator.test.js
```diff
@@ -47,3 +47,9 @@ test("invalid password without symbol should return false", () => {
   // Assert
   expect(result).toEqual(false);
 });
+
+test("password with at least 5 characters and 1 uppercase, but without symbol and number should return false", () => {
+  const password = "bearA";
+  const result = isValidPassword(password);
+  expect(result).toEqual(false);
+});
```
cjyuan, line 55: What I meant was, among all these conditions (which a valid password must satisfy),
1. Have at least 5 characters.
2. Have at least one English uppercase letter (A-Z)
3. Have at least one English lowercase letter (a-z)
4. Have at least one number (0-9)
5. Have at least one of the following non-alphanumeric symbols: ("!", "#", "$", "%", ".", "*", "&")
6.  Must not be any previous password in the passwords array. 

You prepared tests only for conditions 4, 5, 6 and missed tests for conditions 1, 2, 3.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
cjyuan, line 1: The description and the code do not match in some of the test cases.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -13,17 +13,17 @@ test("should return '1st' for 1", () => {
 });
 
 test("should return '2nd' for 2", () => {
-  expect(getOrdinalNumber(2)).toEqual("2nd");
+  expect(getOrdinalNumber(42352)).toEqual("42352nd");
 });
```
cjyuan, line 17: You can also prepare this test case as:
```
test("should append 'nd' to numbers that end in 2 but not end in 12", () => {
  // Code to test some samples (e.g., 2, 22, 122, 42352)
  ...
});
```

cjyuan (commented):
Changes look good.

I have few more suggestions and you can decide if you want to make any change based on the suggestions.

Feel free to mark this PR as completed anytime.

PR #309, opened 2025-02-20
LONDON-JAN-25 | ANDREI FILIPPOV | Module-Structuring-and-Testing-Data | WEEK 3

Droid-An:
## Learners, PR Template

Self checklist

## Changelist

Complete Sprint 3 coursework 

## Questions

no questions

### DIFF
--- a/Sprint-3/1-key-implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-key-implement/2-is-proper-fraction.js
@@ -8,7 +8,9 @@
 // write one test at a time, and make it pass, build your solution up methodically
 
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
+    if (Math.abs(numerator) < Math.abs(denominator)) return true;
+    else if (Math.abs(numerator) > Math.abs(denominator)) return false;
+    else if (Math.abs(numerator) == Math.abs(denominator)) return false;
 }
 
 // here's our helper again
@@ -40,6 +42,7 @@ assertEquals(improperFraction, false);
 // target output: true
 // Explanation: The fraction -4/7 is a proper fraction because the absolute value of the numerator (4) is less than the denominator (7). The function should return true.
 const negativeFraction = isProperFraction(-4, 7);
+assertEquals(negativeFraction, true);
 // ====> complete with your assertion
 
 // Equal Numerator and Denominator check:
@@ -47,7 +50,12 @@ const negativeFraction = isProperFraction(-4, 7);
 // target output: false
 // Explanation: The fraction 3/3 is not a proper fraction because the numerator is equal to the denominator. The function should return false.
 const equalFraction = isProperFraction(3, 3);
+assertEquals(equalFraction, false);
+
 // ====> complete with your assertion
 
 // Stretch:
 // What other scenarios could you test for?
+
+const improperNegativeFraction = isProperFraction(-5, 2);
+assertEquals(improperNegativeFraction, false);
--- a/Sprint-3/1-key-implement/3-get-card-value.js
+++ b/Sprint-3/1-key-implement/3-get-card-value.js
@@ -8,7 +8,16 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
-    if (rank === "A") return 11;
+  rank = card.slice(0, -1);
+  if (rank === "A") return 11;
+  else if (rank === "J" || rank === "Q" || rank === "K") return 10;
+  else if (
+    typeof Number(rank) == "number" &&
+    Number(rank) > 1 &&
+    Number(rank) < 10
+  )
+    return Number(rank);
+  else return "Invalid card rank.";
 }
 
 // You need to write assertions for your function to check it works in different cases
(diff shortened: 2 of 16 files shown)


Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -25,20 +34,23 @@ function assertEquals(actualOutput, targetOutput) {
...
-assertEquals(aceofSpades, 11);
+const aceOfSpades = getCardValue("A♠");
+assertEquals(aceOfSpades, 11);
 
 // Handle Number Cards (2-10):
 // Given a card with a rank between "2" and "9",
 // When the function is called with such a card,
 // Then it should return the numeric value corresponding to the rank (e.g., "5" should return 5).
-const fiveofHearts = getCardValue("5♥");
+const fiveOfHearts = getCardValue("2♥");
```
ChrisCookOC, line 44: I assume this comes from changing it to different values to test it is working, which is a great thing to do. 

Just be careful when you are committing because the name of the variable is now misleading!

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -8,7 +8,16 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
-    if (rank === "A") return 11;
+  rank = card.slice(0, -1);
+  if (rank === "A") return 11;
+  else if (rank === "J" || rank === "Q" || rank === "K") return 10;
+  else if (
+    typeof Number(rank) == "number" &&
```
ChrisCookOC, line 15: This works, but are you able to find any functions you can call that check if something is a number instead? If they exist it is often best to use them and avoid having to code type checks like this!

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -49,3 +61,8 @@ const fiveofHearts = getCardValue("5♥");
 // Given a card with an invalid rank (neither a number nor a recognized face card),
 // When the function is called with such a card,
 // Then it should throw an error indicating "Invalid card rank."
+const invalid = getCardValue("five of hearts");
```
ChrisCookOC, line 64: What should happens if I give you the input of "5.123♥"? Is that reflected in the program?

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
```diff
@@ -1,5 +1,14 @@
 function getOrdinalNumber(num) {
- return "1st";
+  numToString = String(num);
+  if (numToString[numToString.length - 1] == 1 && num != 11) {
```
ChrisCookOC, line 3: You repeat this multiple times
`numToString[numToString.length - 1]`

Can you see the benefits of doing this array access only once? if so, how would you do it?

ChrisCookOC (commented):
You've made good progress here, seems you are getting the hang of unit testing and using jest which is great to see :)

Sprint-3/1-key-implement/2-is-proper-fraction.js
```diff
@@ -8,7 +8,9 @@
 // write one test at a time, and make it pass, build your solution up methodically
 
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
+    if (Math.abs(numerator) < Math.abs(denominator)) return true;
```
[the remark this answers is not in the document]
Droid-An: Oh, now I see how I can simplify that. I just removed last two conditions which are unnecessary

Sprint-3/1-key-implement/3-get-card-value.js
Droid-An: Yes, you're right. I was just short on time to check the correct names for the card values. I've updated it now.

Sprint-3/1-key-implement/3-get-card-value.js
Droid-An: Yes, I found the `isNaN` function, which seems to do exactly that.

Sprint-3/1-key-implement/3-get-card-value.js
Droid-An: Thank you for pointing this out! I hadn’t considered that case. I resolved it by adding an `isInteger` condition.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
Droid-An: Yes, I can assign that value to the `lastDigit` variable and use that variable throughout the code.

Sprint-3/1-key-implement/2-is-proper-fraction.js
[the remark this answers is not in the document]
ChrisCookOC: Yep exactly :)

ChrisCookOC (commented):
Thanks for taking the time to make some more changes :) Looks good.

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -8,7 +8,17 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
-    if (rank === "A") return 11;
+  rank = card.slice(0, -1);
+  if (rank === "A") return 11;
+  else if (rank === "J" || rank === "Q" || rank === "K") return 10;
+  else if (
+    !isNaN(rank) &&
+    Number.isInteger(+(rank)) &&
```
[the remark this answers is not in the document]
Droid-An: Oh, thank you for the remark. Testing showed that `!isNaN` is not necessary here. 
Article about `isInteger` on MDN says:

> If the value is `NaN`, return false.

So `isInteger` do job of both checks.

PR #311, opened 2025-02-20
WM ITP Jan 2025 | HATEF EIDI| Module-Structuring-and-Testing-Data | WEEK3| Sprint 3

HatefEidi:
## Learners, PR Template

Self checklist

- [* ] I have committed my files one by one, on purpose, and for a reason
- [ *] I have titled my PR with COHORT_NAME | FIRST_NAME LAST_NAME | REPO_NAME | WEEK 
- [* ] I have tested my changes
- [ *] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [ *] My changes meet the [requirements](./README.md) of this task

## Changelist

Briefly explain your PR.
In this sprint, we focus of testing our data and file using console.assert and jest. 

## Questions


Sprint-3/1-key-implement/1-get-angle-type.js
```diff
@@ -7,9 +7,19 @@
...
    if (angle === 90) return "Right angle";
    // read to the end, complete line 36, then pass your test here
+
+   else if (angle< 90) return "Acute angle";
+
+   else if(angle >90 && angle<180) return "Obtuse angle";
+
+   else if(angle === 180) return "Straight angle";
+
+   else if(angle > 180 && angle < 360) return "Reflex angle";
```
halilibrahimcelik, line 22: you can also check additional  scenarios like what if number is greater than 360, I know it is not an ideal case for this function but  it would be good challenge for you to return a lets say "out of range" text if number is greater than 360

Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
```diff
@@ -10,15 +10,27 @@ test("should identify right angle (90°)", () => {
...
+{
+  expect(getAngleType(180)).toEqual("Straight angle");  
+});
 
 // Case 5: Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
+test("Should identify Reflex angle (320)", ()=>{
+  expect(getAngleType(320)).toEqual("Reflex angle");
+})
```
halilibrahimcelik, line 36: you can also add a test for what if number is greater than 360, consider this as a challenge

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
```diff
@@ -1,5 +1,41 @@
...
+    // We check if the number is 1, and if the number is 11, then we take the number and add the suffix "th" and return it, but if the number has any last two digits but 11, then we take the number and add the suffix "st" and return it. example: 111 => 111th, 21 => 21st
+     if (theLastDigit === 1) {
+        if(theFirstTwoDigits === 11) {
+            return num + "th";
+        }
+        else{
+            return num + "st";
+        }
+    }
+
```
halilibrahimcelik, line 14: It is good to see you use nesting if check in your logic, but nesting if checks makes your code little bit complicated, try not to nest too much if else statements if it's not necessary.  You can make this code more clean if you first check numbers for 1,2,3 when you get the argument and then continue to  apply your if else logic

halilibrahimcelik (approved):
Looking good, you met all the requirements for this task. It is also good you push your changes with detailed commits. Keep up the good work.

PR #316, opened 2025-02-21
ITP-Jan_25|  Sheetal_Kharab|  Module_structuring_and_data | week_7

sheetalkharab:
## Learners, PR Template

Self checklist

## Changelist

I update in some function and write some test cases

## Questions

### DIFF
--- a/Sprint-3/1-key-implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-key-implement/2-is-proper-fraction.js
@@ -8,7 +8,17 @@
 // write one test at a time, and make it pass, build your solution up methodically
 
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
+    if (numerator < denominator){
+       return true;
+    }else if(numerator>denominator){
+       return false; 
+    }else if(numerator<0 && denominator>0){
+      return true; 
+    }else if(numerator===denominator){
+      return false;
+    }else{
+      return false;
+    }
 }
 
 // here's our helper again
@@ -40,6 +50,7 @@ assertEquals(improperFraction, false);
 // target output: true
 // Explanation: The fraction -4/7 is a proper fraction because the absolute value of the numerator (4) is less than the denominator (7). The function should return true.
 const negativeFraction = isProperFraction(-4, 7);
+assertEquals(negativeFraction, true);
 // ====> complete with your assertion
 
 // Equal Numerator and Denominator check:
@@ -47,7 +58,9 @@ const negativeFraction = isProperFraction(-4, 7);
 // target output: false
 // Explanation: The fraction 3/3 is not a proper fraction because the numerator is equal to the denominator. The function should return false.
 const equalFraction = isProperFraction(3, 3);
+assertEquals(equalFraction, false);
 // ====> complete with your assertion
 
 // Stretch:
 // What other scenarios could you test for?
+// we can check as numerator =5 , denominator=-3
--- a/Sprint-3/1-key-implement/3-get-card-value.js
+++ b/Sprint-3/1-key-implement/3-get-card-value.js
@@ -11,8 +11,8 @@ function getCardValue(card) {
   let rank = card.slice(0, -1);// extract rank everything except last character(suit)
     if (rank === "A") {
       return 11;
-    }else if(!isNaN(rank) && rank>="2" && rank<="9"){
-      return parseInt(rank, 10);
+    }else if(!isNaN(Number(rank)) && Number(rank) >= 2 && Number(rank) <= 10 ){
+      return Number(rank);
     }else if (rank === "K" || rank === "J" || rank === "Q" || rank === "10") {
       return 10;
     }else{
--- a/Sprint-3/1-key-implement/3-get-card-value.js
+++ b/Sprint-3/1-key-implement/3-get-card-value.js
@@ -8,7 +8,17 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
-    if (rank === "A") return 11;
+  let rank = card.slice(0, -1);// extract rank everything except last character(suit)
+    if (rank === "A") {
+      return 11;
+    }else if(!isNaN(rank) && rank>="2" && rank<="9"){
+      return parseInt(rank, 10);
+    }else if (rank === "K" || rank === "J" || rank === "Q" || rank === "10") {
+      return 10;
+    }else{
+      throw new Error("Invalid card rank.");
+    }
+
 }
 
 // You need to write assertions for your function to check it works in different cases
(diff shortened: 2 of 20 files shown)


Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -6,6 +6,27 @@ test("should return 11 for Ace of Spades", () => {
...
+test("should return 5 for Five of Hearts", () => {
+    const fiveofHearts = getCardValue("5♥");
+    expect(fiveofHearts).toEqual(5);
+    });
+
 // Case 3: Handle Face Cards (J, Q, K):
+test("should return 10 for Face Cards", () => {
+    const faceCards = getCardValue("10♦");
+    expect(faceCards).toEqual(10);
+    });
```
cjyuan, line 18: The suggested tests (described in the comment) is for cards J, Q, K.

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -6,6 +6,27 @@ test("should return 11 for Ace of Spades", () => {
     });
 
 // Case 2: Handle Number Cards (2-10):
+test("should return 5 for Five of Hearts", () => {
+    const fiveofHearts = getCardValue("5♥");
+    expect(fiveofHearts).toEqual(5);
+    });
+
```
cjyuan, line 13: Is there a need to limit the tests for a specific suit?

We can also use a for-loop to test all number cards comprehensively from 2 to 10.

Sprint-3/3-mandatory-practice/implement/count.js
```diff
@@ -1,5 +1,5 @@
 function countChar(stringOfCharacters, findCharacter) {
-    return 5
+    return stringOfCharacters.split('').filter (char => char===findCharacter).length;
```
cjyuan, line 2: What is the meaning of `.filter (char => char===findCharacter).length`?

Sprint-3/4-stretch-investigate/creditCardValidator.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+//check if any non numeric digit present in card number return false
+test("card number with non-numeric character should return false", ()=> {
+    expect(validateCardNumber("4992739aa160000a")).toEqual(false);
+});
+
+// if card number have all same digit return false
+test("card number where all digit should be same should return false ", ()=> {
+    expect(validateCardNumber("1111111111111111")).toEqual(false);
+});
```
cjyuan, line 26: This test value may fail to identify if the function can truly return false when the input value contains all the same digit because the function can also return false when the sum of the digits is not greater than 16.

Can you try `expect(validateCardNumber("8888888888888888")).toEqual(false);`

Sprint-3/4-stretch-investigate/find.js
```diff
@@ -20,6 +20,17 @@ console.log(find("code your future", "z"));
...
+*/
 // b) What is the if statement used to check
+/*
+it will check the current char Index  match char.if it match it will return index of that char from String else return -1
+*/
 // c) Why is index++ being used?
+// To increment index value .so we can check whole string and loop can continue till last char of string .
+// every time with index++ ity gave new index for next character in String.
 // d) What is the condition index < str.length used for?
+//It insure loop will stop when our index should be greater then length of string.eg: 17<16 it will stop as string length is 16.
```
cjyuan, line 36: This statement is not quite correct.

Suppose `str.length` is 16 and `char` cannot be found in `str`.
In such case, when the loop stop, `index` is not 17. Can you find out the exact value of `index` when the loop stop?

cjyuan (commented):
There are a few bugs. I left you some comments and suggestions to help you find them.

Sprint-3/1-key-implement/2-is-proper-fraction.js
```diff
@@ -8,7 +8,17 @@
...
+       return true;
+    }else if(numerator>denominator){
+       return false; 
+    }else if(numerator<0 && denominator>0){
+      return true; 
+    }else if(numerator===denominator){
+      return false;
+    }else{
+      return false;
+    }
```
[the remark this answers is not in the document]
sheetalkharab: Thank you for review. yes it will always false as it will check  always after if(numerator< denominator) and else is unnecessary. And i made changes in function and follow your suggestion and use js built in function Math.abs().

Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
```diff
@@ -10,15 +10,30 @@ test("should identify right angle (90°)", () => {
...
+test("should identify Straight angle (180°)", () => {
+  expect(getAngleType(180)).toEqual("Straight angle");
+});
 
 // Case 5: Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
+test("should identify Reflex angle (270°)", () => {
+  expect(getAngleType(270)).toEqual("Reflex angle");
+});
```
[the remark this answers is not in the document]
sheetalkharab: Thank you, I add 3 expect statement and test it according to your suggestion.

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -6,6 +6,27 @@ test("should return 11 for Ace of Spades", () => {
...
 // Case 3: Handle Face Cards (J, Q, K):
+test("should return 10 for Face Cards", () => {
+    const faceCards = getCardValue("10♦");
+    expect(faceCards).toEqual(10);
+    });
+
+
 // Case 4: Handle Ace (A):
+test("should return 10 for handleAce", () => {
```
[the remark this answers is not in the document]
sheetalkharab: Thank you I changed it .

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
sheetalkharab: I changed as we need to test for value J,Q,K according to your suggestion.

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
sheetalkharab: No, there is no need of suit in description as we are checking values for 2 to 10 and suit does not matter here. I tried to use for loop to check different value according to your suggestion. Thank you

Sprint-3/3-mandatory-practice/implement/count.js
sheetalkharab: this filter method create a new array with the element which pass condition . Here we need to check each character from stringOfCharacter which is equal to findCharacter. for example we have stringOfCharacters= "aaaaaaaabs" and findCharacter= a. so with filter method we will access each character from stringOfCharacter and fine which one is same to findCharacer. and length will give =8.

Sprint-3/4-stretch-investigate/creditCardValidator.test.js
sheetalkharab: yes its failing. I made some changes in function to make it pass .return statement should be place after because it return false before checking last condition. Thank you

cjyuan (commented):
Changes look good.

`Sprint-3/1-key-implement/3-get-card-value.js` could still use some improvement.

Feel free to mark this as completed after you have made the change to `getCardValue()`.

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -11,8 +11,8 @@ function getCardValue(card) {
   let rank = card.slice(0, -1);// extract rank everything except last character(suit)
     if (rank === "A") {
       return 11;
-    }else if(!isNaN(rank) && rank>="2" && rank<="9"){
-      return parseInt(rank, 10);
+    }else if(!isNaN(Number(rank)) && Number(rank) >= 2 && Number(rank) <= 10 ){
```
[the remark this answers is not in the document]
sheetalkharab: yes , it will pass this check as we are not strictly validate integer value which is wrong. so I implement regex check for this condition. Thank you.

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -6,20 +6,27 @@ test("should return 11 for Ace of Spades", () => {
...
-test("should return 10 for Face Cards", () => {
-    const faceCards = getCardValue("10♦");
-    expect(faceCards).toEqual(10);
+test("should return 10 for Face Cards(J, Q, K)", () => {
+    const faceCardsJack = getCardValue("J♠");
+    const faceCardsQueen = getCardValue("Q♠");
+    const faceCardsKing = getCardValue("K♠");
+    expect(faceCardsJack).toEqual(10);
+    expect(faceCardsQueen).toEqual(10);
+    expect(faceCardsKing).toEqual(10);
```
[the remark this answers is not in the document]
sheetalkharab: Thank you, I updated it as you suggested.

PR #323, opened 2025-02-22
London_ITP_Jan_2025 | Maryna Romanova | Module_Structuring_and_Testing_Data | Coursework/Sprint_1

MarynaRomashca:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -1,6 +1,9 @@
 let count = 0;
 
 count = count + 1;
+console.log(count)
+
+// 1 - From running slightly =0, then add extra line with +1. With new count =0+1=1, is assing
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
-// Describe what line 3 is doing, in particular focus on what = is doing
+// Describe what line 3 is doing, in particular focus on what = is doing - the line 3 provide new value for count
--- a/Sprint-1/1-key-exercises/2-initials.js
+++ b/Sprint-1/1-key-exercises/2-initials.js
@@ -5,7 +5,13 @@ let lastName = "Johnson";
 // Declare a variable called initials that stores the first character of each string.
 // This should produce the string "CKJ", but you must not write the characters C, K, or J in the code of your solution.
 
-let initials = ``;
+function initials(character) {
+    return character.toString().slice(0,1);
+  }
+  
+  console.log(`The first character of each string is ${initials(firstName)}${initials(middleName)}${initials(lastName)}`);
+
+
 
 // https://www.google.com/search?q=get+first+character+of+string+mdn
 
--- a/Sprint-1/1-key-exercises/2-initials.js
+++ b/Sprint-1/1-key-exercises/2-initials.js
@@ -11,7 +11,8 @@ function initials(character) {
   
   console.log(`The first character of each string is ${initials(firstName)}${initials(middleName)}${initials(lastName)}`);
 
-
+  let initials = ${firstName[0]}${middleName[0]}${lastName[0]};
+  console.log(The first character of each string is ${initials});
 
 // https://www.google.com/search?q=get+first+character+of+string+mdn
 
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -3,7 +3,10 @@ const maximum = 100;
 
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 
-// In this exercise, you will need to work out what num represents?
+console.log(`Random number between ${minimum} and ${maximum}: ${num}`);
+
+
+// In this exercise, you will need to work out what num represents? - generates a random integer between minimum and maximum by formula
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
-// Try logging the value of num and running the program several times to build an idea of what the program is doing
+// Try logging the value of num and running the program several times to build an idea of what the program is doing - Each time run the program, num will be different, and it will always be between 1 and 100, inclusive. 
--- a/Sprint-1/2-mandatory-errors/4.js
+++ b/Sprint-1/2-mandatory-errors/4.js
@@ -1,2 +1,2 @@
 const 12HourClockTime = "20:53";
-const 24hourClockTime = "08:53";
+const 24hourClockTime = "08:53";  //I don't understand the task
--- a/Sprint-1/2-mandatory-errors/4.js
+++ b/Sprint-1/2-mandatory-errors/4.js
@@ -1,2 +1,4 @@
 const 12HourClockTime = "20:53";
 const 24hourClockTime = "08:53";  //I don't understand the task
+const hourClockTimeOne = "20:53";
+const hourClockTimeTwo = "08:53";
--- a/Sprint-1/3-mandatory-interpret/1-percentage-change.js
+++ b/Sprint-1/3-mandatory-interpret/1-percentage-change.js
@@ -13,6 +13,16 @@ console.log(`The percentage change is ${percentageChange}`);
 
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
 // I see just one function call - console.log (line 10)
+// Function Calls:
+// Line 4: carPrice.replaceAll(",", "")
+
+// Line 4: Number(carPrice.replaceAll(",", ""))
+
+// Line 5: priceAfterOneYear.replaceAll(",", "")
+
+// Line 5: Number(priceAfterOneYear.replaceAll(",", ""))
+
+// Line 10: console.log(`The percentage change is ${percentageChange}`)
 
 // b) Run the code and identify the line where the error is coming from - why is this error occurring? How can you fix this problem?
 // line 5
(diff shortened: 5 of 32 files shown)


MarynaRomashca:
Kindly asking review my to review my Sprint_1.

Sprint-1/1-key-exercises/2-initials.js
```diff
@@ -5,7 +5,13 @@ let lastName = "Johnson";
...
 // This should produce the string "CKJ", but you must not write the characters C, K, or J in the code of your solution.
 
-let initials = ``;
+function initials(character) {
+    return character.toString().slice(0,1);
+  }
+  
+  console.log(`The first character of each string is ${initials(firstName)}${initials(middleName)}${initials(lastName)}`);
+
+
```
cjyuan, line 14: Your approach works. However, 
- There are simpler ways to retrieve the first character of a string than using `.slice()`.
- We don't have to implement a function just to return the first character of a string.

You can try asking AI for suggestions.

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -3,7 +3,10 @@ const maximum = 100;
...
 
-// In this exercise, you will need to work out what num represents?
+console.log(`Random number between ${minimum} and ${maximum}: ${num}`);
+
+
+// In this exercise, you will need to work out what num represents? - generates a random integer between minimum and maximum by formula
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
-// Try logging the value of num and running the program several times to build an idea of what the program is doing
+// Try logging the value of num and running the program several times to build an idea of what the program is doing - Each time run the program, num will be different, and it will always be between 1 and 100, inclusive.
```
cjyuan, line 12: Your description is correct.

To test your understanding, how would you write an expression (without using any variable in the expression) that can yield a random integer between -5 and 0 (including both -5 and 0)?

Sprint-1/2-mandatory-errors/4.js
```diff
@@ -1,2 +1,2 @@
 const 12HourClockTime = "20:53";
-const 24hourClockTime = "08:53";
\ No newline at end of file
+const 24hourClockTime = "08:53";  //I don't understand the task
```
cjyuan, line 2: If you run the code, the JS engine will indicate errors.
Can you explain the cause of the errors, and maybe even suggest a solution?

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -12,11 +12,13 @@ console.log(`The percentage change is ${percentageChange}`);
 // Read the code and then answer the questions below
 
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
+// I see just one function call - console.log (line 10)
```
cjyuan, line 15: There are more than one function call.

Suggestion: Lookup the syntax of a function call in JS.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -12,11 +12,13 @@ console.log(`The percentage change is ${percentageChange}`);
...
-
+// line 5
 // c) Identify all the lines that are variable reassignment statements
-
+//line 4 and 5
 // d) Identify all the lines that are variable declarations
-
+// all the lines with let and const 1, 2, 7, 8
 // e) Describe what the expression Number(carPrice.replaceAll(",","")) is doing - what is the purpose of this expression?
+// carPrice.replaceAll(",", "") removes all commas from the string, producing "10000" and "8543".
```
cjyuan, line 24: Your description of `carPrice.replaceAll(",", "")` is spot on.
Can you also describe the purpose of `Number(...)`? It is part of the expression too.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -25,3 +25,9 @@ console.log(`£${pounds}.${pence}`);
 
 // To begin, we can start with
 // 1. const penceString = "399p": initialises a string variable with the value "399p"
+// 2. const penceStringWithoutTrailingP = penceString.substring( 0, penceString.length - 1); : removing p from the number 399
+// 3. const paddedPenceNumberString : we are checking if the number has min 3 digits, if less than we will add "0" in the start. 
+// 4. paddedPenceNumberString.length - 2 : calculates the position 2 characters from the end.
+// 5..substring(paddedPenceNumberString.length - 2 : extracts the last 2 characters. 
+// 6. .padEnd(2, "0") : if our new number (paddedPenceNumberString.length) will have less then 2 digits we will add extra 0.
```
cjyuan, line 32: Can we expect this program to work as intended even if we deleted `.padEnd(2, "0")` from the code?

cjyuan (commented):
Code is generally good. I left you some comments and suggestions.

Why is there only one commit in this Sprint-1 branch? The backlog of this exercise has this suggestion:
> Make regular small commits in this branch with clear messages.

This Sprint-1 branch contains also modified files in the Sprint-2 and Sprint-3 folders. As a result, it makes reviewing modified files very inconvenient. Can you try fixing your Sprint-N (N = 1, 2, 3) branches (perhaps with the help of a volunteer) so that your Sprint-N branch contains only modified files in the "Sprint-N" folder?

Sprint-1/1-key-exercises/1-count.js
```diff
@@ -1,6 +1,9 @@
 let count = 0;
 
 count = count + 1;
+console.log(count)
+
+// 1 - From running slightly =0, then add extra line with +1. With new count =0+1=1, is assing
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
-// Describe what line 3 is doing, in particular focus on what = is doing
+// Describe what line 3 is doing, in particular focus on what = is doing - the line 3 provide new value for count
```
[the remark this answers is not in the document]
MarynaRomashca: Line 3 is performing an assignment operation. Specifically:

The expression on the right-hand side (count + 1) is evaluated first. Since count is currently 0, this results in 1.

The assignment operator (=) then takes this computed value (1) and stores it in the left-hand side variable (count).

As a result, the value of count is updated from 0 to 1.

Key Point:
The = operator does not mean equality (like in mathematics); instead, it assigns the value on the right to the variable on the left.

Sprint-1/1-key-exercises/2-initials.js
MarynaRomashca: You are right
let initials = `${firstName[0]}${middleName[0]}${lastName[0]}`;

console.log(`The first character of each string is ${initials}`);

Sprint-1/1-key-exercises/4-random.js
MarynaRomashca: I would do the same: 
const minimum = -5;
const maximum = 0;

console.log(`Random number between ${minimum} and ${maximum}: ${num}`);

Sprint-1/2-mandatory-errors/4.js
MarynaRomashca: SyntaxError: Invalid or unexpected token
 I cannot use a variable name that starts with a number in JavaScript. Variable names must start with a letter, underscore (_), or dollar sign ($).

const hourClockTimeOne = "20:53";
const hourClockTimeTwo = "08:53";

Sprint-1/3-mandatory-interpret/1-percentage-change.js
MarynaRomashca: Yes! The purpose of Number(...) in this code is to convert a string into a number.

The initial values of carPrice and priceAfterOneYear are strings ("10,000" and "8,543").

The .replaceAll(",", "") function removes the commas from these strings, resulting in "10000" and "8543".

The Number(...) function then converts these cleaned string values into actual numeric values (10000 and 8543).

Sprint-1/3-mandatory-interpret/3-to-pounds.js
MarynaRomashca: In our case if paddedPenceNumberString always has at least 2 digits, removing .padEnd(2, "0") is fine.

But, if paddedPenceNumberString has only 1 digit in the pence part, it will cause incorrect formatting.

cjyuan: By the time the statement involving `.padEnd(2, "0")` is called, can `paddedPenceNumberString` ever contain only 1 digit in the pence part?

Sprint-1/1-key-exercises/1-count.js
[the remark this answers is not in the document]
cjyuan: We can also say "***increment*** count by one".

cjyuan:
Your answers are good! Can you also update the files in `Sprint-1` folder accordingly and push them to Github? After that, feel free to mark this PR as "Complete".

Sprint-1/3-mandatory-interpret/3-to-pounds.js
MarynaRomashca: No, by the time .padEnd(2, "0") is called, paddedPenceNumberString will always have at least 2 digits in the pence part, so the .padEnd(2, "0") is actually unnecessary in specifically in this case.

MarynaRomashca:
All is Complete, Thanks so much!

PR #340, opened 2025-02-26
ITP-Jan 2025 | London | Maryna Romanova | Module-Structuring-and-Testing-Data | Week 6 | Sprint 3

MarynaRomashca:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/1-key-implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-key-implement/2-is-proper-fraction.js
@@ -9,6 +9,8 @@
 
 function isProperFraction(numerator, denominator) {
     if (numerator < denominator) return true;
+    else return false;
+     
 }
 
 // here's our helper again
@@ -40,14 +42,15 @@ assertEquals(improperFraction, false);
 // target output: true
 // Explanation: The fraction -4/7 is a proper fraction because the absolute value of the numerator (4) is less than the denominator (7). The function should return true.
 const negativeFraction = isProperFraction(-4, 7);
-// ====> complete with your assertion
+assertEquals(negativeFraction, true);
 
 // Equal Numerator and Denominator check:
 // Input: numerator = 3, denominator = 3
 // target output: false
 // Explanation: The fraction 3/3 is not a proper fraction because the numerator is equal to the denominator. The function should return false.
 const equalFraction = isProperFraction(3, 3);
-// ====> complete with your assertion
+assertEquals(equalFraction, false);
 
 // Stretch:
-// What other scenarios could you test for?
+// What other scenarios could you test for? - we can text non-integers
+
--- a/Sprint-3/1-key-implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-key-implement/2-is-proper-fraction.js
@@ -7,11 +7,18 @@
 // complete the rest of the tests and cases
 // write one test at a time, and make it pass, build your solution up methodically
 
-function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
-    else return false;
-     
+function isProperFractionFix(numerator, denominator) {
+  if (denominator === 0) return false; 
+  return Math.abs(numerator) < Math.abs(denominator);
 }
+
+// function isProperFraction(numerator, denominator) {
+//     if (numerator < denominator) return true;
+//     else return false;
+
+    
+     
+// }
 
 // here's our helper again
 function assertEquals(actualOutput, targetOutput) {
@@ -53,4 +60,9 @@ assertEquals(equalFraction, false);
 
 // Stretch:
 // What other scenarios could you test for? - we can text non-integers
+const wtoNegativeFraction = isProperFractionFix(-3, -2);
+assertEquals(equalFraction, false);
+
+
+
 
--- a/Sprint-3/1-key-implement/3-get-card-value.js
+++ b/Sprint-3/1-key-implement/3-get-card-value.js
@@ -8,8 +8,13 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
+  const rank = card.slice(0, -1);
     if (rank === "A") return 11;
-}
+    if (rank === "J" || rank === "Q" || rank === "K" || rank === "10") return 10;
+    if (rank >= "2" && rank <= "9") return parseInt(rank);
+    throw new Error("Invalid card rank");
+  }
+
 
 // You need to write assertions for your function to check it works in different cases
 // we're going to use this helper function to make our assertions easier to read
@@ -33,19 +38,32 @@ assertEquals(aceofSpades, 11);
 // When the function is called with such a card,
 // Then it should return the numeric value corresponding to the rank (e.g., "5" should return 5).
 const fiveofHearts = getCardValue("5♥");
-// ====> write your test here, and then add a line to pass the test in the function above
+assertEquals(fiveofHearts, parseInt(rank));
 
 // Handle Face Cards (J, Q, K):
 // Given a card with a rank of "10," "J," "Q," or "K",
 // When the function is called with such a card,
 // Then it should return the value 10, as these cards are worth 10 points each in blackjack.
+const fiveofCard = getCardValue("Q♥");
+assertEquals(fiveofCard, 10);
 
 // Handle Ace (A):
 // Given a card with a rank of "A",
 // When the function is called with an Ace,
 // Then it should, by default, assume the Ace is worth 11 points, which is a common rule in blackjack.
+const fiveofAce = getCardValue("A♥");
+assertEquals(fiveofAce, 11);
 
 // Handle Invalid Cards:
 // Given a card with an invalid rank (neither a number nor a recognized face card),
 // When the function is called with such a card,
 // Then it should throw an error indicating "Invalid card rank."
+const fiveofInvalid = getCardValue("Z♥");
+assertEquals(fiveofInvalid, Error);
+
+
+
+
+
+
+
--- a/Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
+++ b/Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
@@ -10,15 +10,27 @@ test("should identify right angle (90°)", () => {
 // Case 2: Identify Acute Angles:
 // When the angle is less than 90 degrees,
 // Then the function should return "Acute angle"
+test("should identify right angle (90°)", () => {
+  expect(getAngleType(90)).toEqual("Right angle");
+});
 
 // Case 3: Identify Obtuse Angles:
 // When the angle is greater than 90 degrees and less than 180 degrees,
 // Then the function should return "Obtuse angle"
+test("should identify right angle (90°)", () => {
+  expect(getAngleType(90)).toEqual("Right angle");
+});
 
 // Case 4: Identify Straight Angles:
 // When the angle is exactly 180 degrees,
 // Then the function should return "Straight angle"
+test("should identify right angle (90°)", () => {
+  expect(getAngleType(90)).toEqual("Right angle");
+});
 
 // Case 5: Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
+test("should identify right angle (90°)", () => {
+  expect(getAngleType(90)).toEqual("Right angle");
+});
--- a/Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
+++ b/Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
@@ -10,27 +10,27 @@ test("should identify right angle (90°)", () => {
 // Case 2: Identify Acute Angles:
 // When the angle is less than 90 degrees,
 // Then the function should return "Acute angle"
-test("should identify right angle (90°)", () => {
-  expect(getAngleType(90)).toEqual("Right angle");
+test("should identify acute angle (<90°)", () => {
+  expect(getAngleType(45)).toEqual("Acute angle");
 });
 
 // Case 3: Identify Obtuse Angles:
 // When the angle is greater than 90 degrees and less than 180 degrees,
 // Then the function should return "Obtuse angle"
-test("should identify right angle (90°)", () => {
-  expect(getAngleType(90)).toEqual("Right angle");
+test("should identify obtuse angle (120°)", () => {
+  expect(getAngleType(120)).toEqual("Obtuse angle");
 });
 
 // Case 4: Identify Straight Angles:
 // When the angle is exactly 180 degrees,
 // Then the function should return "Straight angle"
-test("should identify right angle (90°)", () => {
-  expect(getAngleType(90)).toEqual("Right angle");
+test("should identify straight angle (180°)", () => {
+  expect(getAngleType(180)).toEqual("Straight angle");
 });
 
 // Case 5: Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
-test("should identify right angle (90°)", () => {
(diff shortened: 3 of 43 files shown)


MarynaRomashca (commented):
Please check my Sprint 3

cjyuan:
Can you fix this branch so that it only contains modified files in the "Sprint-3" folder?

MarynaRomashca:
@cjyuan Done, the "Sprint-3" folder is fixed. Please review it. Many thanks.

Sprint-3/1-key-implement/2-is-proper-fraction.js
```diff
@@ -9,6 +9,8 @@
 
 function isProperFraction(numerator, denominator) {
     if (numerator < denominator) return true;
+    else return false;
+
```
cjyuan, line 13: Mathematically speaking, -3/2 and -1/0 are not proper fractions (but -3 < 2 and -1 < 0) and -2/-3 is a proper fraction (but -2 > -3). 

We need to consider the cases that `numerator` and `denominator` can be negative, positive, or zero.

However, if we were to consider all possible combinations of negative/positive/zero numerator and denominator, we would have to use a lot of if-statements. One way to reduce the complexity of the code is to "remove the negative sign" in `numerator` and `denominator`. If both numerator and denominator are non-negative numbers, then the if-statements at lines 11-12 would work perfectly.

 ***Input normalisation***, the process of pre-processing the input values before they are used in computation, can often simplify complexity of code, improve performance, and prevent errors.

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -8,8 +8,13 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
+  const rank = card.slice(0, -1);
     if (rank === "A") return 11;
-}
+    if (rank === "J" || rank === "Q" || rank === "K" || rank === "10") return 10;
+    if (rank >= "2" && rank <= "9") return parseInt(rank);
```
cjyuan, line 14: Does your function return the value you expected from each of the following function calls?

```
getCardValue("29♠");
getCardValue("0x02♠");
getCardValue("2.1♠")
```

Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
```diff
@@ -10,15 +10,27 @@ test("should identify right angle (90°)", () => {
...
+test("should identify right angle (90°)", () => {
+  expect(getAngleType(90)).toEqual("Right angle");
+});
 
 // Case 3: Identify Obtuse Angles:
 // When the angle is greater than 90 degrees and less than 180 degrees,
 // Then the function should return "Obtuse angle"
+test("should identify right angle (90°)", () => {
+  expect(getAngleType(90)).toEqual("Right angle");
+});
```
cjyuan, line 22: Why are all the tests testing the same thing (testing for "Right angle" 4 times) ?

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
```diff
@@ -3,4 +3,11 @@ function isProperFraction(numerator, denominator) {
     // add your completed function from key-implement here
 }
 
+
+function assertEquals(actualOutput, targetOutput) {
+    console.assert(
+      actualOutput === targetOutput,
+      `Expected ${actualOutput} to equal ${targetOutput}`
+    );
+  }
```
cjyuan, line 12: The function is to be tested in `2-is-proper-fraction.test.js`. So we don't need the `assertEquals()` at lines 6-12.

After you reimplemented `isProperFraction()` in `1-key-implement/2-is-proper-fractions.js`, you can copy the code of that function here.

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
```diff
@@ -5,7 +5,13 @@ test("should return true for a proper fraction", () => {
 });
 
 // Case 2: Identify Improper Fractions:
+const improperFraction = isProperFraction(5, 2);
+assertEquals(improperFraction, false);
```
cjyuan, line 9: This exercise expects us to write Jest test code to test the function.

So to test improper fraction, we can express our test code as
```
test("should return false for a improper fraction", () => {
 // We can test multiple sets of values belonging to the same case
  expect(isProperFraction(5, 2)).toEqual(false);  
  expect(isProperFraction(5, -2)).toEqual(false);
  expect(isProperFraction(-5, 2)).toEqual(false);    
  expect(isProperFraction(-5, -2)).toEqual(false);
});
```

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -6,6 +6,17 @@ test("should return 11 for Ace of Spades", () => {
     });
 
 // Case 2: Handle Number Cards (2-10):
+const fiveofHearts = getCardValue("5♥");
+assertEquals(fiveofHearts, parseInt(rank));
+
```
cjyuan, line 11: We should express Jest test code similar to those expressed at lines 3-6 in this file.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -8,6 +8,17 @@ const getOrdinalNumber = require("./get-ordinal-number");
 // When the number is 1,
 // Then the function should return "1st"
 
+test("should return '3rd' for 3", () => {
+    expect(getOrdinalNumber(3)).toEqual("3rd");
+});
```
cjyuan, line 13: This is good. We can also generalise the test case to include numbers like 23, 53, 103. For example,

```
test("should append suffix 'rd' to numbers ending in 3 (but not in 13)", () => {
    expect(getOrdinalNumber(3)).toEqual("3rd");
    expect(getOrdinalNumber(23)).toEqual("23rd");
    expect(getOrdinalNumber(53)).toEqual("53rd");
    expect(getOrdinalNumber(103)).toEqual("103rd");            
});
```

Sprint-3/3-mandatory-practice/implement/repeat.test.js
```diff
@@ -16,6 +16,13 @@ test("should repeat the string count times", () => {
     expect(repeatedStr).toEqual("hellohellohello");
     });
 
+test("should repeat the string count times", () => {
+    const str = "hi";
+    const count = 3;
+    const repeatedStr = repeat(str, count);
+    expect(repeatedStr).toEqual("hihihi");
+    });
+
```
cjyuan, line 25: What about those cases described in the comments at lines 26-39?

For the negative count case, you can prepare your test as:
```
test("should throw an error when count is negative", () => {
    expect(() => {
      repeat('hi', -1)         
    }).toThrow();
});
```
Please note that  when we expect a function to throw an error, we need to wrap the function call inside `() => { ... }`.

Sprint-3/4-stretch-investigate/find.js
```diff
@@ -19,7 +19,7 @@ console.log(find("code your future", "z"));
...
 // Pay particular attention to the following:
 
-// a) How the index variable updates during the call to find
-// b) What is the if statement used to check
-// c) Why is index++ being used?
-// d) What is the condition index < str.length used for?
+// a) How the index variable updates during the call to find - index start with O and going on with +1 until not reach str.length
+// b) What is the if statement used to check - in this situation "if" check if index finally the same with char, if it's the same the function returning index
+// c) Why is index++ being used? - it's increases index after each one while so we can't get stuck. 
+// d) What is the condition index < str.length used for? - function doesn't work if the index higher then str.lenght. So function can understands in which limits need to work.
```
cjyuan, line 25: For question (d), suppose `str` is `"ABCD"` and `char` is `'a'`.
1. What is the value of `index` in the last iteration of the while-loop? 
2. What is the value of `index` when the execution reaches line 10?

Why not try using an AI tool such as ChatGPT to check grammar in your writings?

Sprint-3/4-stretch-investigate/password-validator.js
```diff
@@ -1,5 +1,20 @@
-function passwordValidator(password) {
-    return password.length < 5 ? false : true
+function passwordValidator(password, passwords) {
+    password.length < 5;
```
cjyuan, line 2: This statement does not do anything except comparing two values.

As a result, your function would not be able to return `false` when the given password is "Aa1$".

Sprint-3/4-stretch-investigate/password-validator.test.js
```diff
@@ -15,6 +15,7 @@ To be valid, a password must:
 You must breakdown this problem in order to solve it. Find one test case first and get that working
 */
 const isValidPassword = require("./password-validator");
+
```
cjyuan, line 18: The Jest test code is missing in this file.

Sprint-3/4-stretch-investigate/cardnumber.js
```diff
@@ -0,0 +1,23 @@
...
+    return cardNumber;
+}
+
+function cardValidator (card_number) {
+    if (card_number.length === 16) {
+        return true;
+    } else {
+        return false;
+    }
+}
```
cjyuan, line 17: In the card validation exercise, there are other requirements for a card number to be considered valid.
As this is the "stretch" part of the exercise, you can complete the implementation of this function when you have time.

cjyuan (commented):
Some of the functions could use some improvement or bug fix.

Most of my comments are related to Jest tests. Do you need assistance on using Jest?

Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
MarynaRomashca: sorry my mistake

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
MarynaRomashca: got it

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
MarynaRomashca: so deleted the lines 6-12. and fix the test in the file ' 2-is-proper-fraction.test.js '

Sprint-3/4-stretch-investigate/password-validator.test.js
MarynaRomashca: Added

Sprint-3/4-stretch-investigate/password-validator.js
MarynaRomashca: Fixed

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
MarynaRomashca: Thanks, included.

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
MarynaRomashca: Done

Sprint-3/1-key-implement/2-is-proper-fraction.js
MarynaRomashca: Fixed

Sprint-3/1-key-implement/3-get-card-value.js
MarynaRomashca: No, the function does not return this numeric. In all three cases, the function throws an "Invalid card rank" error.

Sprint-3/3-mandatory-practice/implement/repeat.test.js
MarynaRomashca: Thanks, very useful.

Sprint-3/4-stretch-investigate/find.js
MarynaRomashca: yes, I know that chat GPT is good for grammar correction, but I use it every day for my work, and I don't have enough practice to use my own brain, and I can't pass my Duo Lingo test (Just 80%), it's mean that I would be able to move to the another CYF's level. That why I'm trying to use my brain in the lessons to train my self. 
Thanks for understanding and sorry for mistakes.

cjyuan: I wouldn't encourage you to blindly copy the answer prepared by ChatGPT.
When I used ChatGPT to check grammar in my writing or improve my writing, it always explained why the corrected version was better. I find the explanation helpful in improving my writing skills.

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
cjyuan, line 1: Why not update this function as well?
Have you run the Jest test after you made the changes?

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -1,22 +1,29 @@
...
+expect(handleAce).toEqual(11);
+});
 
-// Case 5: Handle Invalid Cards:
-const fiveofInvalid = getCardValue("Z♥");
-assertEquals(fiveofInvalid, Error);
+// // Case 5: Handle Invalid Cards:
+test("Handle Invalid Cards", () => {
+    expect(() => getCardValue("Z♥")).toThrow("Invalid");
+  });
```
cjyuan, line 29: Can you include these test values in this test, and then run this script using Jest? I don't think your script can recognise them as invalid cases (because of the way strings are compared).
```
getCardValue("29♠");
getCardValue("2.1♠");
```

If you have installed Jest, you can run this particular test script as `npx jest 3-get-card-value.test.js` in the folder where this file is located.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -8,8 +8,14 @@ const getOrdinalNumber = require("./get-ordinal-number");
 // When the number is 1,
 // Then the function should return "1st"
 
-test("should return '3rd' for 3", () => {
+
+
+test("should return '3rd' for 3 and more", () => {
```
cjyuan, line 13: The description of this test is not yet updated accordingly.

cjyuan (commented):
I think `getCardValue()` can still use some improvement.

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
MarynaRomashca: All tests pass !

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
MarynaRomashca: You are right.

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
MarynaRomashca: Done.

MarynaRomashca:
@cjyuan All updated. Please approve it.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -10,7 +10,7 @@ const getOrdinalNumber = require("./get-ordinal-number");
 
 
 
-test("should return '3rd' for 3 and more", () => {
+test("should return all mentioned test bellow", () => {
```
cjyuan, line 13: This description does not tell what is being tested in this particular test.

When there is an error, Jest will display this message but not all of the code at lines 14-16. The person running the test would have to take an extra step to view the code in the test script to find out what the tests are about.

cjyuan (commented):
`getCardValue()` is solid now.

One of the test descriptions could be improved. You can mark this as "Complete" after you update the description.

PR #344, opened 2025-03-01
ITP LONDON - JAN 2025 | Samunta_Sunuwar | Module-Structuring-and-Testing-Data | SPRINT 3|Week 6

Samunta:

### DIFF
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -11,8 +11,8 @@ function getAngleType(angle) {
   if (angle === 90) return "Right angle";
   // read to the end, complete line 36, then pass your test here
   else if (angle > 0 && angle < 90) return "Acute angle";
-  else if (angle > 90 && angle <= 120) return "Obtuse Angle";
-  else if (angle > 120 && angle <= 180) return "Straight Angle";
+  else if (angle > 90 && angle < 180) return "Obtuse Angle";
+  else if ((angle = 180)) return "Straight Angle";
   else if (angle > 180 && angle <= 360) return "Reflex Angle";
   else
     return "This angle is not correct. Please check your input and try again";
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -8,9 +8,18 @@
 // Then, write the next test! :) Go through this process until all the cases are implemented
 
 function getAngleType(angle) {
-   if (angle === 90) return "Right angle";
-   // read to the end, complete line 36, then pass your test here
+  if (angle === 90) return "Right angle";
+  // read to the end, complete line 36, then pass your test here
+  else if (angle > 0 && angle < 90) return "Acute angle";
+  else if (angle > 90 && angle <= 120) return "Obtuse Angle";
+  else if (angle > 120 && angle <= 180) return "Straight Angle";
+  else if (angle > 180 && angle <= 360) return "Reflex Angle";
+  else
+    return "This angle is not correct. Please check your input and try again";
 }
+//Printing following for visualization
+console.log(`Should print Acute Angle : ${getAngleType(40)}`);
+console.log(`Should print Invalid Message : ${getAngleType(400)}`);
 
 // we're going to use this helper function to make our assertions easier to read
 // if the actual output matches the target output, the test will pass
@@ -44,13 +53,18 @@ assertEquals(acute, "Acute angle");
 // Then the function should return "Obtuse angle"
 const obtuse = getAngleType(120);
 // ====> write your test here, and then add a line to pass the test in the function above
+assertEquals(obtuse, "Obtuse Angle");
 
 // Case 4: Identify Straight Angles:
 // When the angle is exactly 180 degrees,
 // Then the function should return "Straight angle"
+const straight = getAngleType(180);
 // ====> write your test here, and then add a line to pass the test in the function above
+assertEquals(straight, "Straight Angle");
 
 // Case 5: Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
+const reflex = getAngleType(185);
 // ====> write your test here, and then add a line to pass the test in the function above
+assertEquals(reflex, "Reflex Angle");
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -12,8 +12,8 @@ function getAngleType(angle) {
   // read to the end, complete line 36, then pass your test here
   else if (angle > 0 && angle < 90) return "Acute angle";
   else if (angle > 90 && angle < 180) return "Obtuse Angle";
-  else if ((angle = 180)) return "Straight Angle";
-  else if (angle > 180 && angle <= 360) return "Reflex Angle";
+  else if (angle === 180) return "Straight Angle";
+  else if (angle > 180 && angle < 360) return "Reflex Angle";
   else
     return "This angle is not correct. Please check your input and try again";
 }
--- a/Sprint-3/1-key-implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-key-implement/2-is-proper-fraction.js
@@ -8,18 +8,19 @@
 // write one test at a time, and make it pass, build your solution up methodically
 
 function isProperFraction(numerator, denominator) {
-  // return (
+  // if (
   //   numerator < denominator &&
   //   Number.isInteger(numerator) &&
   //   Number.isInteger(denominator)
-  // );
-  if (
+  // )
+  //   return true;
+  // else return false;
+  //refactored code (I am using Number.Integer method to make sure that decimal integers return as false)
+  return (
     numerator < denominator &&
     Number.isInteger(numerator) &&
     Number.isInteger(denominator)
-  )
-    return true;
-  else return false;
+  );
 }
 
 // here's our helper again
@@ -65,8 +66,10 @@ assertEquals(equalFraction, false);
 // Stretch:
 // What other scenarios could you test for?
 
+//Scenario where numerator is in decimal number
 const decimalFraction = isProperFraction(1.8, 4);
 assertEquals(decimalFraction, false);
 
+//Scenario where denominator is undefined
 const nullDenominator = isProperFraction(5);
 assertEquals(nullDenominator, false);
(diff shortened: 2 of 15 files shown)


Sprint-3/1-key-implement/1-get-angle-type.js
```diff
@@ -8,9 +8,18 @@
...
 
 function getAngleType(angle) {
-   if (angle === 90) return "Right angle";
-   // read to the end, complete line 36, then pass your test here
+  if (angle === 90) return "Right angle";
+  // read to the end, complete line 36, then pass your test here
+  else if (angle > 0 && angle < 90) return "Acute angle";
+  else if (angle > 90 && angle < 180) return "Obtuse Angle";
+  else if ((angle = 180)) return "Straight Angle";
+  else if (angle > 180 && angle <= 360) return "Reflex Angle";
```
cjyuan, line 16: This condition does not quite match the description "angle is greater than 180 degrees and less than 360 degrees"

Sprint-3/1-key-implement/1-get-angle-type.js
```diff
@@ -8,9 +8,18 @@
 // Then, write the next test! :) Go through this process until all the cases are implemented
 
 function getAngleType(angle) {
-   if (angle === 90) return "Right angle";
-   // read to the end, complete line 36, then pass your test here
+  if (angle === 90) return "Right angle";
+  // read to the end, complete line 36, then pass your test here
+  else if (angle > 0 && angle < 90) return "Acute angle";
+  else if (angle > 90 && angle < 180) return "Obtuse Angle";
+  else if ((angle = 180)) return "Straight Angle";
```
cjyuan, line 15: `=` is an assignment operator (and not an equality operator).

Sprint-3/1-key-implement/2-is-proper-fraction.js
```diff
@@ -8,7 +8,19 @@
...
+  //   Number.isInteger(denominator)
+  // )
+  //   return true;
+  // else return false;
+  //refactored code (I am using Number.Integer method to make sure that decimal integers return as false)
+  return (
+    numerator < denominator &&
+    Number.isInteger(numerator) &&
+    Number.isInteger(denominator)
+  );
```
cjyuan, line 23: Mathematically speaking, -3/2 and -1/0 are not proper fractions and -2/-3 is a proper fraction.

If we consider all possible combination of negative/positive/zero numerator/denominator, we would need to use a lot of if-statements.
However, if we "removed the negative sign" in `numerator` and `denominator` first, we would not need to consider as many cases. ***Input normalisation", the process of pre-processing the input values before they are used in computation, can often simplify complexity of code, improve performance, and prevent errors.

JS has a built-in function you can use to compute absolute value of a number.

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -8,7 +8,40 @@
...
+  if (rank === "A") {
+    return 11;
+  } else if (rank === "K") {
+    return 10;
+  } else if (rank === "Q") {
+    return 10;
+  } else if (rank === "J") {
+    return 10;
+  } else if (rank === "1") {
+    return 1;
```
cjyuan, line 25: I think the number cards typically range from 2-10.

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -8,7 +8,40 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
-    if (rank === "A") return 11;
+  if (card.startsWith("10")) {
+    // Check for "10" before extracting first character
+    return 10;
+  }
+  const rank = card.charAt(0);
```
cjyuan, line 15: Invalid card value could also starts with a valid prefix. For examples,

"QQ♠", "10000♠", "299♠", "3.14♠"

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -8,7 +8,40 @@
...
+  if (rank === "A") {
+    return 11;
+  } else if (rank === "K") {
+    return 10;
+  } else if (rank === "Q") {
+    return 10;
+  } else if (rank === "J") {
```
cjyuan, line 22: We can also use the logical OR operator (`||`) to combine the condition in these if-statements into one condition.

Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
```diff
@@ -10,15 +10,39 @@ test("should identify right angle (90°)", () => {
...
+  expect(getAngleType(40)).toEqual("Acute angle");
+});
 
 // Case 3: Identify Obtuse Angles:
 // When the angle is greater than 90 degrees and less than 180 degrees,
 // Then the function should return "Obtuse angle"
-
+test("should identify Obtuse Angles(91° to 179°)", () => {
+  expect(getAngleType(115)).toEqual("Obtuse Angle");
+});
```
cjyuan, line 22: You can include multiple test values in each test case. For example,
```
test("should identify Obtuse Angles(91° to 179°)", () => {
  expect(getAngleType(91)).toEqual("Obtuse Angle");
  expect(getAngleType(115)).toEqual("Obtuse Angle");
  expect(getAngleType(179)).toEqual("Obtuse Angle");
});
```

If you try that, you may detect some bug in your function.

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
```diff
@@ -6,6 +6,24 @@ test("should return true for a proper fraction", () => {
...
 // Case 2: Identify Improper Fractions:
 
+test("should return false for fraction where Numerator is grater than Denominator", () => {
+  expect(isProperFraction(10, 2)).toEqual(false);
+});
+
 // Case 3: Identify Negative Fractions:
+test("should return true for fraction where Numerator and Denominator are negative integer as long as Denominator is greater than Numerator", () => {
+  expect(isProperFraction(-6, -3)).toEqual(true);
+});
```
cjyuan, line 16: In math, both -6/-3 and 6/3 are equal to 2. So -6/-3 should not be considered a proper fraction. (You can double check the definition of "Proper fraction")

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -9,5 +9,19 @@ const getOrdinalNumber = require("./get-ordinal-number");
...
+});
+
+// Case 3
+test("should return '3rd' for 3", () => {
+  expect(getOrdinalNumber(3)).toEqual("3rd");
+});
+//Case 4
+test("should return '13th' for 13", () => {
+  expect(getOrdinalNumber(13)).toEqual("13th");
+});
```
cjyuan, line 27: We could test more values to make the test more comprehensive. For examples, 113, 101, 99991, 211.

Sprint-3/3-mandatory-practice/implement/repeat.js
```diff
@@ -1,5 +1,9 @@
-function repeat() {
-    return "hellohellohello";
+function repeat(str, count) {
+  if (count < 0) {
+    return "Count cannot be negative";
+  } else {
+    return str.repeat(count);
+  }
```
cjyuan, line 6: How would the caller distinguish the result of the following two function calls?
1. `repeat("Count cannot be negative", 1)`
2. `repeat("", -1)`

cjyuan (commented):
There are a few bugs. Otherwise the code is good. Well done.

I left you some comments and suggestions.

Sprint-3/1-key-implement/1-get-angle-type.js
Samunta: Thank you I have made the corrections.

Sprint-3/1-key-implement/3-get-card-value.js
Samunta: Thank you, I have used Math.abs() to my parameters in the function and corrected the code

Samunta: Thank you, I have removed this

Sprint-3/1-key-implement/2-is-proper-fraction.js
Samunta: Thank you I have made the necessary chages

Sprint-3/1-key-implement/3-get-card-value.js
Samunta: Thank you I have made the necessary corrections.

Sprint-3/1-key-implement/3-get-card-value.js
Samunta: Thank you I have made the necessary corrections.

Sprint-3/1-key-implement/1-get-angle-type.js
Samunta: Thank you I have made the corrections

Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
Samunta: Thank you I have corrected my function and the test is working as expected

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
Samunta: Thank you I have corrected this

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
Samunta: Thank you, I have added these extra test cases

Sprint-3/3-mandatory-practice/implement/repeat.js
Samunta: @cjyuan Hi CJ, thank you for your feedback. I am so sorry I don't think I follow. Please could you elaborate a bit.

cjyuan: When the caller checks for error, the caller would not be able to differentiate a normal function call (1st call) from a function call that results in an error (the 2nd call). The chance is slim but is possible. 

You can check out how `String.prototype.repeat()` deals with negative parameter and practice using `throw` to signal error.

Samunta: Thankyou so much, I have updated my code and test case with throw to signal

PR #356, opened 2025-03-08
LONDON_JAN25 | GIOSEF FERRARO | STRUCTURING_AND_TESTING_DATA | SPRINT 1 | WEEK 6

Giosef92:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -3,7 +3,21 @@ const maximum = 100;
 
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 
+console.log(num);
+
 // In this exercise, you will need to work out what num represents?
+
+//is a random integer between 1 and 100,inclusive;
+
 // Try breaking down the expression and using documentation to explain what it means
+
+//math.random() generates a random floating point number between 0 and 1(exclusive);
+//(maximum - minimum +1) calculates the range of the random number;
+//Math.random()*(maximum - minimum +1) ensures that the generated number is within the specified range;
+//Math.floor() rounds down the number to the nearest integer;
+//+minimum shifts the range of values from 0-99 to 1-100;
+
 // It will help to think about the order in which expressions are evaluated
+
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+//The final value of num will be a random integer between 1 and 100 ,inclusive.  num could be any of the following values ( it 's just an examples): 45, 100, 23 , 56 etc..
--- a/Sprint-1/3-mandatory-interpret/1-percentage-change.js
+++ b/Sprint-1/3-mandatory-interpret/1-percentage-change.js
@@ -2,21 +2,33 @@ let carPrice = "10,000";
 let priceAfterOneYear = "8,543";
 
 carPrice = Number(carPrice.replaceAll(",", ""));
-priceAfterOneYear = Number(priceAfterOneYear.replaceAll("," ""));
+priceAfterOneYear = Number(priceAfterOneYear.replaceAll(",", ""));
 
 const priceDifference = carPrice - priceAfterOneYear;
 const percentageChange = (priceDifference / carPrice) * 100;
 
-console.log(`The percentage change is ${percentageChange}`);
+console.log(`The percentage change is ${percentageChange}%`);
 
 // Read the code and then answer the questions below
 
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
+//There are 4 function calls in the code. replaceAll()Line: 4 , 5 . Number() ,line:4 , 5.
 
 // b) Run the code and identify the line where the error is coming from - why is this error occurring? How can you fix this problem?
+//the error is in line 5. This error is happening because replaceAll() is not supported in some older versions of JavaScript.priceAfterOneYear = Number(priceAfterOneYear.replace(/,/g, ""));
 
 // c) Identify all the lines that are variable reassignment statements
+//Line 4: carPrice = Number(carPrice.replaceAll(",", ""));
+//Line 5: priceAfterOneYear = Number(priceAfterOneYear.replaceAll(",", ""));
 
 // d) Identify all the lines that are variable declarations
+/*The following lines are variable declarations:
+
+Line 1: let carPrice = "10,000";
+Line 2: let priceAfterOneYear = "8,543";
+Line 7: const priceDifference = carPrice - priceAfterOneYear;
+Line 8: const percentageChange = (priceDifference / carPrice) * 100;  */
 
 // e) Describe what the expression Number(carPrice.replaceAll(",","")) is doing - what is the purpose of this expression?
+//replaceAll(",", "") removes all commas from the string carPrice.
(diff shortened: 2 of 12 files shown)


Sprint-1/1-key-exercises/4-random.js
```diff
@@ -3,7 +3,21 @@ const maximum = 100;
...
+
+//is a random integer between 1 and 100,inclusive;
+
 // Try breaking down the expression and using documentation to explain what it means
+
+//math.random() generates a random floating point number between 0 and 1(exclusive);
+//(maximum - minimum +1) calculates the range of the random number;
+//Math.random()*(maximum - minimum +1) ensures that the generated number is within the specified range;
+//Math.floor() rounds down the number to the nearest integer;
+//+minimum shifts the range of values from 0-99 to 1-100;
```
suziebrown, line 18: Good explanation 👍

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -2,21 +2,33 @@ let carPrice = "10,000";
...
-console.log(`The percentage change is ${percentageChange}`);
+console.log(`The percentage change is ${percentageChange}%`);
 
 // Read the code and then answer the questions below
 
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
+//There are 4 function calls in the code. replaceAll()Line: 4 , 5 . Number() ,line:4 , 5.
 
 // b) Run the code and identify the line where the error is coming from - why is this error occurring? How can you fix this problem?
+//the error is in line 5. This error is happening because replaceAll() is not supported in some older versions of JavaScript.priceAfterOneYear = Number(priceAfterOneYear.replace(/,/g, ""));
```
suziebrown, line 18: Line 5 was causing an error even in the newest version of JavaScript. It looks like you worked out what the actual problem was, do you want to update your explanation?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,20 @@ console.log(result);
 // For the piece of code above, read the code and then answer the following questions
 
 // a) How many variable declarations are there in this program?
+// 5 variable declaration: movieLength, remainingSeconds, totalMinutes, remainingMinutes, totalHours.
```
suziebrown, line 15: Are you sure that's all of them?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,20 @@ console.log(result);
...
 
 // b) How many function calls are there?
+// 2 functions
```
suziebrown, line 18: Can you be more specific? Which functions are called? You could include line numbers

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,20 @@ console.log(result);
...
 
 // c) Using documentation, explain what the expression movieLength % 60 represents
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators
+//The modulus operator (%) calculates the remainder when dividing movieLength by 60.
 
 // d) Interpret line 4, what does the expression assigned to totalMinutes mean?
+// converts the movie length into minutes
```
suziebrown, line 25: Can you give a bit more detail?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,20 @@ console.log(result);
...
 
 // e) What do you think the variable result represents? Can you think of a better name for this variable?
+//A better name could be movieTime .
 
 // f) Try experimenting with different values of movieLength. Will this code work for all values of movieLength? Explain your answer
+//no this code will not work properly for all values of movieLength
```
suziebrown, line 31: Do you have an example of a length that doesn't work?

suziebrown (commented):
Well done 👍 
Most of your explanations are very clear, but there are a few where you could give more detail.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
Giosef92: The code converts car prices from strings with commas into integer numbers by removing the commas using .replace(/,/g, "") and then converting them to numbers with Number(). This allows the calculation of the price difference and the percentage change. I replaced replaceAll() with replace(/,/g, "") because the latter has broader compatibility. 
Thank you for taking the time to review my code—I really appreciate your feedback!

Sprint-1/3-mandatory-interpret/2-time-format.js
Giosef92: I missed the result variable in my initial response. After reviewing the code again, here are all the variable declarations:"

movieLength (Line 1): Represents the movie's length in seconds.
remainingSeconds (Line 3): The remainder of seconds when dividing movieLength by 60.
totalMinutes (Line 4): The total minutes calculated from the movie length in seconds.
remainingMinutes (Line 5): The remainder of minutes after dividing totalMinutes by 60.
totalHours (Line 6): The total hours calculated from the total minutes.
result (Line 8): A formatted string representing the movie time in hours: minutes:seconds format.

Sprint-1/3-mandatory-interpret/2-time-format.js
Giosef92: There are two function calls in this code:

Template literal (Line 8): A string interpolation function is used here to format the string \${totalHours}:${remainingMinutes}:${remainingSeconds}`.
console.log() (Line 9): This function prints the result variable to the console.

Sprint-1/3-mandatory-interpret/2-time-format.js
Giosef92: The expression assigned to totalMinutes performs the following steps:

First, it calculates how many seconds remain after dividing the total movieLength by 60, using the modulo operator (movieLength % 60).
Then, it subtracts these remaining seconds from the total movieLength to obtain the number of full seconds that form complete minutes.
Finally, it divides this result by 60 to convert the total number of seconds into minutes. Essentially, this gives the total number of full minutes in the movie.
This step is crucial for breaking down the movie length into hours, minutes, and seconds for better readability.

Sprint-1/3-mandatory-interpret/2-time-format.js
Giosef92: Here's one:

movieLength = 45 (45 seconds).
Expected output: "0:0:45" (0 hours, 0 minutes, 45 seconds).
Actual output: The code will correctly output 0:0:45, but it won't display any "hours" or "minutes" for very short movie lengths like this.
Issue: For very small values of movieLength (especially those under 60 seconds), the code will still work but the interpretation becomes somewhat less useful, as it lacks a general format to handle such short durations.
movieLength = 59 (59 seconds):
Expected output: "0:0:59"
Actual output: Will still output correctly as 0:0:59, but it might feel a bit weird because the logic is designed for longer durations (greater than 60 seconds).

PR #359, opened 2025-03-08
ITPJAN|SARAAMIRI|Module-structuring-and-testing-data|sprint3|week3

saraamiiri:


Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -49,3 +45,39 @@ const fiveofHearts = getCardValue("5♥");
...
+
+const queenOfSpades = getCardValue("Q♠");
+assertEquals(queenOfSpades, 10);
+
+const kingOfSpades = getCardValue("K♠");
+assertEquals(kingOfSpades, 10);
+
+// Test for invalid card
+try {
+  getCardValue("X♠");
```
mhmdk, line 80: if GetCardValue() didn't throw an exception, and returned a value like 10 for example, wouldn't the test still pass?

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
```diff
@@ -1,11 +1,16 @@
...
+  expect(isProperFraction(2, 3)).toBe(true);
 });
 
-// Case 2: Identify Improper Fractions:
+test("Identifies Improper Fraction", () => {
+  expect(isProperFraction(5, 2)).toBe(false);
+});
 
-// Case 3: Identify Negative Fractions:
+test("Identifies Negative Proper Fraction", () => {
```
mhmdk, line 10: should also have a test for negative improper fraction.

Haseeb-Khalil (approved):
Looks Good

PR #377, opened 2025-03-11
ITP -Northwest- JAN 2025 | Jan Lo | Module-Structuring-and-Testing-Data Sprint 3 | WEEK 5

JanLlw:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -9,6 +9,12 @@
 
 function getAngleType(angle) {
    if (angle === 90) return "Right angle";
+   else if (angle > 0 && angle < 90) return "Acute angle";
+   else if (angle > 90 && angle < 180) return "Obtuse angle";
+   else if (angle === 180) return "Straight angle";
+   else if (angle > 180 && angle < 360) return "Reflex angle";
+   
+   
    // read to the end, complete line 36, then pass your test here
 }
 
@@ -44,13 +50,19 @@ assertEquals(acute, "Acute angle");
 // Then the function should return "Obtuse angle"
 const obtuse = getAngleType(120);
 // ====> write your test here, and then add a line to pass the test in the function above
+assertEquals(obtuse, "Obtuse angle");
 
 // Case 4: Identify Straight Angles:
 // When the angle is exactly 180 degrees,
 // Then the function should return "Straight angle"
 // ====> write your test here, and then add a line to pass the test in the function above
+const straight = getAngleType(180);
+assertEquals(straight, "Straight angle");
 
 // Case 5: Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
 // ====> write your test here, and then add a line to pass the test in the function above
+const reflex = getAngleType(270);
+assertEquals(reflex, "Reflex angle");
+
--- a/Sprint-3/1-key-implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-key-implement/2-is-proper-fraction.js
@@ -9,6 +9,11 @@
 
 function isProperFraction(numerator, denominator) {
     if (numerator < denominator) return true;
+    else if (numerator > denominator) return false;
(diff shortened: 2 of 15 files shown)


alastair87:
@JanLlw Please include an explanation of your PR as this is critical to us being able to review it meaningfully. Once you've done that I or another volunteer will look at it again.

Sprint-3/1-key-implement/1-get-angle-type.js
```diff
@@ -9,6 +9,12 @@
 
 function getAngleType(angle) {
    if (angle === 90) return "Right angle";
+   else if (angle > 0 && angle < 90) return "Acute angle";
+   else if (angle > 90 && angle < 180) return "Obtuse angle";
+   else if (angle === 180) return "Straight angle";
+   else if (angle > 180 && angle < 360) return "Reflex angle";
+   
+   
    // read to the end, complete line 36, then pass your test here
```
alastair87, line 18: Excellent. One thing you could look into as an extension exercise is to think about what might happen if `angle` were not a number. JavaScript has functionality for checking this and returning an error that you could learn about, and using it would make your code more robust in the real world.

Sprint-3/1-key-implement/2-is-proper-fraction.js
```diff
@@ -9,6 +9,11 @@
 
 function isProperFraction(numerator, denominator) {
     if (numerator < denominator) return true;
+    else if (numerator > denominator) return false;
+    else if (-numerator > denominator) return true;
+    else if (numerator === denominator) return false;
+    else return "Error";
+ 
 }
```
alastair87, line 17: Good. You could again thing about type safety here.

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -8,9 +8,22 @@
...
+  if (rank === "3") return 3;
+  if (rank === "4") return 4;
+  if (rank === "5") return 5;
+  if (rank === "6") return 6;
+  if (rank === "7") return 7;
+  if (rank === "8") return 8;
+  if (rank === "9") return 9;
+  if(["10", "J", "Q", "K"].includes(rank)) return 10;
+  throw new Error("Invaild card rank")
 }
```
alastair87, line 23: This works. You might want to consider if `if` is the cleanest way of doing this. JavaScript has another control flow statement known as a `switch` block which may be more elegant if you have many conditions like above (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch). Either works.

Sprint-3/3-mandatory-practice/implement/count.js
```diff
@@ -1,5 +1,10 @@
...
-}
-
+    let count = 0;
+    for (let char of stringOfCharacters) {
+      if (char === findCharacter) {
+        count++;
+      }
+    }
+    return count;
+  }
```
alastair87, line 9: Good.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
```diff
@@ -1,5 +1,14 @@
...
+if (num === 11 || num ===12 || num ===13) return num + "th";
+const lastDigit = num % 10;
+if (lastDigit === 1) return num + "st";
+else if (lastDigit === 2) return num + "nd";
+else if (lastDigit === 3) return num + "rd";
+else return num + "th";
 }
 
+
+
```
alastair87, line 11: Good though it's important to format your code correctly for legibility and some languages like Python require it. You can do this manually or else there is a 'Format Document' option in VS Code which is probably preferable.

Sprint-3/3-mandatory-practice/implement/repeat.js
```diff
@@ -1,5 +1,7 @@
-function repeat() {
-    return "hellohellohello";
+function repeat(str, count) {
+    if (count < 0) throw new Error("Error");
+    return str.repeat(count);
+
```
alastair87, line 4: You're not decrementing count so this function will recurse infinitely. THere needs to be a stopping condition that is reached.

PR #463, opened 2025-05-30
North West | ITP-May-25 | Geraldine Edwards | Sprint-1

Geraldine-Edwards:
## Learners, PR Template

Self checklist

## Changelist

This PR includes my responses to the Key Exercises, Mandatory Errors, Mandatory Interpret and the Stretch Explore files.

## Questions


Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -25,3 +25,18 @@ console.log(`£${pounds}.${pence}`);
...
+
+4. const pounds = paddedPenceNumberString.substring(0, paddedPenceNumberString.length - 2): Firstly, the length() method is used again and the length of paddedPenceNumberString (which is "399") is 3. We then subtract - 2 from that value, which gives us 1. The substring() method is then called on paddedPenceNumberString, extracting the characters from index 0 to index 1 (but NOT including index 1) of paddedPenceNumberString, which gives us "3". So pounds is now "3".
```
CameronDowner, line 33: Do we call a method to get the length of a string?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,45 @@ console.log(result);
...
+/*
+I experimented with various values and received the following results when changing the value of movieLength 1. movieLength = 900, result was "0:15:0" (0 hours, 15 minutes, 0 seconds)
+2. movieLength = -3600, result was "-1:0:0" (negative 1 hour, 0 minutes, 0 seconds)
+3. movieLength = 0, result was "0:0:0" (0 hours, 0 minutes, 0 seconds)
+4. movieLength = -457.93, result was "0:-7:-37.93000000000001" (negative 7 minutes, negative 37.93 seconds)
+5. movieLength = true, result was "0:0:1" (0 hours, 0 minutes, 1 second)
+6. movieLength = "movie", result was "NaN:Nan:Nan" (not a number, not a number, not a number)
+This code does provide returned values however as you can see from the list above, it does not give expected or meaningful results for all values of movieLength. For example, negative values, boolean values and string values do not return meaningful results. The code is designed to work with positive integer values representing the length of a movie in seconds, so although we will receive a result for other variations of values in movieLength, they will not be meaningful or useful
```
CameronDowner, line 55: How could we stop the program from continuing if someone puts in the wrong type of value?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,45 @@ console.log(result);
...
+The variable result represents the total length of the movie in hours, minutes and seconds in other words a hh:mm:ss format. A better name for this variable could be formattedMovieLength, as this describes the value of that variable more accurately. The word "result" is a very vague term and does not give any indication of what the variable actually represents.
+*/
 
 // f) Try experimenting with different values of movieLength. Will this code work for all values of movieLength? Explain your answer
+/*
+I experimented with various values and received the following results when changing the value of movieLength 1. movieLength = 900, result was "0:15:0" (0 hours, 15 minutes, 0 seconds)
+2. movieLength = -3600, result was "-1:0:0" (negative 1 hour, 0 minutes, 0 seconds)
+3. movieLength = 0, result was "0:0:0" (0 hours, 0 minutes, 0 seconds)
+4. movieLength = -457.93, result was "0:-7:-37.93000000000001" (negative 7 minutes, negative 37.93 seconds)
+5. movieLength = true, result was "0:0:1" (0 hours, 0 minutes, 1 second)
```
CameronDowner, line 53: What do you think JavaScript is doing behind the scenes `true` turns into `0:0:1`?

Sprint-1/3-mandatory-interpret/2-time-format.js
Geraldine-Edwards: That's a great question...I did a bit of searching online and on Stack Overflow there was a suggestion to a member to use a conditional statement in their similar program. So for this code, maybe we could add a check at the start to make sure that the input is only a positive whole number and that the movie length is a minimum number of seconds, and that the input is not a boolean  or if they are we could return an error message. 
I'm not 100% sure of the syntax I could use but I imagine it is along the lines of:

If (movieLength !== Number || movieLength < 0 || movieLength === "boolean" || movieLength !== Number.isInteger(movieLength) {
console.error("Error: the movie length must be a positive whole number in seconds only.")
}

// then the existing code starts...

Or something to that effect?

Sprint-1/3-mandatory-interpret/2-time-format.js
CameronDowner: This is exactly right. When we perform math operators on other types they can be coerced into numeric values.

You can see it in action with this example, or something as simple as `true + true = 2`

Sprint-1/3-mandatory-interpret/2-time-format.js
CameronDowner: Yeah, this is exactly how we could validate the input before we try and use it. We typically do this at the start of the program to prevent unexpected things happening - like we've seen in this program.

It finished 'correctly' but the output doesn't make sense.

This could be written as:

```javascript
if (!Number.isInteger(movieLength) || movieLength < 0) {
  // exit program
}
```

`Number.isInteger` checks it's a `number` type & that it's a whole number.

If you wanted to check the type explicitly, you can use the `typeof` [operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof) on variables or values.

```javascript
typeof 1; // "number"

const name = "Cameron";
typeof name; // "string"
```

CameronDowner:
This is looking really good!

PR #511, opened 2025-06-10
LONDON | May-2025 | Anuar Duisenbek | Module-Structuring-and-Testing-Data | Sprint-1

AnuarBey:
## Learners, PR Template

Self checklist

## Changelist

 Resolved all errors in the Errors section: ran files with Node.js... interpreted the error messages, and fixed the bugs using error documentation

 Interpreted and analyzed larger programs in the Interpret section: used console.log and MDN to understand syntax, operators, and functions

### DIFF
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -7,3 +7,9 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+// This code generates a random whole number between 1 and 100.
+// Math.random() returns a decimal between 0 and 1 (not including 1).
+// We multiply it by 100 to get a number between 0 and 99.999...
+// Math.floor() removes the decimal part, so we get an integer.
+// We add 1 to shift the range to 1–100 instead of 0–99.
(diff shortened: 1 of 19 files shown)


Sprint-1/1-key-exercises/4-random.js
```diff
@@ -7,3 +7,9 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+/* This code generates a random whole number between 1 and 100.
+ Math.random() returns a decimal between 0 and 1 (not including 1).
```
cjyuan, line 12: You can also use the concise and precise ***interval notation*** to describe a range of values.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -23,5 +23,37 @@ console.log(`£${pounds}.${pence}`);
...
+  0,
+  paddedPenceNumberString.length - 2
+);
+
+Extract the pound part (everything except the last 2 characters)
+
+5.const pence = paddedPenceNumberString
+  .substring(paddedPenceNumberString.length - 2)
+  .padEnd(2, "0");
+  Extract the pence part (last 2 characters), and make sure it's 2 digits long
```
cjyuan, line 53: Could we expect this program to work as intended for any valid `penceString` if we deleted `.padEnd(2, "0")` from the code?

cjyuan (commented):
You could consider adding `.idea` into `.gitignore` so that the folder won't get pushed to GitHub. The files in that folder don't appear to be necessary for the exercises.

Explanation is clear . 

I only have one question for your regarding usage of `.padEnd()`.

AnuarBey:
Thank you for pointing out the .padEnd() usage. After reviewing the code, I realised it isn’t really necessary here, as the output is correct for all valid inputs even without it. I’ve removed .padEnd() as suggested — thanks for the reminder!

Also, I have now added .idea/ to .gitignore so it won’t be pushed to the repository in the future.

PR #514, opened 2025-06-11
London|May-2025|Alexandru Pocovnicu|Coursework/sprint 2

alexandru-pocovnicu:
## Learners, PR Template

Self checklist

- [ ] I have titled my PR with REGION | COHORT_NAME | FIRST_NAME LAST_NAME | PROJ_NAME

## Changelist

In this PR I have  fixed the errors, done the debugging, implemented functions, interpreted larger programs and wrote tests for edge cases

## Questions

### DIFF
--- a/Sprint-2/5-stretch-extend/format-time.js
+++ b/Sprint-2/5-stretch-extend/format-time.js
@@ -4,8 +4,15 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
-  if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+  const minutes=(time.slice(3,5));
+  if(hours===24){
+    return `00:${minutes} am`
+  }
+  else if(hours===12){
+    return `12:${minutes} pm`
+  }
+  else if (hours > 12) {
+    return `${hours - 12}:${minutes} pm`;
   }
   return `${time} am`;
 }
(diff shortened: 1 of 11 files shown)


Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,8 +4,15 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
-  if (hours > 12) {
```
bodfaj13, line 7: Nice start! I see you've made progress in converting 24-hour time to 12-hour format, and your tests for "08:00" and "23:00" are passing — that's a good sign.

A few things to consider:

Your current implementation seems to only work for hh:00 times. What happens if the input is "23:45" or "13:30"?

Try testing edge cases like "00:00", "12:00", and "12:59". Do those return the expected outputs?

Think about how you're handling the minutes — right now, it looks like you're assuming they’re always 00.

Also, what would happen if the hour is 00 or 12? These are usually special cases when converting to a 12-hour format.

I’d suggest expanding your test cases to cover different times across the day — especially early morning, noon, and late night — and seeing what breaks. Once you've done that, it should help guide the necessary fixes in your function.

Let me know if you get stuck anywhere — happy to nudge you in the right direction!

bodfaj13: I noticed on line 8 you used 00 as a number. Just a heads-up: in JavaScript, numbers that start with 0 can be interpreted as octal (base-8) literals. However, using 00 like that isn’t allowed in modern JavaScript (strict mode), and it will throw a syntax error.
Thank you; you are almost there.
Cheers.

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,8 +4,16 @@
...
-    return `${hours - 12}:00 pm`;
+  const minutes=(time.slice(3,5));
+  if(hours===00)return `12:${minutes} am`
```
bodfaj13, line 8: I noticed on line 8 you used 00 as a number. Just a heads-up: in JavaScript, numbers that start with 0 can be interpreted as octal (base-8) literals. However, using 00 like that isn’t allowed in modern JavaScript (strict mode), and it will throw a syntax error.

If you meant zero, just use 0 instead of 00.
Good job so far, you are almost there.

bodfaj13 (approved):
Thanks for being so meticulous.

PR #517, opened 2025-06-11
London | ITP-May-2025 | Seddiq Azam | Module-Structuring-and-Testing-Data | Sprint-1

sedazam:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.
This pull request concerns the sprint-1 tasks for the module. 
Implemented various functions to generate random integers within a specified range, minimum, and maximum. 
used Math.random() scaled to the range size, with Math.floor() to ensure an integer results. 
identified common errors in the code

## Questions

### DIFF
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -7,3 +7,27 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+
+// Answers:
+
+// num is a variable that returns a random number between 1 and 100, inclusive. 
+
+// - Math.floor(); rounds down a number to its nearest integer. 
+// - Math.random(); it returns an unexpected value within the specified range. Usually it is between 0 and 1, but excluding 1. 
+// - (maximum - minimum + 1); This formula represents the range. Maximum means the maximum number minus the minimum number + 1. 
+// - minimum; this is the minimum number added to the formula to generate a number within the range. 
+
+// The formula works as follows:
+
+//	•	We use the formula (maximum - minimum + 1) to calculate the size of the desired range.
+//    In the example above, maximum = 100 and minimum = 1.
+//    So: 100 - 1 = 99, then 99 + 1 = 100.
+//	•	Math.random() generates a decimal number between 0 (inclusive) and 1 (exclusive).
+//    For example: 0.0001, 0.5342, 0.9999.
+// 	•	We multiply the result of Math.random() by (maximum - minimum + 1) to scale it to the desired range.
+//    Example: 0.9999 * 100 = 99.99.
+// 	•	We then use Math.floor() to round the result down to the nearest whole number.
+//    Example: Math.floor(99.99) = 99.
+// 	•	Finally, we add the minimum value to shift the result into the correct range.
+//    Example: 99 + 1 = 100.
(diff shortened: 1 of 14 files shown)


Sprint-1/1-key-exercises/4-random.js
```diff
@@ -7,3 +7,21 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
...
+
+//  num is a variable that stores a random integer between 1 and 100, inclusive.
+//	•	Math.floor() rounds a number down to the nearest integer.
+//	•	Math.random() generates a decimal number between 0 (inclusive) and 1 (exclusive).
+//	•	(maximum - minimum + 1) calculates the size of the range of possible numbers. Here, it is 100 - 1 + 1 = 100.
+//	•	minimum is added at the end to shift the result so that it falls within the desired range starting from the minimum value.
+
+// How the formula works:
+//	1.	Calculate the size of the range using (maximum - minimum + 1). In this example, with maximum = 100 and minimum = 1, the range size is 100 - 1 + 1 = 100.
+//	2.	Math.random() generates a decimal number between 0 (inclusive) and 1 (exclusive), such as 0.0001 or 0.9999.
```
cjyuan, line 22: We can also use the concise and precise ***interval notation*** to describe a range of values.

***Returns*** is the standard technical term for what a function outputs. It aligns with programming language terminology, especially in documentation and developer communication.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -25,3 +25,31 @@ console.log(`£${pounds}.${pence}`);
...
+// This line extracts the pound portion of the value by taking all characters except the last two.
+// For example, if the input is "399p", it becomes "399", and this line extracts "3" as the pound value.
+
+
+// 5. const pence = paddedPenceNumberString
+//    .substring(paddedPenceNumberString.length - 2)
+//    .padEnd(2, "0");
+// This line extracts the last two digits, which represent the pence.
+// The `.padEnd(2, "0")` ensures that even if the pence value is only one digit (e.g., "5"), it becomes two digits (e.g., "05").
+// So "5p" would correctly display as "£0.05" instead of "£0.5".
```
cjyuan, line 49: Could we expect this program to work as intended for any valid `penceString` if we deleted `.padEnd(2, "0")` from the code?

Sprint-2/1-key-errors/0.js
cjyuan, line 1: To keep this branch clean, can you revert the changes made to this file? It is not related to the Sprint-1 exercise.

cjyuan (commented):
Explanation is clear. Well done.

I only have a challenge question for you.

Sprint-1/1-key-exercises/4-random.js
sedazam: Thank you, this is quite a helpful tip. 
I learned that the symbol [ means the value is included. For example, [2, 5] includes both 2 and 5. 
The symbol ( means the value is excluded. For example, (2, 5) excludes both 2 and 5.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
sedazam: .padEnd(2, "0") ensures that the pence value always has two digits, by padding with a zero on the right if necessary. 
if we remove .padEnd(2, "0"), the value "5p" will be displayed as "£0.5" 
.substring(paddedPenceNumberString.length - 2) will return just "5" when the input is only one digit long. 
This leads to incorrect currency formatting, since "£0.5" means 50p, not 5p.

Sprint-2/1-key-errors/0.js
sedazam: Done.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
sedazam: I tested the code with different values. When I used "999p", the output was "£9.99" — both with and without .padEnd(2, "0").

After doing some research, I found that the line:

returns the last two characters of the string. In this case, it correctly returns "99", which is already two characters long.

Therefore, using .padEnd(2, "0") is unnecessary in this situation.

PR #525, opened 2025-06-12
LONDON | ITP-MAY-2025 | Hibo Sharif | Module-Structuring-and-Testing-Data | Sprint 2

hibosharif202504:
##Self checklist

## Changelist
I have committed the files and fixed the errors in the codes.

## Questions

### DIFF
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -1,8 +1,19 @@
+
 // Predict and explain first...
 //  =============> write your prediction here
+/* When I run the code in the terminal it comes back as an error in both location 
+let str -- error indicates that 'str' has already been declared. I can see that the'str' 
+variable was declared in the function. This cannot be duplicated and it will result in error.
+ return -- this function has also resulted in error, 'illegal return statement, as result to the 
+ error in the code.*/
 
 // call the function capitalise with a string input
 // interpret the error message and figure out why an error is occurring
+/* When I added an input string (week) to the function capitalise I got the same error as the string 
+is identified in two location with the function and again inside the function, I have looked at the documentation 
+in this website https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let  
+Its one of the rules in Javascript that doesn't allow function parameters to be redeclared by let or const, this if 
+executed will output error syntax Identifier has already been declared. */
 
 function capitalise(str) {
   let str = `${str[0].toUpperCase()}${str.slice(1)}`;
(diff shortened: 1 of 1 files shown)


Sprint-2/1-key-errors/1.js
```diff
@@ -18,3 +26,13 @@ console.log(decimalNumber);
...
 // =============> write your new code here
+
+function convertToPercentage(decimalNumber) {
+  const percentage = `${decimalNumber * 100}%`;
+  return percentage;
+}
+const decimalNumber = 0.5;
+console.log(convertToPercentage(decimalNumber));
+
+/* In the corrected code the (decimalNumber) is only declared in the function.
```
cjyuan, line 37: You also have `decimalNumber` declared at line 34!

Your code correction is spot on. 

Could you try using AI to see how else we can explain why moving line 17 out of the function resolve the problem?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -16,4 +16,8 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    let heightM = height/ 100; // converting height from meter to centimeter 
+    let bmi = weight/Math.pow(heightM,2);
+    return bmi.toFixed(1);
```
cjyuan, line 21: What **type** of value do you expect the function to return? A number or a string?
Does your function return the type of value you expect it to return?

Sprint-2/3-mandatory-implement/3-to-pounds.js
```diff
@@ -4,3 +4,25 @@
...
+  0,
+  penceString.length - 1);
+
+const paddedPenceNumberString = penceStringWithoutTrailingP.padStart(3, "0");
+
+const pounds = paddedPenceNumberString.substring( 0, paddedPenceNumberString.length - 2);
+
+const pence = paddedPenceNumberString
+  .substring(paddedPenceNumberString.length - 2)
+  .padEnd(2, "0");
```
cjyuan, line 20: Can you improve the indentation of these few lines of code?

Sprint-2/1-key-errors/1.js
hibosharif202504: Hi, 
const decimalNumber = 0.5; is incorrect at line 17, because decimalNumber is a parameter that has been  declared inside the function and it would be wrong to redeclare it again with value of const decimalNumber = 0.5;
This line shadows the parameter. It ignores whatever value was passed into the function and forces it to always use 0.5. That defeats the purpose of accepting an argument.

Sprint-2/3-mandatory-implement/1-bmi.js
hibosharif202504: The expected value is a string because .toFixed(1) returns string.
To get a number as an output we can use return Number(bmi.toFixed(1));
so if the desired output is a number rather than a string the .toFixed can be wrapped by Number()

PR #541, opened 2025-06-13
NW | ITP-May-25 | Geraldine Edwards | Module-Structuring-and-Testing-Data | Sprint-3

Geraldine-Edwards:
## Learners, PR Template

Self checklist

## Changelist

This PR includes my responses and code to questions in the Key Implement, Mandatory Rewrite, Mandatory Practice  and the Stretch Investigate files.

## Questions

### DIFF
--- a/Sprint-3/1-key-implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-key-implement/2-is-proper-fraction.js
@@ -8,7 +8,9 @@
 // write one test at a time, and make it pass, build your solution up methodically
 
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
+  // ensure that the both inputs are evaluated to positive numbers by using Math.abs(), eg -4 returns 4, 4 returns 4, and 0 returns 0
+  if (Math.abs(numerator) < Math.abs(denominator)) return true;
+  return false; //if the inputs don't satisfy the expression it returns false
 }
 
 // here's our helper again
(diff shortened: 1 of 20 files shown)


Sprint-3/1-key-implement/2-is-proper-fraction.js
```diff
@@ -8,7 +8,9 @@
 // write one test at a time, and make it pass, build your solution up methodically
 
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
+  // ensure that the both inputs are evaluated to positive numbers by using Math.abs(), eg -4 returns 4, 4 returns 4, and 0 returns 0
+  if (Math.abs(numerator) < Math.abs(denominator)) return true;
+  return false; //if the inputs don't satisfy the expression it returns false
```
CameronDowner, line 13: Do we need an `if` statement here?

`Math.abs(numerator) < Math.abs(denominator)` evaluates to `true` or `false` already

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -8,7 +8,22 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
-    if (rank === "A") return 11;
+  // use an array to hold the values of the face cards.
+  const faceCards = ["10", "J", "Q", "K"];
```
CameronDowner, line 12: This will work as expected - but is `10` a face card?

If I wanted to extend this functionality & used this array in another method `isFaceCard`, this would start giving me the wrong answer

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
```diff
@@ -5,7 +5,16 @@ test("should return true for a proper fraction", () => {
 });
 
 // Case 2: Identify Improper Fractions:
+test("should return true for a improper fraction", () => {
+  expect(isProperFraction(5, 2)).toEqual(false);
```
CameronDowner, line 9: Jest also has `.toBeTruthy()`: https://jestjs.io/docs/expect#tobetruthy

Why might that **not** be a good thing to use here?

CameronDowner (commented):
Really strong coursework - you are using some advanced practises here (e.g. parameterised tests, breaking down functions, throwing & testing for errors)

I have left some more advanced questions, but will mark as `Complete` because you've met all the criteria already

Sprint-3/1-key-implement/2-is-proper-fraction.js
Geraldine-Edwards: Ah yes, I can see now that I can just return the expression instead of using the if statement - I'll fix this now. :)

Sprint-3/1-key-implement/3-get-card-value.js
Geraldine-Edwards: oh gosh of course! I think I got confused as I was just thinking about the values of the cards being 10. I've removed 10 from the array and adjusted the following if statement to cover numbers 2-10.

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
Geraldine-Edwards: I imagine it's not great to use at this test probably because if something is "truthy" it is less strict.. or something to that effect? I would need to make sure that an improper fraction was handled strictly to get the exact correct values.

PR #562, opened 2025-06-17
LONDON | ITP-MAY-2025 | Hibo Sharif | Module-Structuring-and-Testing-data | Sprint 3

hibosharif202504:
## Learners, PR Template

Self checklist

## Changelist

I have implemented functions and run tests on the file to verify that it is running correctly.

## Questions

### DIFF
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -8,13 +8,21 @@
 // Then, write the next test! :) Go through this process until all the cases are implemented
 
 function getAngleType(angle) {
-   if (angle === 90) return "Right angle";
-   if (angle < 90) return "Acute angle";
-   if (angle > 90 && angle < 180) return "Obtuse angle";
-   if (angle === 180) return "Straight angle";
-   if (angle > 180 && angle < 360) return "Reflex angle";
-   return "Invalid angle"; 
-// read to the end, complete line 36, then pass your test here
+  if (angle === 90) {
+    return "Right angle";
+  } else if (angle < 90) {
+    return "Acute angle";
+  } else if (angle > 90 && angle < 180) {
+    return "Obtuse angle";
+  } else if (angle === 180) {
+    return "Straight angle";
+  } else if (angle > 180 && angle < 360) {
+    return "Reflex angle";
+  } else {
+    return "Invalid angle";
+  }
+
+  // read to the end, complete line 36, then pass your test here
 }
 
 // we're going to use this helper function to make our assertions easier to read
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -9,7 +9,12 @@
 
 function getAngleType(angle) {
    if (angle === 90) return "Right angle";
-   // read to the end, complete line 36, then pass your test here
+   if (angle < 90) return "Acute angle";
+   if (angle > 90 && angle < 180) return "Obtuse angle";
+   if (angle === 180) return "Straight angle";
+   if (angle > 180 && angle < 360) return "Reflex angle";
+   return "Invalid angle"; 
+// read to the end, complete line 36, then pass your test here
 }
 
 // we're going to use this helper function to make our assertions easier to read
@@ -43,14 +48,19 @@ assertEquals(acute, "Acute angle");
 // When the angle is greater than 90 degrees and less than 180 degrees,
 // Then the function should return "Obtuse angle"
 const obtuse = getAngleType(120);
+assertEquals(obtuse, "Obtuse angle");
 // ====> write your test here, and then add a line to pass the test in the function above
 
 // Case 4: Identify Straight Angles:
 // When the angle is exactly 180 degrees,
 // Then the function should return "Straight angle"
+const straight = getAngleType(180);
+assertEquals(straight, "Straight angle");
 // ====> write your test here, and then add a line to pass the test in the function above
 
 // Case 5: Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
+const reflex = getAngleType(270);
+assertEquals(reflex, "Reflex angle");
 // ====> write your test here, and then add a line to pass the test in the function above
--- a/Sprint-3/4-stretch-investigate/card-validator.js
+++ b/Sprint-3/4-stretch-investigate/card-validator.js
@@ -1,0 +1,43 @@
+function cardValidator(cardNumber) {
+    //Check if length is exactly 16
+    if (cardNumber.length !== 16) {
+      return false;
+    }
+  
+    //Check all characters are digits
+    for (let i = 0; i < cardNumber.length; i++) {
+        if (cardNumber[i] < '0' || cardNumber[i] > '9') 
+            return false;
+      }
+    
+    // Check not all digits are the same
+    
+    let allSame = true;
+for (let i = 1; i < cardNumber.length; i++) {
+  if (cardNumber[i] !== cardNumber[0]) {
+    allSame = false;
+    break;
+  }
+}
+  if (allSame) {
+    return false;
+  }
+    //Check sum of all digits is greater than 16
+    const sum = cardNumber
+      .split("")
+      .reduce((acc, digit) => acc + Number(digit), 0);
+    if (sum <= 16) {
+      return false;
+    }
+  
+    //Check last digit is even
+    const lastDigit = Number(cardNumber[cardNumber.length - 1]);
+    if (lastDigit % 2 === 0) {
+      return false;
+    }
+  
+    return true;
+  }
+  
+  module.exports = cardValidator;
+  
--- a/Sprint-3/4-stretch-investigate/card-validator.js
+++ b/Sprint-3/4-stretch-investigate/card-validator.js
@@ -1,8 +1,10 @@
 function cardValidator(cardNumber) {
   //Check if length is exactly 16
-  if (cardNumber.length !== 16) {
-    return false;
-  }
+  if (cardNumber.length !== 16) return false;
+
+  const lastDigit = Number(cardNumber[cardNumber.length - 1]);
+  if (lastDigit % 2 === 0) return false;
+
   let sum = 0;
   let allSame = true;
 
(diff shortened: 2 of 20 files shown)


Sprint-3/4-stretch-investigate/card-validator.js
```diff
@@ -0,0 +1,43 @@
...
+    
+    // Check not all digits are the same
+    
+    let allSame = true;
+for (let i = 1; i < cardNumber.length; i++) {
+  if (cardNumber[i] !== cardNumber[0]) {
+    allSame = false;
+    break;
+  }
+}
```
rohan-corbyn, line 21: Given the structure of the rest of this function i.e that false is returned as soon as a check fails.
What is the purpose of storing allSame in a boolean value for this check? The previous check did not need a variable called charactersAreAllDigits.
There is nothing wrong with either approach but it's best to write code in a consistent style, especially within the same function.

Sprint-3/4-stretch-investigate/card-validator.js
```diff
@@ -0,0 +1,43 @@
...
+  if (allSame) {
+    return false;
+  }
+    //Check sum of all digits is greater than 16
+    const sum = cardNumber
+      .split("")
+      .reduce((acc, digit) => acc + Number(digit), 0);
```
rohan-corbyn, line 28: Nice to see some functional programming being utilised here with the reduce() method but it also follows a different style to the rest of this function. 

Is using split("") to convert the String into an array absolutely necessary here? Previously you took a different approach and iterated through the characters of the String without needing to convert the String into an array first which is less space efficient but more importantly less consistent with the rest of the coding style in this function.

Sprint-3/4-stretch-investigate/find.js
```diff
@@ -20,6 +20,21 @@ console.log(find("code your future", "z"));
...
+
+
 // b) What is the if statement used to check
+// it's checking if the character at the position of the string matches what function is looking for, when 
+// it reaches that character the value is returned, in this case it's 7 and when it's no longer found it comes
+// back with -1.
+
 // c) Why is index++ being used?
+// it is used to add 1 to the index variable every time it goes through the loop to the point the character is found.
+
```
rohan-corbyn, line 36: Perfectly accurate. Do you know the name of the ++ operator? It can be a useful term to know when describing this behaviour.

Sprint-3/4-stretch-investigate/password-validator.js
```diff
@@ -1,5 +1,13 @@
...
+function passwordValidator(password, previousPasswords = []) {
+    if (password.length < 5) return false;
+
+    const passwordUpper = /[A-Z]/.test(password);
+    const passWordLower = /[a-z]/.test(password);
+    const passwordNumber =/[0-9]/.test(password);
+    const passwordSpecial =/[!#$%.&*]/.test(password);
+    const notPreviousPassword = !previousPasswords.includes(password);
+
+    return passwordUpper && passWordLower && passwordNumber && passwordSpecial && notPreviousPassword
```
rohan-corbyn, line 10: This is a great way to write a validator. Storing each of the boolean checks in individual variables and returning a condition that combines them all at the end. 

Given that you have taken this approach for passwordUpper, passWordLower, passwordNumber etc. 
Is there a way you could store the result of the first check "password.length < 5" in a variable and include it in the final returned condition in the same way? 
Can you see how this would be more consistent with the coding style used in the rest of the function?

rohan-corbyn (changes requested):
Great work! - Some really good solutions here. 
It's great to see you exploring a range of coding styles including for..of loops and functional programming techniques such as reduce(). I'd only suggest that you try and use one coding style / approach at a time within a single function where possible as it will make the code clearer and help you identify opportunities for refactoring and improving efficiency.

Sprint-3/4-stretch-investigate/password-validator.js
hibosharif202504: Hi, @rohan-corbyn

I have created the variable lengthIsValid and included it in the final return condition.

Sprint-3/1-key-implement/1-get-angle-type.js
```diff
@@ -9,7 +9,12 @@
 
 function getAngleType(angle) {
    if (angle === 90) return "Right angle";
-   // read to the end, complete line 36, then pass your test here
+   if (angle < 90) return "Acute angle";
+   if (angle > 90 && angle < 180) return "Obtuse angle";
+   if (angle === 180) return "Straight angle";
+   if (angle > 180 && angle < 360) return "Reflex angle";
+   return "Invalid angle"; 
+// read to the end, complete line 36, then pass your test here
```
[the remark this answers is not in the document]
hibosharif202504: Hi, @rohan-corbyn

The conditional statement can be changed to if.. else or switch. I have changed it with if..else statement.

If another developer modified the code using console.log() it will not return nothing it will return undefined. The console.log()
would need an if..else conditional statement to output one message at a time, if we use the if conditional statement with the console.log more than one messages might be true or checked resulting in multiple outputs. 
Also using the console.log can cause bugs in the code.

Sprint-3/4-stretch-investigate/card-validator.js
```diff
@@ -1,43 +1,34 @@
...
-    break;
+    if (cardNumber[i] !== cardNumber[0]) {
+      allSame = false;
+    }
   }
-}
-  if (allSame) {
+  if (allSame) return false;
+  if (sum <= 16) return false;
+
```
rohan-corbyn, line 25: Nice work moving everything into 1 for loop! Thanks for taking my suggestions onboard. As the efficiency improvement would likely be requested of you in a professional context, it makes this very good practice.

**Required Fixes:**
From looking at the code one bug may have been introduced. Seeing as your for-loop now begins at 0, will the condition on line 19 ever return false during the first iteration of the loop? You need to fix this by adding another condition into the if statement on line 19. If I have miss-read something and there isn't a bug, please let me know.

**Required Efficiency Improvements:**
You have made it better but two modifications could really help.

1. line 24: Consider the if statement that returns false if sum <= 16.  If we move this check into the for-loop we could exit out of the for-loop as soon as we know the sum is going to be over 16. Imagine if the number started with 98, we will have taken only 2 loops instead of 16.

2. Line 26: The check for "if the last digit is even" is a fast operation as we are assessing a single digit instead of looping through all of the digits like we are above. It would be more sensible to place it before the for-loop.

For-loops are expensive processes. They involve repeating a lot of steps. When looking for performance enhancements and assessing the impact a change, developers often analyse existing for-loops. There are often opportunities for substantial improvements and a high chance of bugs. The "if statement here" on the other hand, is just a single check. Move it before the for-loop and you could avoid looping through the entire string if the validation for "last digit isn't even" returns false.

**Syntax Improvements:**

1. Curly braces used on line 19 aren't consistent with your other if statements in the loop and function.

2. The examples below show another style you could use to write all of the "if - return" statements in a more concise way:

Instead of:
` if (allSame) return false; `
 I would write:
` return !allSame;`
 
Another example but can apply to all of them:
` if (lastDigit % 2 === 0) { return false }`
 I would write:
` return lastDigit % 2 !== 0;`

allSame, and lastDigit % 2 !== 0 are both boolean values: returning either "true" or "false". Just sharing this with you. You can choose a style you prefer and that you feel works best in a given use case. It will become intuitive with practice.

Again, very good work so far! Just a few requested changes here. My apologies for the long comment.

Note: I can see the logic for allSame has changed slightly but it looks like this was a "bugfix" and not a regression. Just double checking, thanks

rohan-corbyn (changes requested):
Very positive changes made with the last commit!
I can see you've understood my suggestions well and have worked hard to improve the code. I have made a few more requests for you to go through because I feel they are valuable opportunities to learn and practice crucial concepts.
thanks for your effort so far!

Please request a review once you have gone through my review so that I get notified and can read your comments and code changes together. These docs might help if you are unsure how to do this:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review

Sprint-3/4-stretch-investigate/card-validator.js
hibosharif202504: Hi @rohan-corbyn, 

Yes that is right there is a need for extra condition in line 19. The for loop starts at i =0 and the comparison char !== cardNumber[0] would never be true during the first iteration because the same character being compared to itself and it will not action the allSame flag from updating correctly. I have added extra condition :
if(i > 0 && char !== cardNumber[0])
    allSame = false;
The new condition was added to ensure the comparison starts from the second characters onward. Thanks for catching this, its now resolved.

I have moved the lastDigit before the for-loop, I have moved the sum check inside the loop.
I have also used less curly braces to keep the style consistent.

I have made adjustments to the code regarding the if-return statements that you have suggested. 

Thank you for the review I have learned so much from this exercise.

Sprint-3/4-stretch-investigate/card-validator.js
```diff
@@ -1,43 +1,34 @@
...
+  for (let i = 0; i < cardNumber.length; i++) {
+    const char = cardNumber[i];
+
     //Check all characters are digits
-    for (let i = 0; i < cardNumber.length; i++) {
-        if (cardNumber[i] < '0' || cardNumber[i] > '9') 
-            return false;
-      }
-    
+    if (cardNumber[i] < "0" || cardNumber[i] > "9") return false;
```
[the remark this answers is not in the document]
hibosharif202504: I have made adjustments in the code.

rohan-corbyn (changes requested):
Hello @hibosharif202504, I appear to have miss guided you in the last review I made. 
I am genuinely sorry for any confusion I have caused.
I left a comment requesting **a single line change** to fix this.
After that I believe you have completed this sprint. Great work for persevering.

rohan-corbyn (approved):
@hibosharif202504 has worked hard to produce some high quality solutions and consider all of my suggestions. Great work for persevering and my apologies again for miss-reading the requirements one one of the problems.

Sprint-3/4-stretch-investigate/card-validator.js
hibosharif202504: Hi @rohan-corbyn ,
 I have learned so much from your review as it helped me to re-evaluate my work and focus more on the use of e.g, for-loop.

I have removed the line, and again, thank you for taking the time to review my work.

PR #564, opened 2025-06-17
Glasgow| May-2025 | Bassam Alshujaa| Sprint-2

Bassam-Radwan:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.
 I have solved and answered all files in this Sprint 
## Questions

### DIFF
--- a/Sprint-2/2-mandatory-debug/0.js
+++ b/Sprint-2/2-mandatory-debug/0.js
@@ -15,7 +15,6 @@ console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
 // Finally, correct the code to fix the problem
 //  =============> write your new code here
 function multiply(a, b) {
-  console.log(a * b);
   return a*b;
 }
 
--- a/Sprint-2/3-mandatory-implement/3-to-pounds.js
+++ b/Sprint-2/3-mandatory-implement/3-to-pounds.js
@@ -4,13 +4,13 @@
 // You will need to declare a function called toPounds with an appropriately named parameter.
 
 // You should call this function a number of times to check it works for different inputs
-function topounds(price){
+function toPounds(price){
     const penceString = price.toString()
-    const getpence = penceString.substring(0,penceString.length -1)
-    const paddedpence = getpence.padStart(3, '0') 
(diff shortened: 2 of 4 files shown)


YoanHlebarov:
Please update your PR to exclude Sprint-1 files from it!

Sprint-2/2-mandatory-debug/0.js
```diff
@@ -1,14 +1,22 @@
...
 
 console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
 
 // =============> write your explanation here
+// A return was added to the function as its value so when we call it or reuse it as a value it will be defined. 
 
 // Finally, correct the code to fix the problem
 //  =============> write your new code here
+function multiply(a, b) {
+  console.log(a * b);
```
YoanHlebarov, line 18: I would remove this line as in this case we are logging the value in the following code block.

Sprint-2/3-mandatory-implement/3-to-pounds.js
YoanHlebarov, line 1: It's a good practice to write variable names using camel case. Though this is the most common approach other options exists such as pascal case, snake case, etc.

Sprint-2/5-stretch-extend/format-time.js
YoanHlebarov, line 1: Could you write more test cases, perhaps some edge cases too? 
```
const targetOutput = "8:00 am";
const targetOutput = "28:00 am";
```
for example

YoanHlebarov (commented):
Pretty good, one main point here to do with the code - add more tests for the time function, you want to break it and throw everything at it, and only when you are happy it runs the way you expect it to with rubbish input, then you can be sure it will work well in production. 

Also make sure you only add the correct files for the PR, including some of the temp files.

Sprint-2/5-stretch-extend/format-time.js
Bassam-Radwan: I have added more test cases
Thank you @YoanHlebarov

Bassam-Radwan:
Thank you @YoanHlebarov, could you please add the complete label if it does not need any modification. I appreciate it.

PR #596, opened 2025-06-23
WESTMIDLANDS|ITP-MAY2025|SARA_TAHIR|Sprint 3 coursework

SaraTahir28:
## Learners, PR Template

Self checklist

- [X] I have committed my files one by one, on purpose, and for a reason
- [X] I have titled my PR with REGION | COHORT_NAME | FIRST_NAME LAST_NAME | PROJ_NAME
- [X] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/contributing/)
- [X] My changes meet the [requirements](./README.md) of this task

## Changelist

Briefly explain your PR.
This pull request has all the coursework with updated functions and corresponding tests which are all passing now.

## Questions

### DIFF
--- a/Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
+++ b/Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
@@ -1,13 +1,12 @@
-function getAngleType(angle) {
-   if (angle === 90) return "Right angle";
-   if(angle<90) return "Acute angle";
-   if(angle>90 && angle<180) return "Obtuse angle";
-   if(angle === 180) return "Straight angle"
-   if (angle>180 && angle <360) return "Reflex angle"
-   
-   // read to the end, complete line 36, then pass your test here
+function getAngleType(angle)  {
+  if (angle <= 0 || angle >= 360) return "Invalid angle";
+  if (angle === 90) return "Right angle";
+  if (angle < 90) return "Acute angle";
+  if (angle > 90 && angle < 180) return "Obtuse angle";
+  if (angle === 180) return "Straight angle";
+  if (angle > 180 && angle < 360) return "Reflex angle";
+  return "Invalid angle"; 
 }
-
 
 
 
--- a/Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
+++ b/Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
@@ -1,7 +1,11 @@
 function getAngleType(angle) {
-  if (angle === 90) return "Right angle";
-  // replace with your completed function from key-implement
-
+   if (angle === 90) return "Right angle";
+   if(angle<90) return "Acute angle";
+   if(angle>90 && angle<180) return "Obtuse angle";
+   if(angle === 180) return "Straight angle"
+   if (angle>180 && angle <360) return "Reflex angle"
+   
+   // read to the end, complete line 36, then pass your test here
 }
 
 
--- a/Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
+++ b/Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
@@ -1,6 +1,6 @@
 const getAngleType = require("./1-get-angle-type");
 
-  test("should identify right angle (90°)", () => {
+  /*test("should identify right angle (90°)", () => {
   expect(getAngleType(90)).toEqual("Right angle");});
 
   test("should identify acute angle (45°)", () => {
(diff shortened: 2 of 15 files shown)


Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
```diff
@@ -1,7 +1,11 @@
 function getAngleType(angle) {
-  if (angle === 90) return "Right angle";
-  // replace with your completed function from key-implement
-
+   if (angle === 90) return "Right angle";
+   if(angle<90) return "Acute angle";
+   if(angle>90 && angle<180) return "Obtuse angle";
+   if(angle === 180) return "Straight angle"
+   if (angle>180 && angle <360) return "Reflex angle"
```
cjyuan, line 6: The spec isn't clear whether `angle` can be assigned a number not in the interval (0, 360).
When `angle` is >= 360, what should the function return? (Also, by definition, angles <= 0 are not considered acute angles.)

When we implement a function that can return a value, to ensure reliability, we should ensure it will always return a defined value instead of `undefined` (which represents "no return value").
If the parameter, `angle`, is not within the recognised range, we can design the function to return a special value (e.g., "Invalid angle") or throw an error.

Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
```diff
@@ -1,10 +1,24 @@
...
+
+  test("should identify obtuse angle (120°)", () => {
+  expect(getAngleType(120)).toEqual("Obtuse angle");});
+
+  test("should identify straight angle (180°)", () => {
+  expect(getAngleType(180)).toEqual("Straight angle");});
+
+  test("should identify reflex angle (220°)", () => {
+  expect(getAngleType(220)).toEqual("Reflex angle");});
+
```
cjyuan, line 17: We could generalise each test to cover all values that belong to the same category, and pick some sample values in the category to test. For example,
```
test("should identify reflex angle (180 < angle < 360)", () => {
  expect(getAngleType(300)).toEqual("Reflex angle");
  expect(getAngleType(359.999)).toEqual("Reflex angle");
  expect(getAngleType(180.001)).toEqual("Reflex angle");
});
```

Sprint-3/3-mandatory-practice/implement/count.js
```diff
@@ -1,5 +1,14 @@
...
+    if (!stringOfCharacters.includes(findCharacter))
+    return 0
+    else {
+    let count = 0;
+    for (let char of stringOfCharacters) {
+        if(char === findCharacter){
+            count++;
+        }}
+    return count
+    }
```
cjyuan, line 11: Can you improve the indentation of this code?

Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
SaraTahir28: Sorted these issues and and considered for all values between 0-360 as well. updated the code

Sprint-3/2-mandatory-rewrite/1-get-angle-type.test.js
SaraTahir28: I have grouped the tests togther now. I understand how to do them now but 2 weeks ago when I made the assignment I didnt kmow we could write tests generically as well.

Sprint-3/2-mandatory-rewrite/3-get-card-value.js
```diff
@@ -1,5 +1,11 @@
 function getCardValue(card) {
-    // replace with your code from key-implement
-    return 11;
+  const rankChar = card.slice(0, -1); // Remove the last character (suit)
+  const rankInt = parseInt(rankChar); // Try to convert rank to a number
```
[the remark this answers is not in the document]
SaraTahir28: updated function and tests after review.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -11,3 +11,18 @@ const getOrdinalNumber = require("./get-ordinal-number");
...
+    });
+test("should return '4th' for 4", () => {
+    expect(getOrdinalNumber(4)).toEqual("4th");
+    });
+test("should return '28th' for 28", () => {
+    expect(getOrdinalNumber(28)).toEqual("28th");
+    });
+test("should return '32nd' for 32", () => {
+    expect(getOrdinalNumber(32)).toEqual("32nd");
+    });
```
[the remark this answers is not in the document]
SaraTahir28: grouped the tests together.

Sprint-3/3-mandatory-practice/implement/repeat.js
```diff
@@ -1,5 +1,12 @@
-function repeat() {
-    return "hellohellohello";
+function repeat(str, count) {
+  if (count < 0) {
+    return "Invalid Input";
+  }
+  return `${str}`.repeat(count);
```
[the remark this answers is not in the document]
SaraTahir28: Ahh makes sense. fixed it. I was returning it as a string rather than throwing an error.

cjyuan:
Changes look good. Well done.

PR #611, opened 2025-06-26
LONDON| MAY 2025 | HAKAN MURAT KAVUT | Module-Structuring-and-Testing-Data - Sprint 3

hmkavut:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
+++ b/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
@@ -1,6 +1,7 @@
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
-    // add your completed function from key-implement here
+    if (Math.abs(numerator) < Math.abs(denominator)) return true;
+    if (Math.abs(numerator) > Math.abs(denominator)) return true;
+    if (Math.abs(numerator) === Math.abs(denominator)) return true;
 }
 
 module.exports = isProperFraction;
--- a/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
+++ b/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
@@ -5,7 +5,16 @@ test("should return true for a proper fraction", () => {
 });
 
 // Case 2: Identify Improper Fractions:
+test("should return true for a improper fraction", () => {
+  expect(isProperFraction(5, 2)).toEqual(true);
+});
 
 // Case 3: Identify Negative Fractions:
+test("should return true for a negative fraction", () => {
+  expect(isProperFraction(-4, 7)).toEqual(true);
+});
 
 // Case 4: Identify Equal Numerator and Denominator:
+test("should return true for a equal fraction", () => {
+  expect(isProperFraction(3, 3)).toEqual(true);
+});
--- a/Sprint-3/3-mandatory-practice/implement/count.js
+++ b/Sprint-3/3-mandatory-practice/implement/count.js
@@ -1,5 +1,16 @@
 function countChar(stringOfCharacters, findCharacter) {
-    return 5
+    if (!stringOfCharacters.includes(findCharacter)) {
+        return "No occurrence";
+    }
+
+    let count = 0;
+    for (let char of stringOfCharacters) {
+        if (char === findCharacter) {
+            count++;
+        }
+    }
+    return count;
 }
+
 
 module.exports = countChar;
--- a/Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
+++ b/Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
@@ -11,3 +11,21 @@ const getOrdinalNumber = require("./get-ordinal-number");
 test("should return '1st' for 1", () => {
     expect(getOrdinalNumber(1)).toEqual("1st");
     });
+test("should return '2nd' for 2", () => {
+    expect(getOrdinalNumber(2)).toEqual("2nd");
+    });
+test("should return '3rd' for 3", () => {
+    expect(getOrdinalNumber(3)).toEqual("3rd");
+    });
+test("should return 'th' for 11", () => {
+    expect(getOrdinalNumber(11)).toEqual("th");
+    });
+test("should return 'th' for 12", () => {
+    expect(getOrdinalNumber(12)).toEqual("th");
+    });
+test("should return 'th' for 13", () => {
+    expect(getOrdinalNumber(13)).toEqual("th");
+    });
+test("should return 'th' for 4", () => {
+    expect(getOrdinalNumber(4)).toEqual("th");
+    });
--- a/Sprint-3/3-mandatory-practice/implement/repeat.js
+++ b/Sprint-3/3-mandatory-practice/implement/repeat.js
@@ -1,5 +1,15 @@
-function repeat() {
-    return "hellohellohello";
+function repeat(str, count) {
+
+  if (Number(count) > 1) {
+    let result=""
+    for (let i = 0; i <= count-1; i++) {
+      result += str;
+    }
+    return result;
+  }
+  if (count == 1) return str;
+  if (count == 0) return "";
+  return "Negative counts are not valid.";
 }
 
 module.exports = repeat;
--- a/Sprint-3/3-mandatory-practice/implement/repeat.test.js
+++ b/Sprint-3/3-mandatory-practice/implement/repeat.test.js
@@ -10,23 +10,41 @@ const repeat = require("./repeat");
 // Then it should repeat the str count times and return a new string containing the repeated str values.
 
 test("should repeat the string count times", () => {
-    const str = "hello";
-    const count = 3;
-    const repeatedStr = repeat(str, count);
-    expect(repeatedStr).toEqual("hellohellohello");
-    });
+  const str = "hello";
+  const count = 4;
+  const repeatedStr = repeat(str, count);
+  expect(repeatedStr).toEqual("hellohellohellohello");
+});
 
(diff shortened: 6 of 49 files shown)


Sprint-3/3-mandatory-practice/implement/count.js
```diff
@@ -1,5 +1,16 @@
 function countChar(stringOfCharacters, findCharacter) {
-    return 5
+    if (!stringOfCharacters.includes(findCharacter)) {
+        return "No occurrence";
+    }
```
cjyuan, line 4: When a function can return different types of value, it will make consuming the return value inconvenience.

For example, let `str1` and `str2` be two variables storing some string values.
How can we use `countChar()` to check if `str1` contains more space characters than `str2`? 


Can you think of a better value to return when `stringOfCharacters` does not contain any instance of `findCharacter`?

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -11,3 +11,21 @@ const getOrdinalNumber = require("./get-ordinal-number");
 test("should return '1st' for 1", () => {
     expect(getOrdinalNumber(1)).toEqual("1st");
     });
+test("should return '2nd' for 2", () => {
+    expect(getOrdinalNumber(2)).toEqual("2nd");
+    });
```
cjyuan, line 16: To ensure thorough testing, we need broad scenario coverage. Listing individual values, however, can quickly lead to an unmanageable number of test cases.

```
});
```

You may also discover a bug in your implementation when you test your function using more samples.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -11,3 +11,21 @@ const getOrdinalNumber = require("./get-ordinal-number");
...
+    expect(getOrdinalNumber(3)).toEqual("3rd");
+    });
+test("should return 'th' for 11", () => {
+    expect(getOrdinalNumber(11)).toEqual("th");
+    });
+test("should return 'th' for 12", () => {
+    expect(getOrdinalNumber(12)).toEqual("th");
+    });
+test("should return 'th' for 13", () => {
+    expect(getOrdinalNumber(13)).toEqual("th");
```
cjyuan, line 27: We could also specify these tests under the category, "should append 'th' for numbers ending in 11, 12, 13".

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -11,3 +11,21 @@ const getOrdinalNumber = require("./get-ordinal-number");
...
+    });
+test("should return 'th' for 4", () => {
+    expect(getOrdinalNumber(4)).toEqual("th");
```
cjyuan, line 30: We could generalise this test to, "should append 'th' for numbers ending in 0, 4-9".

Sprint-3/3-mandatory-practice/implement/repeat.js
```diff
@@ -1,5 +1,15 @@
-function repeat() {
-    return "hellohellohello";
+function repeat(str, count) {
+
+  if (Number(count) > 1) {
```
cjyuan, line 3: Why do you explicitly convert `count` to a number?  Isn't `count` already a number?

Sprint-3/3-mandatory-practice/implement/repeat.js
```diff
@@ -1,5 +1,15 @@
...
+    let result=""
+    for (let i = 0; i <= count-1; i++) {
+      result += str;
+    }
+    return result;
+  }
+  if (count == 1) return str;
+  if (count == 0) return "";
+  return "Negative counts are not valid.";
```
cjyuan, line 12: How would the caller distinguish the result of the following two function calls?
1. `repeat("Negative counts are not valid.", 1)`
2. `repeat("", -1)`

Sprint-3/3-mandatory-practice/implement/repeat.test.js
```diff
@@ -10,23 +10,41 @@ const repeat = require("./repeat");
...
 // case: Negative Count:
 // Given a target string str and a negative integer count,
 // When the repeat function is called with these inputs,
 // Then it should throw an error or return an appropriate error message, as negative counts are not valid.
+test("should repeat empty string for count equals zero", () => {
+  const str = "hello";
+  const count = -5;
+  const repeatedStr = repeat(str, count);
+  expect(repeatedStr).toEqual("Negative counts are not valid.");
+});
```
cjyuan, line 50: If you modified `repeat()` to throw an error when `count` is negative, and you wanted to test if the function can throw an error as expected, you can use `.toThrow()`. You can find out more about how to use `.toThrow()` here: https://jestjs.io/docs/expect#tothrowerror (Note: Pay close attention to the syntax of the example)

Sprint-3/4-stretch-investigate/password-validator.js
```diff
@@ -1,6 +1,38 @@
...
-}
+    if (password.length < 5) return false;
+
+    let previousPasswords = ["HelloWorld01#", "HelloWorld02$", "HelloWorld03%", "HelloWorld04.", "HelloWorld05*", "HelloWorld06&"];
+    let specials = ["!", "#", "$", "%", ".", "*", "&"];
+    let numbers = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'];
+
+    // Build alphabet arrays manually
+    let smallLetters = [];
+    let upperLetters = [];
```
cjyuan, line 10: String also have `includes()` method which can be used to check if a character is one of the characters in the string.

We could also prepare `numbers`, `smallLetters` as
```
  // Declaring them using const can prevent them from being accidently reassigned a different value.
  const numbers = "0123456789";   
  const smallLetters = "abcdefghijklmnopqrstuvwxyz"; 
```

Sprint-3/4-stretch-investigate/password-validator.test.js
```diff
@@ -15,12 +15,67 @@ To be valid, a password must:
...
+    const password = "HELLOWORLD02#";
+    // Act
+    const result2 = isValidPassword(password);
+    // Assert
+    expect(result2).toEqual(false);
+}
+);
+test("password has not number", () => {
+    // Arrange
+    const password = "HELLOWORLD#";
```
cjyuan, line 48: The function can return `false` for multiple reasons.

For example, the function might return `false` for "HELLOWORLD#" not only because it lacks a digit, but also because it lacks a lowercase letter.
As a result, we can't be certain that the function correctly handles the case described in the test since multiple conditions are being violated simultaneously.

cjyuan:
Can you also include a brief description of this PR in the "Changelist" section?
image

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
```diff
@@ -1,6 +1,7 @@
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
-    // add your completed function from key-implement here
+    if (Math.abs(numerator) < Math.abs(denominator)) return true;
+    if (Math.abs(numerator) > Math.abs(denominator)) return true;
+    if (Math.abs(numerator) === Math.abs(denominator)) return true;
```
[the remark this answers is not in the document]
hmkavut: My function was completely wrong and I modified

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
```diff
@@ -5,7 +5,16 @@ test("should return true for a proper fraction", () => {
...
 
 // Case 3: Identify Negative Fractions:
+test("should return true for a negative fraction", () => {
+  expect(isProperFraction(-4, 7)).toEqual(true);
+});
 
 // Case 4: Identify Equal Numerator and Denominator:
+test("should return true for a equal fraction", () => {
+  expect(isProperFraction(3, 3)).toEqual(true);
+});
```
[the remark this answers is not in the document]
hmkavut: I removed equality at the function and changed expected output as false.

Sprint-3/3-mandatory-practice/implement/repeat.js
hmkavut: explicitly convert is removed at repeat.js

Sprint-3/3-mandatory-practice/implement/repeat.test.js
hmkavut: Test case is updated with using toThrow() function

Sprint-3/4-stretch-investigate/password-validator.test.js
hmkavut: I added new test cases for multiple conditions

hmkavut:
Hi @cjyuan,

I updated my PR based on your comments and I edited  changelist part.

Sprint-3/3-mandatory-practice/implement/count.js
hmkavut: I removed this part so the function always returns  number value

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
hmkavut: I changed test cases format and grouped them.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
hmkavut: I updated test cases

hmkavut:
Thank you @cjyuan

PR #614, opened 2025-06-26
London | May-2025 | Ikenna Agulobi | Structuring and Testing Data sprint-3

ike-agu:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.
This PR includes my completed exercises for Sprint 3 in the Structuring and Testing Data module.

###The exercises included in this PR:###

1- key-implement
2- mandatory-rewrite
3- mandatory-practice
4- stretch-investigate

## Questions

Hello @cjyuan - I wasn't able to complete the test cases for Card validator in Stretch-investigate. Given that it's not among mandatory exercises,  can  I come back to it when I have more time as I'm a bit behind?

### DIFF
--- a/Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
+++ b/Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
@@ -1,7 +1,10 @@
 function getAngleType(angle) {
   if (angle === 90) return "Right angle";
   // replace with your completed function from key-implement
-
+  if (angle < 90) return "Acute angle";
+  if (angle > 90 && angle < 180) return "Obtuse angle";
+  if (angle === 180) return "Straight angle";
+  if (angle > 180 && angle < 360) return "Reflex angle";
 }
 
 
@@ -13,6 +16,6 @@ function getAngleType(angle) {
 
 // Don't get bogged down in this detail
 // Jest uses CommonJS module syntax by default as it's quite old
-// We will upgrade our approach to ES6 modules in the next course module, so for now 
+// We will upgrade our approach to ES6 modules in the next course module, so for now
 // we have just written the CommonJS module.exports syntax for you
 module.exports = getAngleType;
--- a/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
+++ b/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
@@ -10,5 +10,5 @@ function isProperFraction(numerator, denominator) {
       return false;
     }
 }
-console.log(isProperFraction(1, 6));
+// console.log(isProperFraction(-4, 6));
 module.exports = isProperFraction;
--- a/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
+++ b/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
@@ -1,7 +1,6 @@
 function isProperFraction(numerator, denominator) {
     if (numerator < denominator) return true;
     // add your completed function from key-implement here
-    if (numerator < denominator) return true;
     if (numerator >= denominator) return false;
 }
 
--- a/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
+++ b/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
@@ -2,7 +2,7 @@ function isProperFraction(numerator, denominator) {
 
     // add your completed function from key-implement here
     if(denominator === 0){
-      return "Denominator cannot be zero"
+    throw new Error("Denominator cannot be zero")
     }
     if( Math.abs(numerator) < Math.abs(denominator)){
       return true;
@@ -10,5 +10,8 @@ function isProperFraction(numerator, denominator) {
       return false;
     }
 }
+
+console.log(isProperFraction(-3, 0));
+
 // console.log(isProperFraction(-4, 6));
 module.exports = isProperFraction;
--- a/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
+++ b/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
@@ -1,6 +1,8 @@
 function isProperFraction(numerator, denominator) {
     if (numerator < denominator) return true;
     // add your completed function from key-implement here
+    if (numerator < denominator) return true;
+    if (numerator >= denominator) return false;
 }
 
 module.exports = isProperFraction;
--- a/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
+++ b/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
@@ -1,7 +1,14 @@
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
-    // add your completed function from key-implement here
-    if (numerator >= denominator) return false;
-}
 
+    // add your completed function from key-implement here
+    if(denominator === 0){
+      return "Denominator cannot be zero"
+    }
+    if( Math.abs(numerator) < Math.abs(denominator)){
+      return true;
+    }else{
+      return false;
+    }
+}
+console.log(isProperFraction(1, 6));
 module.exports = isProperFraction;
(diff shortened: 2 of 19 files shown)


cjyuan:
Your PR is created against CYF's `coursework/sprint-3` and not against CYF's `main`. Can you change the current PR so that it is compared to CYF's `main`?

ike-agu:
Hi @cjyuan , I've updated my PR to point to CodeYourFuture:main as requested. Please let me know if anything else needs to be changed. Thank you!

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -10,4 +10,23 @@ const getOrdinalNumber = require("./get-ordinal-number");
 
 test("should return '1st' for 1", () => {
     expect(getOrdinalNumber(1)).toEqual("1st");
-    });
+ });
+
+test("Should return '2nd' for 2",()=>{
+  expect(getOrdinalNumber(2)).toEqual("2nd");
+})
```
cjyuan, line 17: To ensure thorough testing, we need broad scenario coverage. Listing individual values, however, can quickly lead to an unmanageable number of test cases.

```
});
```

Sprint-3/3-mandatory-practice/implement/repeat.js
```diff
@@ -1,5 +1,10 @@
-function repeat() {
-    return "hellohellohello";
+function repeat(str, count) {
+  if (count < 0) {
+    return "Negative numbers are not valid!";
+  } else {
```
cjyuan, line 4: How would the caller distinguish the result of the following two function calls?
1. `repeat("Negative numbers are not valid!", 1)`
2. `repeat("", -1)`

Sprint-3/3-mandatory-practice/implement/repeat.test.js
```diff
@@ -21,12 +21,34 @@ test("should repeat the string count times", () => {
...
 // case: Negative Count:
 // Given a target string str and a negative integer count,
 // When the repeat function is called with these inputs,
 // Then it should throw an error or return an appropriate error message, as negative counts are not valid.
+
+test("should return an error message", () => {
+  const str = "hello"
+  const count = -2
+  const errorMessage = repeat(str, count)
+  expect(errorMessage).toEqual("Negative numbers are not valid!");
```
cjyuan, line 53: If you modified `repeat()` to throw an error when `count` is negative, and you wanted to test if the function can throw an error as expected, you can use `.toThrow()`. You can find out more about how to use `.toThrow()` here: https://jestjs.io/docs/expect#tothrowerror (Note: Pay close attention to the syntax of the example)

Sprint-3/4-stretch-investigate/card-validator.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+  }
+  if (repetitionNumber === 16) {
+    return "not valid, all numbers are the same";
+  }
+
+  if (cardNumber[cardNumber.length - 1] % 2 == 1)
+    return "Invalid card: odd final number";
+
+  if (sum < 16) return "Sum less than 16 ";
```
cjyuan, line 28: According to the spec, the sum of all the digits of a valid card number must be greater than 16 (> 16).

The opposite of `> 16` is `<= 16`.

Sprint-3/4-stretch-investigate/password-validator.test.js
```diff
@@ -10,17 +10,49 @@ To be valid, a password must:
...
-    const password = "12345";
+    const password = "Abc1!";
     // Act
     const result = isValidPassword(password);
     // Assert
     expect(result).toEqual(true);
 }
-);
\ No newline at end of file
+);
```
cjyuan, line 26: The check only ensure the function can return `true` when the password is correct; it does not ensure the function can return `false` when a password does not meet the condition specified.

The same applies to the other tests in this script.

Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
```diff
@@ -1,7 +1,10 @@
 function getAngleType(angle) {
   if (angle === 90) return "Right angle";
   // replace with your completed function from key-implement
-
+  if (angle < 90) return "Acute angle";
+  if (angle > 90 && angle < 180) return "Obtuse angle";
+  if (angle === 180) return "Straight angle";
+  if (angle > 180 && angle < 360) return "Reflex angle";
```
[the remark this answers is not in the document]
ike-agu: I've updated the function get angle type to return "Invalid input for angle <=0 or > 360".

Sprint-3/2-mandatory-rewrite/3-get-card-value.js
```diff
@@ -1,5 +1,15 @@
 function getCardValue(card) {
     // replace with your code from key-implement
-    return 11;
+    let rank = card.slice(0, -1); // this line removes the emojis
+
+    if (rank === "A") {
+      return 11;
+    } else if (+rank >= 2 && +rank <= 10) {
+      return +rank;
```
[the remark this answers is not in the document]
ike-agu: I've tested the above function calls, I realised my function does not work as expected.So, I've updated my code to ensure that only the standard card ranks are accepted.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -10,4 +10,23 @@ const getOrdinalNumber = require("./get-ordinal-number");
...
+test("Should return '11th' for 11",()=>{
+  expect(getOrdinalNumber(11)).toEqual("11th");
+})
+test("Should return '12th' for 12",()=>{
+  expect(getOrdinalNumber(12)).toEqual("12th");
+})
+
+test("Should return '13th' for 13",()=>{
+  expect(getOrdinalNumber(13)).toEqual("13th");
+})
```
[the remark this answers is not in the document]
ike-agu: I've grouped my test cases as you suggested. Thank you!

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
ike-agu: Thanks for the feedback. Now I know that to list individual values when writing test cases will lead to a lot of test cases and hard to maintain. Anyway, I've grouped them as you suggested. Thanks!

Sprint-3/3-mandatory-practice/implement/repeat.test.js
ike-agu: I've fixed my code to throw an error when count is negative. Thanks!

Sprint-3/4-stretch-investigate/password-validator.test.js
ike-agu: I've added test cases to check that my function returns false when password is too short, has no special character, has no number, no upper case and has no lower case as you've requested. Thanks!

ike-agu:
Hi @cjyuan , I've implemented all the changes you requested. Could you please take a look. Thank you 🙏

cjyuan (commented):
Changes look good. Well done.

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
```diff
@@ -1,7 +1,14 @@
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
+
     // add your completed function from key-implement here
-    if (numerator >= denominator) return false;
+    if(denominator === 0){
+      return "Denominator cannot be zero"
```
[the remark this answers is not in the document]
ike-agu: I've updated my function to throw an error instead of returning just a string. Thank you for pointing that out

Sprint-3/4-stretch-investigate/password-validator.test.js
```diff
@@ -56,3 +56,34 @@ test("password must not be any previous password in the passwords array", ()=> {
   const result = isValidPassword(password);
   expect(result).toEqual(false);
 });
+
+test("fails when password is less than 5 characters", () => {
+  const password = "A1!";
```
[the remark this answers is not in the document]
ike-agu: I've included a lowercase.

ike-agu:
Hi @cjyuan,
Thank you for taking the time to review my PR and for providing a lot of valuable feedback. I’ve learned things I didn’t know before, especially the importance of grouping test cases and returning proper error messages instead of just strings. I really appreciate your guidance.

Thanks again!

PR #627, opened 2025-06-30
WM | May-2025 | Abdullah Saleh | Sprint-3

3bdullah-saleh:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions

### DIFF
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -10,7 +10,7 @@
 function getAngleType(angle) {
    if (angle === 90) return "Right angle";
    else if (angle === 45) return "Acute angle";
-   else if (angle > 90 && angle < 180) return "obtuse angle";
+   else if (angle > 90 && angle < 180) return "Obtuse angle";
    else if (angle === 180) return "Straight angle";
    else if (angle > 180 && angle < 360) return "Reflex angle";
    // read to the end, complete line 36, then pass your test here
@@ -47,7 +47,7 @@ assertEquals(acute, "Acute angle");
 // When the angle is greater than 90 degrees and less than 180 degrees,
 // Then the function should return "Obtuse angle"
 const obtuse = getAngleType(120);
-assertEquals(obtuse, "obtuse angle");
+assertEquals(obtuse, "Obtuse angle");
 
 // Case 4: Identify Straight Angles:
 // When the angle is exactly 180 degrees,
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -10,9 +10,9 @@
 function getAngleType(angle) {
    if (angle === 90) return "Right angle";
    else if (angle === 45) return "Acute angle";
-   else if (angle > 90 && angle < 180) return "obtuse angle"
-   else if (angle === 180) return "Straight angle"
-   else if (angle > 180 && angle < 360) return "Reflex angle"
+   else if (angle > 90 && angle < 180) return "obtuse angle";
+   else if (angle === 180) return "Straight angle";
+   else if (angle > 180 && angle < 360) return "Reflex angle";
    // read to the end, complete line 36, then pass your test here
 }
 
@@ -61,4 +61,4 @@ assertEquals(straight, "Straight angle");
 // Then the function should return "Reflex angle"
 // ====> write your test here, and then add a line to pass the test in the function above
 const reflex = getAngleType(240);
-assertEquals(reflex, "Reflex angle")
+assertEquals(reflex, "Reflex angle");
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -9,6 +9,10 @@
 
 function getAngleType(angle) {
    if (angle === 90) return "Right angle";
+   else if (angle === 45) return "Acute angle";
+   else if (angle > 90 && angle < 180) return "obtuse angle"
+   else if (angle === 180) return "Straight angle"
+   else if (angle > 180 && angle < 360) return "Reflex angle"
    // read to the end, complete line 36, then pass your test here
 }
 
@@ -43,14 +47,18 @@ assertEquals(acute, "Acute angle");
 // When the angle is greater than 90 degrees and less than 180 degrees,
 // Then the function should return "Obtuse angle"
 const obtuse = getAngleType(120);
-// ====> write your test here, and then add a line to pass the test in the function above
+assertEquals(obtuse, "obtuse angle");
 
 // Case 4: Identify Straight Angles:
 // When the angle is exactly 180 degrees,
 // Then the function should return "Straight angle"
 // ====> write your test here, and then add a line to pass the test in the function above
+const straight = getAngleType(180);
+assertEquals(straight, "Straight angle");
 
 // Case 5: Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
 // ====> write your test here, and then add a line to pass the test in the function above
+const reflex = getAngleType(240);
+assertEquals(reflex, "Reflex angle")
--- a/Sprint-3/1-key-implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-key-implement/2-is-proper-fraction.js
@@ -8,7 +8,7 @@
 // write one test at a time, and make it pass, build your solution up methodically
 
 function isProperFraction(numerator, denominator) {
-    if (denominator === 0) return "Undefined";
+    if (denominator === 0) throw new Error("Denominator cannot be zero");
     else if (numerator < denominator) return true;
     else return false;
 }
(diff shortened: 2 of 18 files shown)


sathudeva7 (commented):
Always validate input types (string/number) when writing functions that expect numbers.

Sprint-3/1-key-implement/1-get-angle-type.js
```diff
@@ -9,6 +9,10 @@
 
 function getAngleType(angle) {
    if (angle === 90) return "Right angle";
+   else if (angle === 45) return "Acute angle";
+   else if (angle > 90 && angle < 180) return "obtuse angle";
```
illicitonion, line 13: I notice this string uses a different casing ("obtuse angle" vs "Obtuse angle") than the other returned values - is there a reason for this? In general, consistency is really useful when coding.

Sprint-3/1-key-implement/2-is-proper-fraction.js
```diff
@@ -8,7 +8,9 @@
 // write one test at a time, and make it pass, build your solution up methodically
 
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
+    if (denominator === 0) return "Undefined";
```
illicitonion, line 11: This is an interesting value to choose - when someone calls this function, I imagine they'll probably try to treat the return value as a boolean, like in an `if` statement:

```js
if (isProperFraction(n, d)) {
  console.log("Proper!");
} else {
  console.log("Improper :(");
}
```

What would happen in the above code if they called this with `d` of `0`? What do you think they would _want_ to happen? How could you change your function to work as you'd expect in that case?

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -25,27 +30,34 @@ function assertEquals(actualOutput, targetOutput) {
...
+// const aceofSpades = getCardValue("A♠");
+// assertEquals(aceofSpades, 11);
 
 // Handle Number Cards (2-10):
 // Given a card with a rank between "2" and "9",
 // When the function is called with such a card,
 // Then it should return the numeric value corresponding to the rank (e.g., "5" should return 5).
 const fiveofHearts = getCardValue("5♥");
-// ====> write your test here, and then add a line to pass the test in the function above
+assertEquals(fiveofHearts, 5);
```
illicitonion, line 41: You handle 10 differently from other numbers - I'd recommend adding a test for 10 as well, to make sure it works.

In general each time you know your code handles some group of input differently, you probably want a test for that example.

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
```diff
@@ -5,7 +5,15 @@ test("should return true for a proper fraction", () => {
 });
 
 // Case 2: Identify Improper Fractions:
-
+test("should return false for a improper fraction", () => {
+  expect(isProperFraction(5, 3)).toEqual(false);
+});
 // Case 3: Identify Negative Fractions:
+test("should return false for a improper fraction", () => {
+  expect(isProperFraction(-5, 10)).toEqual(true);
```
illicitonion, line 13: The name of the function says it should be false, but the test is checking that it's true - these should probably be consistent!

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
```diff
@@ -1,5 +1,11 @@
 function getOrdinalNumber(num) {
- return "1st";
+    let oneDigit = num % 10;
+    let twoDigit = num % 100;
+    if (twoDigit >= 11 && twoDigit <= 13 ) return `${num}th`;
```
illicitonion, line 4: Good identification of an edge-case here - most people doing this exercise notice 12 but forget about 112!

Sprint-3/3-mandatory-practice/implement/repeat.js
```diff
@@ -1,5 +1,8 @@
-function repeat() {
-    return "hellohellohello";
+function repeat(str, count) {
+  if (count == 1) return `${str}`;
```
illicitonion, line 2: Is there a reason you need to handle 1 specially? What would happen if you removed this branch?

illicitonion (commented):
This is generally looking really good, but I left a few comments of things to think about.

I haven't reviewed the card validator yet, I'll take a look later today, but wanted to leave you some comments you can work on in the mean time!

Sprint-3/4-stretch-investigate/card-validator.js
```diff
@@ -0,0 +1,55 @@
+function cardNumberValidator(cardNumber) {
+  if (cardNumber.length !== 16) {
+    return false;
+  }
+  const numberInArray = cardNumber.toString().split(""); // convert the cardNumber to sting then into an array
```
illicitonion, line 5: It's interesting that you check its length and _then_ convert it to a string. What type do you think `cardNumber` may have which has a `.length` property, but isn't a string?

Sprint-3/4-stretch-investigate/card-validator.js
```diff
@@ -0,0 +1,55 @@
...
+  // check if all numbers are numbers
+  let isAllNumbers = true;
+  for (let index = 0; index <= numberInArray.length - 1; index++) {
+    const currentNumber = numberInArray[index];
+    if (isNaN(currentNumber)) {
+      isAllNumbers = false;
+      break;
```
illicitonion, line 12: What's the benefit of setting a boolean and breaking, rather than just writing `return false;` here?

Sprint-3/4-stretch-investigate/card-validator.js
```diff
@@ -0,0 +1,55 @@
...
+    if (numberInArray[i] !== numberInArray[0]) {
+      hasDifferentDigits = true;
+      break;
+    }
+  }
+  if (!hasDifferentDigits) {
+    return false; // all digits are the same
+  }
+  // check if sum of all the digits are greater than 16.
+    let sum = 0;
```
illicitonion, line 30: Why is this code indented more than the line before?

Sprint-3/4-stretch-investigate/card-validator.test.js
```diff
@@ -0,0 +1,32 @@
+// Implement a function cardNumberValidator
```
illicitonion, line 1: This is a good set of tests, but your function could currently just be:

```js
function cardNumberValidator(number) {
  return false;
}
```

And all the tests would pass. Can you work out why? How would you fix this?

Sprint-3/1-key-implement/1-get-angle-type.js
3bdullah-saleh: Thank you for your time and for reviewing my code.
Regarding this point, there’s no specific reason — it was just a typo, and I’ve fixed it.

Sprint-3/1-key-implement/2-is-proper-fraction.js
3bdullah-saleh: Thank you for the great feedback — that really helped me understand the issue more clearly. I updated the function to throw an error when the denominator is 0, and now it consistently returns a boolean (true or false) for valid inputs.

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js
3bdullah-saleh: yes, I fix them and the function was written wrongly and I fix it

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
3bdullah-saleh: Thanks! That feedback means a lot

Sprint-3/3-mandatory-practice/implement/repeat.js
3bdullah-saleh: I was following the test cases instruction so I handle 1 count specially but in fact when I reviewed the code I noticed that I already handle the case using `.repeat` function.

Sprint-3/2-mandatory-rewrite/3-get-card-value.js
```diff
@@ -1,5 +1,8 @@
 function getCardValue(card) {
-    // replace with your code from key-implement
-    return 11;
+    let rank = card.slice(0, -1);
+    if (rank === "A" || rank === "Ace") return 11;
+    else if (rank >= 2 && rank <= 9) return Number(rank);
+    else if (rank === 10 || rank === "J" || rank === "Q" || rank === "K") return 10;
```
illicitonion, line 5: You could have handled 10 in the "If a number" case or the "J/Q/K" case (both of which work). Why did you decide to choose the approach you did? What benefits/drawbacks does each have?

illicitonion (commented):
This is all looking good - I'll mark this as Complete, but I've left a couple of interesting questions to ponder :)

Sprint-3/3-mandatory-practice/implement/repeat.js
```diff
@@ -1,5 +1,7 @@
-function repeat() {
-    return "hellohellohello";
+function repeat(str, count) {
+  if (count > 0) return `${str.repeat(count)}`;
+  else if (count == 0) return "";
+  else if (count < 0) return "negative counts are not valid";
```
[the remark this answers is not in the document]
3bdullah-saleh: You're right — there's no strong reason for treating them differently.
It would be more consistent to throw an error for invalid input like a negative count, just as I did with the zero denominator case.

Sprint-3/2-mandatory-rewrite/3-get-card-value.js
3bdullah-saleh: Actually, I handled it that way because I was following the structure of the test cases. I assumed that 10 could be treated as part of the numeric range and added to the first condition, rather than grouping it with the string values like "J", "Q", or "K".

PR #628, opened 2025-06-30
Glasgow | ITP May -25 | Mirabelle Morah | Module-Structuring-and-Testing-Data | Coursework/sprint-3

mirabellemorah:
## Learners, PR Template

Self checklist

## Changelist

Provided answers to sprint 3

## Questions

Nil

### DIFF
--- a/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
+++ b/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
@@ -1,6 +1,6 @@
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
-    // add your completed function from key-implement here
+  if (denominator === 0) return false;
+  if (numerator < denominator) return true;
+  if (numerator >= denominator) return false;
 }
-
 module.exports = isProperFraction;
--- a/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
+++ b/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
@@ -1,6 +1,5 @@
 function isProperFraction(numerator, denominator) {
   if (denominator === 0) return false;
-  if (numerator < denominator) return true;
-  if (numerator >= denominator) return false;
+  return Math.abs(numerator) < Math.abs(denominator);
 }
 module.exports = isProperFraction;
--- a/Sprint-3/2-mandatory-rewrite/3-get-card-value.js
+++ b/Sprint-3/2-mandatory-rewrite/3-get-card-value.js
@@ -3,8 +3,10 @@ function getCardValue(card) {
   if (rank === "A") return 11;
 
   if (["K", "Q", "J", "10"].includes(rank)) return 10;
-  const num = Number(rank);
-  if (num >= 2 && num <= 9) return num;
+
+  if (["2", "3", "4", "5", "6", "7", "8", "9"].includes(rank)) {
+    return Number(rank);
+  }
 
   throw new Error("Invalid card rank");
 }
--- a/Sprint-3/2-mandatory-rewrite/3-get-card-value.js
+++ b/Sprint-3/2-mandatory-rewrite/3-get-card-value.js
@@ -1,5 +1,12 @@
 function getCardValue(card) {
-    // replace with your code from key-implement
-    return 11;
+  const rank = card.slice(0, -1);
+  if (rank === "A") return 11;
+
+  if (["K", "Q", "J", "10"].includes(rank)) return 10;
+  const num = Number(rank);
+  if (num >= 2 && num <= 9) return num;
+
+  throw new Error("Invalid card rank");
 }
+
 module.exports = getCardValue;
--- a/Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
+++ b/Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
@@ -1,29 +1,46 @@
 const getCardValue = require("./3-get-card-value");
 
 test("should return 11 for Ace of Spades", () => {
-    const aceofSpades = getCardValue("A♠");
-    expect(aceofSpades).toEqual(11);
-    });
+  expect(getCardValue("A♠")).toEqual(11);
+  expect(getCardValue("A♥")).toEqual(11);
+});
 
 // Case 2: Handle Number Cards (2-10):
-test("should return 5 for 5 of Hearts", () => {
-    const fiveofHearts = getCardValue("5♥");
-    expect(fiveofHearts).toEqual(5);
+
+test("should return correct value for number cards (2-10)", () => {
+  for (let i = 2; i <= 10; i++) {
+    expect(getCardValue(`${i}♠`)).toEqual(i);
+    expect(getCardValue(`${i}♥`)).toEqual(i);
+  }
 });
+
+// Alternate long method
+
+// test("should return the value of number cards (2-10)", () => {
+//   expect(getCardValue("2♠")).toEqual(2);
+//   expect(getCardValue("3♠")).toEqual(3);
+//   expect(getCardValue("4♠")).toEqual(4);
+//   expect(getCardValue("5♠")).toEqual(5);
+//   expect(getCardValue("6♠")).toEqual(6);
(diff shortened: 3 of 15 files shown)


Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
```diff
@@ -1,6 +1,6 @@
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
-    // add your completed function from key-implement here
+  if (denominator === 0) return false;
+  if (numerator < denominator) return true;
+  if (numerator >= denominator) return false;
```
cjyuan, line 4: In mathematics, -A/B == A/-B == -(A/B), and -A/-B == A/B for any integers A and B (B ≠ 0).
They represent a proper fraction if A < B and A ≠ 0 and B ≠ 0.

So `isProperFraction(-4, 3)` should return `false` because 4 >= 3.

Consider comparing the absolute value of the numerator and the denominator instead.

---

We can also use `else` at line 13 to handle any number that does not satisfy the condition of the if statement at line 12.

Sprint-3/2-mandatory-rewrite/3-get-card-value.js
```diff
@@ -1,5 +1,12 @@
 function getCardValue(card) {
-    // replace with your code from key-implement
-    return 11;
+  const rank = card.slice(0, -1);
+  if (rank === "A") return 11;
+
+  if (["K", "Q", "J", "10"].includes(rank)) return 10;
+  const num = Number(rank);
+  if (num >= 2 && num <= 9) return num;
+
```
cjyuan, line 8: In JavaScript, strings that represent valid numeric literals in the language can be safely converted to equivalent numbers. For examples, "0x02", "2.1", or "00_02".

```
getCardValue("0x02♠");
getCardValue("2.1♠");
getCardValue("00_02♠");
```

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -6,6 +6,24 @@ test("should return 11 for Ace of Spades", () => {
     });
 
 // Case 2: Handle Number Cards (2-10):
+test("should return 5 for 5 of Hearts", () => {
+    const fiveofHearts = getCardValue("5♥");
+    expect(fiveofHearts).toEqual(5);
+});
+
```
cjyuan, line 13: When preparing tests, we should ensure the tests cover all possible cases. If we specify a test for individual card, we will need about 53 tests to cover all possible cases. Instead, we could consider classifying all possible values into different categories, and then within each category we test some samples.

For example, one possible category for `getCardValue()` is, "should return the value of number cards (2-10)", and we can prepare the test as 

```
test("should return the value of number cards (2-10)", () => {
    expect(getCardValue("2♣︎")).toEqual(2);
    expect(getCardValue("5♠")).toEqual(5);
    expect(getCardValue("10♥")).toEqual(5);
    // Note: We could also use a loop to check all values from 2 to 10.
});
```

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -6,6 +6,24 @@ test("should return 11 for Ace of Spades", () => {
...
 // Case 3: Handle Face Cards (J, Q, K):
+test("should return 10 for Jack of Diamonds", () => {
+    const jackofDiamonds = getCardValue("J♦");
+    expect(jackofDiamonds).toEqual(10);
+});
+
```
cjyuan, line 19: We could generalise this test to "should return 10 for face cards (J, Q, K)" and check all three ranks J, Q, K).

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
```diff
@@ -1,5 +1,15 @@
...
+  if (typeof num !== "number" || !Number.isInteger(num)) {
+    throw new Error("Input must be an integer.");
+  }
+
+  const suffixes = ["th", "st", "nd", "rd"];
+  const remainder = num % 100;
+
+  return (
+    num +
+    (suffixes[(remainder - 20) % 10] || suffixes[remainder] || suffixes[0])
```
cjyuan, line 11: Can you explain how this expression work?

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -9,5 +9,13 @@ const getOrdinalNumber = require("./get-ordinal-number");
...
 
 test("should return '1st' for 1", () => {
-    expect(getOrdinalNumber(1)).toEqual("1st");
-    });
+  expect(getOrdinalNumber(1)).toEqual("1st");
+});
+
+test("should return '2nd' for 2", () => {
+  expect(getOrdinalNumber(2)).toEqual("2nd");
+});
```
cjyuan, line 17: To ensure thorough testing, we need broad scenario coverage. Listing individual values, however, can quickly lead to an unmanageable number of test cases.

```
});
```

Can you update the tests in this script so that they cover all possible valid integers?

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
mirabellemorah: Thank you for explaining, I'm doing my best to learn and understand this. I've added a new test scenario in the test file (/Module-Structuring-and-Testing-Data/Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.test.js) on line 17 to run a new test  for So isProperFraction(-4, 3)

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
mirabellemorah: Oh true, that would be a lot for 53 tests.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
mirabellemorah: Lines 1-3  checks the kind of input passed in the parameter. If input type is not a number or integer (eg if it's a string) then it gives the error message that the input must be an integer.

Line 6: 'Const suffixes' is an array or ordinal suffixes 
Line 7: and 'const remainder' performs a modulus operation (as we learned in previous tests) on whatever input is passed in the 'num' parameter. Then it gets the digit that remains after that division. Eg 123 % 100 = 1.23 -- therefore const remainder = 23

Lines 9-11: performs multiple things: 

suffixes[(remainder - 20) % 10]
Subtracts 20 from remainder, then gets last digit and uses that last digit to assign the ordinal form
Eg: 

Remainder = 23: (23-20) % 10 = [0.3] 3, so const suffix [3] = rd. There it will assign 23 as '23rd'

Then for '|| suffixes[remainder]' if an answer is undefined it simple uses the remainder with the direct array number: like

if remainder = 1: suffixes[1] = "st"
if remainder = 2: suffixes[2] = "nd"  

etc.

The final fallback if both previous parts are undefined it goes back to  || suffixes[0] which adds 'th' to numbers like 11th  etc.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
mirabellemorah: @cjyuan thanks for your corrections early this morning. I've made corrections now please.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
cjyuan: Would the code still work if we change
`   (suffixes[(remainder - 20) % 10] || suffixes[remainder] || suffixes[0])`
to
`   (suffixes[remainder] || suffixes[(remainder - 20) % 10] || suffixes[0])`
(Change the order of the first two conditions)

----
```
> Eg 123 % 100 = 1.23 -- therefore const remainder = 23
  ^^^^^^^^^^^^^^^^^^^
```  
This unusual notation makes it looks like you are treating `%` as a division operator.

You could just write 123 % 100 = 23 (all programmers should recognise what `%` operator does)

cjyuan:
Changes look good. Well done!

PR #633, opened 2025-07-02
Sheffield| May 2025| Mayowa Fadare| Structure testing sprint 3

mayowa0-7:
## Learners, PR Template

Self checklist

## Changelist

Made changes on the sprint 3 unit of the module structuring and testing, using the template provided

## Questions

### DIFF
--- a/Sprint-3/4-stretch-investigate/password-validator.test.js
+++ b/Sprint-3/4-stretch-investigate/password-validator.test.js
@@ -19,11 +19,14 @@ You must breakdown this problem in order to solve it. Find one test case first a
 const passwordValidator= require("./password-validator");
 
   test("Password must be at least 5 characters long.", () => {
-    expect(passwordValidator("Strong1!")).toBe(true);
+    expect(passwordValidator("St1!")).toBe(false);
 });
-  
+
+  test("password with exactly 5 valid characters long", () => {
+  expect(isValidPassword("Stg1!")).toBe(true);
+});
   test("Password must include at least one uppercase letter.", () =>{
-    expect(passwordValidator("strong1")).toBe(false);
+    expect(passwordValidator("strong1!")).toBe(false);
   } )
   
   test("Password must include at least one lowercase letter.", () =>{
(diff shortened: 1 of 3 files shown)


Sprint-3/4-stretch-investigate/password-validator.test.js
```diff
@@ -14,44 +14,31 @@ To be valid, a password must:
...
-
-module.exports = isValidPassword;
-
+ // Assert
+    
+const passwordValidator= require("./password-validator");
+
+  test("Password must be at least 5 characters long.", () => {
+    expect(passwordValidator("Strong1!")).toBe(true);
+});
```
cjyuan, line 23: Should also test if the function can return false when a password contains less than 5 characters.

Sprint-3/4-stretch-investigate/password-validator.test.js
```diff
@@ -14,44 +14,31 @@ To be valid, a password must:
...
+  
+  test("Password must include at least one uppercase letter.", () =>{
+    expect(passwordValidator("strong1")).toBe(false);
+  } )
```
cjyuan, line 27: The function can return `false` for multiple reasons.

For example, the function might return `false` for "strong1" because 
1. it lacks an uppercase letter, OR
2. it lacks a non-alphanumeric character

As a result, when the function return false, we cannot be certain exactly why it returns false.

cjyuan:
Changes look good.

Leaving responses directly in the comment threads makes tracking the discussion easier. You can try the approach in future PRs.

image

Here is a simplified version of best practices ChatGPT suggested for responding to inline comments in a pull request:
> - ✅ Reply to every comment – Let the reviewer know you saw it.
> - ✏️ Make the change if needed – Fix the code if the comment points out a real issue.
> - 🤔 Explain if you don't agree – If you think the code is fine, politely explain why.
> - ✅ Mark as resolved when done – Only mark comments resolved after you fix or respond.
> - 💬 Keep replies short and polite – Be respectful and to the point.
> - ⏱️ Respond soon – Don’t wait too long to reply.
> - 🧪 Test your changes – Make sure your fixes actually work.
> - 📍 Reply directly under the comment – This keeps the conversation easy to follow.

PR #651, opened 2025-07-10
Glasgow | ITP May -25 | Pandi Simatupang | Module-Structuring-and-Testing-Data | coursework/sprint-3 

PandiSimatupang:
## Learners, PR Template

Self checklist

- [x] My changes meet the requirements of this task

## Changelist

Answers and comments to sprint-3

## Questions

### DIFF
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -7,13 +7,28 @@
 // Write the code to pass the test
 // Then, write the next test! :) Go through this process until all the cases are implemented
 
+// function getAngleType(angle) {
+//   if (angle === 90) return "Right angle";
+//   else if (angle === 180) return "Straight angle";
+//   // read to the end, complete line 36, then pass your test here
+//   else if (angle < 90) return "Acute angle";
+//   else if (angle > 90 && angle < 180) return "Obtuse angle";
+//   else return "Reflex angle";
+// }
+
 function getAngleType(angle) {
-  if (angle === 90) return "Right angle";
-  else if (angle === 180) return "Straight angle";
-  // read to the end, complete line 36, then pass your test here
-  else if (angle < 90) return "Acute angle";
-  else if (angle > 90 && angle < 180) return "Obtuse angle";
-  else return "Reflex angle";
+  switch (true) {
+    case angle === 90:
+      return "Right angle";
+    case angle === 180:
+      return "Straight angle";
+    case angle < 90:
+      return "Acute angle";
+    case angle > 90 && angle < 180:
+      return "Obtuse angle";
+    default:
+      return "Reflex angle";
+  }
 }
 
 // we're going to use this helper function to make our assertions easier to read
--- a/Sprint-3/1-key-implement/1-get-angle-type.js
+++ b/Sprint-3/1-key-implement/1-get-angle-type.js
@@ -8,8 +8,12 @@
 // Then, write the next test! :) Go through this process until all the cases are implemented
 
 function getAngleType(angle) {
-   if (angle === 90) return "Right angle";
-   // read to the end, complete line 36, then pass your test here
+  if (angle === 90) return "Right angle";
+  else if (angle === 180) return "Straight angle";
+  // read to the end, complete line 36, then pass your test here
+  else if (angle < 90) return "Acute angle";
+  else if (angle > 90 && angle < 180) return "Obtuse angle";
+  else return "Reflex angle";
 }
 
 // we're going to use this helper function to make our assertions easier to read
@@ -42,15 +46,21 @@ assertEquals(acute, "Acute angle");
 // Case 3: Identify Obtuse Angles:
 // When the angle is greater than 90 degrees and less than 180 degrees,
 // Then the function should return "Obtuse angle"
+
 const obtuse = getAngleType(120);
 // ====> write your test here, and then add a line to pass the test in the function above
+assertEquals(obtuse, "Obtuse angle");
 
 // Case 4: Identify Straight Angles:
 // When the angle is exactly 180 degrees,
 // Then the function should return "Straight angle"
 // ====> write your test here, and then add a line to pass the test in the function above
+const straight = getAngleType(180);
+assertEquals(straight, "Straight angle");
 
 // Case 5: Identify Reflex Angles:
 // When the angle is greater than 180 degrees and less than 360 degrees,
 // Then the function should return "Reflex angle"
 // ====> write your test here, and then add a line to pass the test in the function above
+const reflex = getAngleType(190);
+assertEquals(reflex, "Reflex angle");
--- a/Sprint-3/1-key-implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-key-implement/2-is-proper-fraction.js
@@ -7,11 +7,26 @@
 // complete the rest of the tests and cases
 // write one test at a time, and make it pass, build your solution up methodically
 
+// function isProperFraction(numerator, denominator) {
+//   if (numerator === 0 || denominator === 0) return false;
+//   else if (numerator === denominator) return false;
+//   else if (Math.abs(numerator) < Math.abs(denominator)) return true;
+//   else if (Math.abs(numerator) > Math.abs(denominator)) return false;
+// }
+
 function isProperFraction(numerator, denominator) {
-  if (numerator === 0 || denominator === 0) return false;
-  else if (numerator === denominator) return false;
-  else if (Math.abs(numerator) < Math.abs(denominator)) return true;
-  else if (Math.abs(numerator) > Math.abs(denominator)) return false;
+  switch (true) {
+    case numerator === 0 || denominator === 0:
+      return false;
+    case numerator === denominator:
+      return false;
+    case Math.abs(numerator) < Math.abs(denominator):
+      return true;
+    case Math.abs(numerator) > Math.abs(denominator):
+      return false;
+    default:
+      return false; // Just in case any weird input gets through
+  }
 }
 
 // here's our helper again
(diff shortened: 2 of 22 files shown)


Sprint-3/1-key-implement/1-get-angle-type.js
```diff
@@ -8,8 +8,12 @@
...
 
 function getAngleType(angle) {
-   if (angle === 90) return "Right angle";
-   // read to the end, complete line 36, then pass your test here
+  if (angle === 90) return "Right angle";
+  else if (angle === 180) return "Straight angle";
+  // read to the end, complete line 36, then pass your test here
+  else if (angle < 90) return "Acute angle";
+  else if (angle > 90 && angle < 180) return "Obtuse angle";
+  else return "Reflex angle";
```
YoanHlebarov, line 16: Here you don't need to write an `if elseif` statement. When you `return` short like this. you can simply write `if` statements each line. Another approach is to use [switch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch).

Sprint-3/1-key-implement/2-is-proper-fraction.js
```diff
@@ -8,7 +8,10 @@
 // write one test at a time, and make it pass, build your solution up methodically
 
 function isProperFraction(numerator, denominator) {
-    if (numerator < denominator) return true;
+  if (numerator === 0 || denominator === 0) return false;
+  else if (numerator === denominator) return false;
+  else if (Math.abs(numerator) < Math.abs(denominator)) return true;
+  else if (Math.abs(numerator) > Math.abs(denominator)) return false;
```
YoanHlebarov, line 14: Same as above.

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -8,7 +8,15 @@
...
+  let rank = card.slice(0, -1); //start from 0 until just before one last character
+  if (rank === "A") return 11;
+  else if ((rank === "10") | (rank === "J") | (rank === "Q") | (rank === "K"))
+    return 10;
+  else if (rank > 1 && rank < 10)
+    return Number(
+      rank
+    ); //need to convert to number, because === compare type and exact number
+  else return null;
 }
```
YoanHlebarov, line 20: Same as above.

Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
```diff
@@ -1,18 +1,15 @@
 function getAngleType(angle) {
   if (angle === 90) return "Right angle";
   // replace with your completed function from key-implement
-
+  else if (angle === 180) return "Straight angle";
+  // read to the end, complete line 36, then pass your test here
+  else if (angle < 90) return "Acute angle";
+  else if (angle > 90 && angle < 180) return "Obtuse angle";
+  else return "Reflex angle";
 }
```
YoanHlebarov, line 9: Same as above.

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
```diff
@@ -1,6 +1,11 @@
...
-    if (numerator < denominator) return true;
-    // add your completed function from key-implement here
+  //if (numerator < denominator) return true;
+  // add your completed function from key-implement here
+
+  if (numerator === 0 || denominator === 0) return false;
+  else if (numerator === denominator) return false;
+  else if (Math.abs(numerator) < Math.abs(denominator)) return true;
+  else if (Math.abs(numerator) > Math.abs(denominator)) return false;
 }
```
YoanHlebarov, line 9: Same as above.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
```diff
@@ -1,5 +1,18 @@
...
+    if (myNum === "1") return `${num}st`;
+    else if (myNum === "2") return `${num}nd`;
+    else if (myNum === "3") return `${num}rd`;
+    else return `${num}th`;
+  } else if (myNum >= 10) {
+    if (myNumLastChar === "1") return `${num}st`;
+    else if (myNumLastChar === "2") return `${num}nd`;
+    else if (myNumLastChar === "3") return `${num}rd`;
+    else return `${num}th`;
+  }
```
YoanHlebarov, line 15: Same as above.

Sprint-3/3-mandatory-practice/implement/repeat.js
```diff
@@ -1,5 +1,10 @@
-function repeat() {
-    return "hellohellohello";
+function repeat(str, rep) {
+  myStr = "";
+  for (let i = 0; i < rep; i++) {
+    myStr = myStr + str;
+  }
+  if (rep < 0) return "negative counts are not valid";
+  return myStr;
```
YoanHlebarov, line 7:
```suggestion
function repeat(str, rep) {
  if (rep < 0) return "negative counts are not valid";
  myStr = "";
  for (let i = 0; i < rep; i++) {
    myStr = myStr + str;
  }
  return myStr;
```

You should always be validating your inputs first, that way there are no dubious values being passed down your program which can create a problem. In this case you should check first that the count is a positive integer, before running any other part of your function.

Sprint-3/4-stretch-investigate/password-validator.js
```diff
@@ -1,6 +1,33 @@
...
+  const symbols = new Array("!", "#", "$", "%", ".", "*", "&");
+
+  const hasSymbol = arr.some((char) => symbols.includes(char));
+  const hasDigit = arr.some((char) => /[0-9]/.test(char));
+  const hasLower = arr.some((char) => /[a-z]/.test(char));
+  const hasUpper = arr.some((char) => /[A-Z]/.test(char));
 
+  const usedPass = !usedPasswords.includes(password);
+  return hasSymbol && hasDigit && hasLower && hasUpper && usedPass;
+}
```
YoanHlebarov, line 31: I would consider moving the check if password is used near the top of the function. Similar to the comment above, to check if it is in the `usedPasswords` list is simple and should be done first, as if it is then the password cannot be used. In your implementation we are doing all checks and then deciding whether the password can be used or not. Consider if using similar approach to other exercises form the module can be used here (eg return short if a check fails).

YoanHlebarov (commented):
The code is correct but it can be improved in terms of performance (doing many checks before returning instead of returning short whenever a check fails) and readability. Think how you can improve the code, is there a way you can write things that are easier to read?

Sprint-3/1-key-implement/1-get-angle-type.js
PandiSimatupang: Ok, I've changed it to switch-case instead of if-else method.

Sprint-3/1-key-implement/2-is-proper-fraction.js
PandiSimatupang: Changed to Switch-case..

Sprint-3/1-key-implement/3-get-card-value.js
PandiSimatupang: modified to switch-case..

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
PandiSimatupang: modify to multiple ifs.. instead of if-else.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
PandiSimatupang: changed to Switch-case method.

Sprint-3/3-mandatory-practice/implement/repeat.js
PandiSimatupang: thanks, I moved the filter if(rep<0) to the first line of the function to skip out fast when it doesn't match.

Sprint-3/4-stretch-investigate/password-validator.js
PandiSimatupang: Thanks, Yes, you're right. It checked all the conditions when it didn't need to. I cleaned it up using short-if statements to make it clearer (at least for me).

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
YoanHlebarov, line 1: That is really good, it is my preferred style of writing such a function, as subjective is it may be. It is easy to read and understand at a glance at what is happening, and each line represents a condition. So if you are reading the code trying to work out what will happen if a specific value is passed into the function, it is really easy to read each condition and determine the outcome, all in a matter of seconds and without having to run the code. Good job here, I like this!

Sprint-3/4-stretch-investigate/password-validator.js
```diff
@@ -1,6 +1,73 @@
...
+//     ? false
+//     : !arr.some((char) => /[0-9]/.test(char))
+//     ? false
+//     : !arr.some((char) => /[a-z]/.test(char))
+//     ? false
+//     : !arr.some((char) => /[A-Z]/.test(char))
+//     ? false
+//     : usedPasswords.includes(password)
+//     ? false
+//     : true;
```
YoanHlebarov, line 51: haha that is rather funny. Yeah... don't write code like this. Sure it works but honestly it's a mess to read. LLMs can also sometimes write code like this, too, so just be smart about it, as you have been in this case!

YoanHlebarov (commented):
These changes are an improvement, well done! I will leave it up to you to explore the switch fall-through but I think it is a good challenge to figure out.

Using `switch` may also not be the best option all the time. In some of the files you have `if` statements which work just as well. In my experience I tend to write `switch` statements rather rarely, so I stick with `if` and return short for each condition. It is one of those things that you get a preference over time and practice, and also influence by your colleagues and the codebase you are contributing to. 

Overall though, this is an improvement so well done!

PR #657, opened 2025-07-13
London | May 2025 | Houssam Lahlah | Coursework/sprint 1

HoussamLh:
## Self Checklist

 I have committed my files one by one, on purpose, and for a reason

 I have titled my PR with REGION | COHORT_NAME | FIRST_NAME LASTNAME | PROJ_NAME

 I have tested my changes (where applicable)

 My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/)

 My changes meet the requirements of this task

## Changelist

1. key-exercise: 
Practiced variable reassignment with count = count + 1
Extracted initials from full names using charAt(0)
Parsed file paths using slice() and lastIndexOf()
Explained how random numbers are generated with Math.random() and Math.floor()

2. mandatory-errors: Identified common learning mistakes (e.g., using const when let is needed)

3. mandatory-interpret: Interpreted JavaScript behaviors like console.log, prompt(), and variable hoisting

4. stretch-explore: Explored the Chrome DevTools Console and JavaScript built-in methods like alert(), prompt(), and typeof

## Questions

Hello,

I’ve completed all parts of the Sprint 1 and opened this pull request for review. Could you please take a look and let me know:

Are my explanations clear and technically correct, especially in the key-exercise and mandatory-interpret sections?

Do the solutions meet the expectations for this stage of the course?
Is there anything I could improve in the way I write comments or explain code logic?
Any feedback would be appreciated — thank you for your time!

Best regards,
Sami

### DIFF
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -4,4 +4,4 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
-//
+// line 3 is increases the value of count by 1 and saves the updated value back into count
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -4,4 +4,4 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
-// line 3 is increases the value of count by 1 and saves the updated value back into count
+// line 3 is increases the value of count by 1 and saves the updated value back into count variable.
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -4,3 +4,4 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+//
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -3,7 +3,74 @@ const maximum = 100;
 
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 
-// In this exercise, you will need to work out what num represents?
-// Try breaking down the expression and using documentation to explain what it means
-// It will help to think about the order in which expressions are evaluated
-// Try logging the value of num and running the program several times to build an idea of what the program is doing
+// 1- In this exercise, you will need to work out what num represents?
+
+// the num gives a random whole number between 1 and 100 like 73, 12, or 100.
+
+// 2- Try breaking down the expression and using documentation to explain what it means
+/*
+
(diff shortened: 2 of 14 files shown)


Sprint-1/1-key-exercises/1-count.js
```diff
@@ -4,3 +4,4 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+// line 3 is increases the value of count by 1 and saves the updated value back into count variable.
```
cjyuan, line 7: Operation like `count = count + 1` is very common in programming, and there is a programming term describing such operation.

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -3,7 +3,74 @@ const maximum = 100;
...
+// 1- In this exercise, you will need to work out what num represents?
+
+// the num gives a random whole number between 1 and 100 like 73, 12, or 100.
+
+// 2- Try breaking down the expression and using documentation to explain what it means
+/*
+
+1. Math.random()
+
+Returns a random decimal number between 0 and 1 but never gives 1.0.
```
cjyuan, line 15: We can also use the concise and precise ***interval notation*** to describe a range of values. 
- `[`, `]` => inclusion
- `(`, `)` => exclusion

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -1,27 +1,31 @@
...
 );
+// 4. Extracts the pounds part by taking all characters except the last two.
+//    For "399", this is "3" (the hundreds place).
 
 const pence = paddedPenceNumberString
   .substring(paddedPenceNumberString.length - 2)
   .padEnd(2, "0");
+// 5. Extracts the last two characters as the pence part.
+//    For "399", this is "99".
+//    Then pads the string on the right with zeros if it’s shorter than 2 characters (not needed here).
```
cjyuan, line 27: When you mentioned "not needed here", do you mean we can delete `.padEnd(2, "0")` at line 24? 

cjyuan (commented):
Your explanation is clear. I only have some suggestions and a questions.

Sprint-1/1-key-exercises/4-random.js
HoussamLh: Thanks for the explanation! 
That helps make the range much clearer. 
I’ll keep this notation in mind.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
HoussamLh: heloo, 
Yes, I meant that .padEnd(2, "0") isn’t needed in this case. If we remove it, the program should still work correctly for any valid penceString, as long as it’s properly formatted. 
Thanks for checking!

cjyuan: Do you know why it is safe to delete `.padEnd(2, "0")` from this script?

cjyuan:
Thanks for responding to each of my comments.

In the future (within CYF courses), I think it might be a good idea to change the label to "Needs label" to clearly signal that your responses/changes are ready to be reviewed.

PR #677, opened 2025-07-21
WM | 25-ITP-May | Nahom Mesfin | Sprint 3 coursework

nahommesfin77:
## Learners, PR Template

Self checklist

## Changelist

In Implement, I wrote functions step-by-step with assertions to cover different cases. In Rewrite, I converted previous assertions to use Jest syntax for clearer and easier testing. In Practice I built and ran tests using Jest and vscode's code runner to ensure all functions meet requirements.


Sprint-3/3-mandatory-practice/implement/count.js
```diff
@@ -1,5 +1,11 @@
 function countChar(stringOfCharacters, findCharacter) {
-    return 5
+    let count = 0;
```
LonMcGregor, line 2: This is very minor - try to keep the indentation the same for all the code. here your first declaration is indented further.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
```diff
@@ -1,5 +1,8 @@
 function getOrdinalNumber(num) {
```
LonMcGregor, line 1: In english we would say "21st" - can you make your code work in these general cases too?

Sprint-3/3-mandatory-practice/implement/repeat.js
```diff
@@ -1,5 +1,5 @@
-function repeat() {
-    return "hellohellohello";
+function repeat(str, count) {
+  return str.repeat(count);
```
LonMcGregor, line 2: What about your error condition you are testing for?

Sprint-3/1-key-implement/3-get-card-value.js
```diff
@@ -8,7 +8,11 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
+  const rank = card.slice(0, -1); // all characters except last
     if (rank === "A") return 11;
+    if (rank >= '2' && rank <= '9') return Number(rank);
```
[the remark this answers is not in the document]
nahommesfin77: Yes, it should have thrown an error, but because of string comparison of "222" and "2" it will pass the condition and logs 222. I have fixed this issue by using regular expression.

- Updated line 13
`if (/^[2-9]$/.test(rank)) return Number(rank);`

Sprint-3/3-mandatory-practice/implement/count.js
nahommesfin77: Ok, I will. thanks for noticing, I have adjusted the identation.

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.js
nahommesfin77: I have updated the function to include cases like 21st, 22nd and so on. I have used a modulo operator to get the last digits of the numbers `num % 10;` and `num % 100;` and used them to determine the cases for each number.
- Updated line 2 to 10

LonMcGregor:
Great work making these changes - you are done with this sprint now

PR #692, opened 2025-08-04
West Midlands | MAY | EMIN AKTURK | Sprint 3 | MODULE-STRUCTURING-AND-TESTING-DATA

eminakturk:
## Learners, PR Template

Self checklist

## Changelist

Briefly explain your PR.

## Questions


jenny-alexander (changes requested):
You have done a lot of work in this PR! Good job. 

In order for me to set the status to complete, there are a few fixes to make (please see comments left throughout the PR). 

In addition to the comments that I left:
- I noticed that a number of tests are failing. It's really important to run the tests yourself and make sure they pass before you submit your PR for review. When you submit a PR with failing tests, it means that your code implementation is 'broken'.

- The files within Sprint-1 are set to be deleted. This will delete the files in the main repository (and for everyone 😱). I think you did this because you made changes to the file my mistake, is that correct? Since it was a mistake, can you revert these changes instead of deleting the files? 
One way to do this is to revert each file to the commit SHA before your deletion commit SHA. Let me know if you can figure out how to do this or if you would like some additional explanation.

jenny-alexander:
@eminakturk - I noticed that your checklist is not quite right.
You have this: [x ] 
😃 Instead, do this: [x]  
(notice how I don't have a space around the x?)
It will format it to a checked off item.

eminakturk:
Hello Jenny, Thank you for your great insights. I have completed fixing all the errors you pointed out. I hope and believe I did an okay job. Please let me know In case if I need to change anything else 😄

jenny-alexander (changes requested):
The implementation you made in 1-get-angle-types.js (lines 11-17) is good. However, we could simplify this even more.
If none of the criteria are met from lines 19-25, then the angle is considered invalid.

So after line 25, we could add a return statement like:
return "Invalid angle: Please provide a valid number";

__________

The files you deleted from sprint 1 directory should be reverted. We don't want to delete repo files. 

Image

eminakturk:
I have uploaded a fix for that and sent you a message from slack, thank you :)

jenny-alexander (changes requested):
In this case, if you push a PR with files that are deleted, it will delete the branches from the main repository.
It's better to revert this deletion completely -> and also good practice for you once you start working at a company and this same thing happens.

Can you research how to revert changes in a PR? Specifically "how to revert accidental deletion of files"?

PR #699, opened 2025-09-07
London | 25-ITP-May | Houssam Lahlah | Sprint 2 | Acoursework/sprint 3

HoussamLh:
**_Self checklist_**

**_Changelist_**

This pull request for syn my fork and updating my branch.

**_Questions_**

Please, Let me know if there is anything else went wrong?

### DIFF
--- a/Sprint-3/2-mandatory-rewrite/3-get-card-value.js
+++ b/Sprint-3/2-mandatory-rewrite/3-get-card-value.js
@@ -1,5 +1,15 @@
 function getCardValue(card) {
-    // replace with your code from key-implement
-    return 11;
+  const rank = card.slice(0, -1); // All characters except the last one (suit)
+
+  if (!rank) throw new Error("Invalid card rank");
+
+  if (rank === "A") return 11;
+  if (["K", "Q", "J", "10"].includes(rank)) return 10;
+
+  const number = Number(rank);
+  if (number >= 2 && number <= 9) return number;
+
+  throw new Error("Invalid card rank");
 }
+
 module.exports = getCardValue;
--- a/Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
+++ b/Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
@@ -1,11 +1,33 @@
 const getCardValue = require("./3-get-card-value");
 
 test("should return 11 for Ace of Spades", () => {
-    const aceofSpades = getCardValue("A♠");
-    expect(aceofSpades).toEqual(11);
-    });
+  expect(getCardValue("A♠")).toEqual(11);
+});
 
-// Case 2: Handle Number Cards (2-10):
-// Case 3: Handle Face Cards (J, Q, K):
-// Case 4: Handle Ace (A):
-// Case 5: Handle Invalid Cards:
+// Case 2: Handle Number Cards (2-10)
+test("should return 5 for Five of Hearts", () => {
+  expect(getCardValue("5♥")).toEqual(5);
+});
+
+test("should return 10 for Ten of Diamonds", () => {
+  expect(getCardValue("10♦")).toEqual(10);
+});
+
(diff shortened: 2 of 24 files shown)


Sprint-3/2-mandatory-rewrite/3-get-card-value.js
```diff
@@ -1,5 +1,15 @@
...
-    return 11;
+  const rank = card.slice(0, -1); // All characters except the last one (suit)
+
+  if (!rank) throw new Error("Invalid card rank");
+
+  if (rank === "A") return 11;
+  if (["K", "Q", "J", "10"].includes(rank)) return 10;
+
+  const number = Number(rank);
+  if (number >= 2 && number <= 9) return number;
```
cjyuan, line 10: Does your function return the value you expected from each of the following function calls?

```
getCardValue("0x02♠");
getCardValue("2.1♠");
getCardValue("0002♠");
```

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -1,11 +1,33 @@
...
+});
+
+// Case 2: Handle Number Cards (2-10)
+test("should return 5 for Five of Hearts", () => {
+  expect(getCardValue("5♥")).toEqual(5);
+});
+
+test("should return 10 for Ten of Diamonds", () => {
+  expect(getCardValue("10♦")).toEqual(10);
+});
```
cjyuan, line 14: When preparing tests, we should ensure the tests cover all possible cases. If we specify a test for individual card, we will need about 53 tests to cover all possible cases. Instead, we could consider classifying all possible values into different categories, and then within each category we test some samples.

For example, one possible category for `getCardValue()` is, "should return the value of number cards (2-10)", and we can prepare the test as 

```
test("should return the value of number cards (2-10)", () => {
    expect(getCardValue("2♣︎")).toEqual(2);
    expect(getCardValue("5♠")).toEqual(5);
    expect(getCardValue("10♥")).toEqual(10);
    // Note: We could also use a loop to check all values from 2 to 10.
});
```

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -1,11 +1,33 @@
...
+  expect(getCardValue("J♣")).toEqual(10);
+});
+
+test("should return 10 for Queen of Spades", () => {
+  expect(getCardValue("Q♠")).toEqual(10);
+});
+
+test("should return 10 for King of Hearts", () => {
+  expect(getCardValue("K♥")).toEqual(10);
+});
```
cjyuan, line 27: We could generalise this test to "should return 10 for face cards (J, Q, K)" and check all three ranks J, Q, K).

Sprint-3/3-mandatory-practice/implement/get-ordinal-number.test.js
```diff
@@ -11,3 +11,20 @@ const getOrdinalNumber = require("./get-ordinal-number");
...
+  expect(getOrdinalNumber(3)).toEqual("3rd");
+});
+
+test("should return '4th' for 4", () => {
+  expect(getOrdinalNumber(4)).toEqual("4th");
+});
+
+test("should return '11th' for 11", () => {
+  expect(getOrdinalNumber(11)).toEqual("11th");
+});
```
cjyuan, line 29: These tests are not comprehensive; they do not quite cover all possible cases.

To ensure thorough testing, we need broad scenario coverage. Listing individual values, however, can quickly lead to an unmanageable number of test cases.

```
});
```

Sprint-3/3-mandatory-practice/implement/repeat.test.js
cjyuan, line 1: Because of a bug, I don't think this script can be executed successfully. Can you try executing this file and fix the error?

In addition, it is a good practice to delete unused code when submitting the code for review.

PR #702, opened 2025-09-09
London | 25-ITP-May | Houssam Lahlah | Sprint 3 | Coursework/sprint-3

HoussamLh:
**_Self checklist_**

**_Changelist_**

Understanding how to validate ranks strictly (not accepting "0002" or "0x02" as valid numbers).

Deciding whether to use explicit checks vs regex for number cards.

Finding the balance between writing enough tests vs not writing 52 individual card tests.

Making tests more generalised (by category) instead of repetitive.

**_Questions_**

- Is my regex approach (/^[2-9]$/) the right level of strictness for checking number cards, or would you prefer a different method?

- For the tests, is it better to keep them category-based with a few examples (as I’ve done), or would it be valuable to also include a loop over all possible ranks (2–10, J, Q, K, A)?

- Should I also test different suits systematically (♠, ♥, ♦, ♣), or is one suit per category enough since the suit doesn’t affect value?

- Is the error handling message "Invalid card rank" clear enough, or would you recommend more specific errors (e.g., "Unsupported card format" vs "Invalid rank")?

- Would using a map/dictionary (e.g., {A:11, J:10, Q:10, K:10}) for ranks be cleaner than multiple if statements, or is the current approach acceptable for this task?

### DIFF
--- a/Sprint-3/3-stretch/password-validator.js
+++ b/Sprint-3/3-stretch/password-validator.js
@@ -17,9 +17,7 @@ function isValidCreditCard(cardNumber) {
 
   // Calculate the sum of all digits and check if greater than 16
   const sum = cardNumber.split('').reduce((acc, digit) => acc + Number(digit), 0);
-  if (sum <= 16) {
-    return false;
-  }
+    return sum <= 16;
 
   // All conditions passed, card is valid
   return true;


Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
```diff
@@ -8,11 +8,19 @@
 // Then, write the next test! :) Go through this process until all the cases are implemented
 
 function getAngleType(angle) {
+<<<<<<< HEAD:Sprint-3/1-key-implement/1-get-angle-type.js
```
LonMcGregor, line 11: Here, where the file contains lines marked between <<<<<< HEAD and >>>>>>> COMMIT_HASH there is a merge conflict. Do you know what this means? Can you spot the problem that will happen if you commit files containing conflicts?

Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
```diff
@@ -0,0 +1,23 @@
...
+  if (angle === 90) return "Right angle";
+  if (angle < 90) return "Acute angle";
+  if (angle > 90 && angle < 180) return "Obtuse angle";
+  if (angle === 180) return "Straight angle";
+  if (angle > 180 && angle < 360) return "Reflex angle";
+  throw new Error("Invalid angle");
+}
+
+
+
```
LonMcGregor, line 11: Did you mean to commit all these blank lines?

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
```diff
@@ -0,0 +1,15 @@
...
+// It checks if the absolute value of the numerator 
+// is less than the absolute value of the denominator.
+
+
+function isProperFraction(numerator, denominator) {
+  if (denominator === 0) {
+    throw new Error("Denominator cannot be zero");
+  }
+
+  return Math.abs(numerator) < Math.abs(denominator);
```
LonMcGregor, line 11: Does this correctly handle all cases? What about negative improper fractions?

LonMcGregor (commented):
Good start on this sprint's tasks, I have spotted a few areas where you could improve code further

A few of your files contain merge conflicts. I've left a comment on the first one for you to look at. I can't review those files further until those issues are resolved.

Great questions! I've left comments at the relevant places within your code.

Sprint-3/2-mandatory-rewrite/1-get-angle-type.js
HoussamLh: hello **_LonMcGregor_**,

**_Thank you_** for your **_feedback_** and I'm sorry about the **_blank lines_**. I should not commit file with all this blank lines.

Sprint-3/2-mandatory-rewrite/3-get-card-value.test.js
```diff
@@ -0,0 +1,36 @@
...
+});
+
+// Case 2: Number cards (2–10)
+test("should return the value of number cards (2–10)", () => {
+  expect(getCardValue("2♣")).toEqual(2);
+  expect(getCardValue("5♠")).toEqual(5);
+  expect(getCardValue("10♥")).toEqual(10);
+
+  // You can even loop for efficiency:
+  for (let i = 2; i <= 10; i++) {
```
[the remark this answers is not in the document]
HoussamLh: Hello **_LonMcGregor_**,

**_Thank you_** for the feedback! 

- I see now that looping over all number cards isn’t necessarily the most efficient approach in testing. Efficiency in this context means **_covering a representative sample of normal data and edge cases_**, so we get confidence that the function works correctly without making tests unnecessarily long or harder to debug.

- I’ve revised the tests to focus on representative normal values (2, 5, 10) rather than looping over all number cards. This way we still test important cases while keeping the tests clear, easy to read, and maintainable. Edge cases like Ace, face cards, and invalid cards are still covered.

Sprint-3/2-mandatory-rewrite/3-get-card-value.js
```diff
@@ -0,0 +1,18 @@
...
+  const rank = card.slice(0, -1); // remove suit (last char)
+
+  if (!rank) throw new Error("Invalid card rank");
+
+  if (rank === "A") return 11;
+  if (["K", "Q", "J"].includes(rank)) return 10;
+  if (rank === "10") return 10;
+
+  // Explicit check for number cards (2–9 only, no leading zeros or decimals)
+  if (/^[2-9]$/.test(rank)) {
```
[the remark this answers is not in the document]
HoussamLh: Hello **_LonMcGregor_**,

**_Thank you_** for the feedback!

I updated the regex to **^(?:[2-9]|10)$** so it now matches all number cards from **2–10** in a single condition. This removes the need for a separate check for "10" while still correctly handling Ace, face cards, and invalid cards.

Sprint-3/2-mandatory-rewrite/3-get-card-value.js
```diff
@@ -0,0 +1,18 @@
+function getCardValue(card) {
+  const rank = card.slice(0, -1); // remove suit (last char)
+
+  if (!rank) throw new Error("Invalid card rank");
+
+  if (rank === "A") return 11;
```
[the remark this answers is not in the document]
HoussamLh: Hello **_LonMcGregor_**,

**_Thanks_**! I agree — for a small set of card ranks, the current approach is simple and readable. Using a map would work, but for this task it would be unnecessarily complicated. I see that for a larger set of unique ranks, a map would be a better choice.

Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
LonMcGregor: Thanks for fixing this

Sprint-3/2-mandatory-rewrite/2-is-proper-fraction.js
LonMcGregor: Good answer!

LonMcGregor:
Great work on this sprint, you are done with these tasks now

PR #714, opened 2025-09-26
NW | 25-ITP-Sep | TzeMing Ho | Sprint 3 | coursework/sprint-3-stretch

TzeMingHo:
## Learners, PR Template

Self checklist

## Changelist

In the stretch exercise, I have answered questions in find.js, created tests and a function for password validation, and created tests and a function for credit card.


Sprint-3/3-stretch/creditCard-validator.js
```diff
@@ -0,0 +1,37 @@
+function creditCardValidator(cardNumber) {
+  // Remove all - and spaces from the input
+  const sanitized = cardNumber.replace(/[-\s]/g, "");
+
+  //check if the length of the sanitized input is 16
+  if (sanitized.length !== 16) {
+    return false;
+  }
+  // check if the sanitized input contains only digits
+  if (!/^\d{16}$/.test(sanitized)) {
```
illicitonion, line 10: By putting a `{16}` here, you're checking the length again. This means we're checking the length twice.

This is fine, but if we wanted to change the expected length of a credit card, we'd need to change both.

Can you think how to avoid this duplication?

Sprint-3/3-stretch/creditCard-validator.test.js
```diff
@@ -0,0 +1,29 @@
+const creditCardValidator = require("./creditCard-validator");
+
+describe("creditCardValidator", () => {
+  test("should return true for a 16 digit long card number", () => {
+    expect(creditCardValidator("1234-5678-9012-3456")).toBe(true);
+    expect(creditCardValidator("1234 5678 9012 3456")).toBe(true);
```
illicitonion, line 6: Your tests all currently assume there are extra characters in the credit card number - is `123456789012345` valid? You may want to test that kind of thing too, to make sure your implementation doesn't _assume_ there are separators.

Sprint-3/3-stretch/creditCard-validator.test.js
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  test("should return false if the card number is not digit only", () => {
+    expect(creditCardValidator("1234-5678-9012-345a")).toBe(false);
```
illicitonion, line 10: Your only example here of an invalid number has a letter only in the last character. But you have _other_ checks that only check the last character as well. It's possible this is only returning false because you're just checking that the last character is 0/2/4/6/8. Can you think how you could add more tests to make sure this isn't the case?

illicitonion (commented):
This is looking really good - I left a few things to think about!

Sprint-3/3-stretch/creditCard-validator.test.js
TzeMingHo: You are right. I should have included tests for card numbers that are more or fewer than 16 digits. As you suggested, I created one test for the case of fewer and another for more.

Sprint-3/3-stretch/find.js
```diff
@@ -20,6 +20,19 @@ console.log(find("code your future", "z"));
 // Pay particular attention to the following:
 
 // a) How the index variable updates during the call to find
+//    The index variable is set to 0 on line 2.
+//    When the function is called, the while loop on line 4 checks if index (0) is less than the length of the string (16).
+//    If the character is not found, the index will increment by 1 on line 8
+//    Now, index is 1, and the while loop will check again, until the character is found or the index is no longer less than the length of the string.
```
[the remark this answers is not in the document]
TzeMingHo: Thank you. :)

Sprint-3/3-stretch/password-validator.js
```diff
@@ -1,6 +1,41 @@
...
+    uppercaseCondition &&
+    lowercaseCondition &&
+    numberCondition &&
+    symbolCondition &&
+    notInPreviousPasswords
+  ) {
+    return true;
+  } else {
+    return false;
+  }
```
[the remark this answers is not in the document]
TzeMingHo: Yeah. I should have simplified the return statement to return ( ... all the conditions), which looks cleaner.

Sprint-3/3-stretch/password-validator.test.js
```diff
@@ -15,12 +15,31 @@ To be valid, a password must:
...
-    // Act
-    const result = isValidPassword(password);
-    // Assert
-    expect(result).toEqual(true);
-}
-);
\ No newline at end of file
+
+describe("passwordValidator", () => {
+  const password = "123Ab*";
```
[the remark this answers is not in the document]
TzeMingHo: Yeah. validPassword fits the purpose of this variable. I might need to declare other passwords for false cases.

illicitonion (commented):
This is looking great! I have yet more things to discuss, though, but hopefully they're all interesting!

PR #745, opened 2025-10-10
Glasgow | 25-ITP-SEP | Mohammed Abdoon | Sprint 3 | Coursework/1-implement-and-rewrite-tests

M-Abdoon:
## Learners, PR Template

Self checklist

## Changelist

I implemented and rewrote tests, figured the errors and tested the code. modified the files to meet the test requirements.

## Questions

No questions for now, thank you.

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
@@ -24,9 +24,6 @@ test("should return 10 for face cards J Q K", () => {
 // Case 4: Handle Ace (A):
 test("should return 11 when entering Ace A", () => {
   expect(getCardValue("A♠")).toEqual(11);
-  expect(getCardValue("A♥")).toEqual(11);
-  expect(getCardValue("A♦")).toEqual(11);
-  expect(getCardValue("A♣")).toEqual(11);
 });
 
 // Case 5: Handle Invalid Cards:
@@ -35,5 +32,4 @@ test(`should return string ("Invalid card rank.") for invalid inputs`, () => {
   expect(getCardValue("♠")).toEqual("Invalid card rank.");
   expect(getCardValue("979")).toEqual("Invalid card rank.");
   expect(getCardValue("1")).toEqual("Invalid card rank.");
-  expect(getCardValue([1, 2, 3])).toEqual("Invalid card rank.");
 });


Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -8,9 +8,13 @@
...
 function isProperFraction(numerator, denominator) {
-  if (numerator < denominator) {
+  if(typeof(numerator) != "number" || typeof(denominator) != "number") {
+    return "[ wrong input ]";
+  }
+  if (numerator < denominator ) {
     return true;
   }
+  return false;
 }
```
cjyuan: In mathematics, -A/B == A/-B == -(A/B), and -A/-B == A/B for any integers A and B (B ≠ 0).
They represent a proper fraction if A < B and B ≠ 0.

So `isProperFraction(-4, 3)` should return `false` because 4 >= 3.

So `isProperFraction(-2, 5)` should return `true` because 2 < 5.

Consider comparing the absolute value of the numerator and the denominator instead.


In addition, to indicate errors, it is best practices to ***throw an error*** (instead of returning an error message).
If you have time, you should check out how to throw an error in JS.

Alternatively, you can choose to return `false` to indicate something is not a fraction.

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -47,13 +51,27 @@ assertEquals(improperFraction, false);
...
+assertEquals(equalFraction, false);
 
 // Stretch:
 // What other scenarios could you test for?
+
+// Try 0 and a negative number check:
+// Input: numerator = -10, denominator = 0
+// target output: true
+const tryZero = isProperFraction(-10, 0);
+assertEquals(tryZero, true);
```
cjyuan: -10/0 is not a proper fraction.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -8,9 +8,17 @@
...
+  let rank = card.slice(0, -1);
   if (rank === "A") {
     return 11;
   }
+  if (rank === "J" || rank === "Q"|| rank === "K") {
+    return 10;
+  }
+  if ((rank => 2) && (rank <= 10)) {
+    return Number(rank);
+  }
```
cjyuan: In JavaScript, strings that represent valid numeric literals in the language can be safely converted to equivalent numbers. For examples, "0x02", "2.1", or "0002".

```
getCardValue("0x02♠");
getCardValue("2.1♠");
getCardValue("0002♠");
```

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
```diff
@@ -9,18 +9,18 @@ test("should identify right angle (90°)", () => {
...
+test("should identify Straight angle (180°)", () => {
+  expect(getAngleType(180)).toEqual("Straight angle");
+});
 
-// Case 5: Identify Reflex Angles:
-// When the angle is greater than 180 degrees and less than 360 degrees,
-// Then the function should return "Reflex angle"
+test("should identify Reflex angle (greater than 180° and less than 360°)", () => {
+  expect(getAngleType(181)).toEqual("Reflex angle");
+});
```
cjyuan: We could specify multiple `expect(...)` statements within each `test()` to cover multiple values that belong to the same case. For example,
```
test("should identify reflex angle (180 < angle < 360)", () => {
  expect(getAngleType(300)).toEqual("Reflex angle");
  expect(getAngleType(359.999)).toEqual("Reflex angle");
  expect(getAngleType(180.001)).toEqual("Reflex angle");
});
```

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -7,7 +7,16 @@ test("should return true for a proper fraction", () => {
...
 
 // Case 2: Identify Improper Fractions:
+test("should return false for an improper Fractions", () => {
+  expect(isProperFraction(9, 4)).toEqual(false);
+});
 
 // Case 3: Identify Negative Fractions:
+test("should return true for a negative Fractions", () => {
+  expect(isProperFraction(-6, 4)).toEqual(true);
+});
```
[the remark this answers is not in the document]
M-Abdoon: Done. I added more test cases to each category.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -8,6 +8,25 @@ test("should return 11 for Ace of Spades", () => {
 });
 
 // Case 2: Handle Number Cards (2-10):
+test("should return the number entered as input", () => {
+  const numberCards = getCardValue("4♠");
+  expect(numberCards).toEqual(4);
+});
+
```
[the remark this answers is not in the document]
M-Abdoon: Ok. I changed the code and added more test cases + added different statements using for loop to test numbers from 2 to 10

M-Abdoon:
No worries at all, I learned several things while trying to rename the branch without having it closed :)

The required changes have been made, thank you.

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -8,9 +8,18 @@
...
+  }
+  
+  if ( denominator === 0) {
+    return false;
+  }
+
+  if (Math.abs(numerator) < Math.abs(denominator) ) {
     return true;
   }
+  return false;
```
cjyuan: Challenge: Can you figure out a way to simplify the codes on lines 15-22 to only one line of code?

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -8,6 +8,32 @@ test("should return 11 for Ace of Spades", () => {
...
 // Case 5: Handle Invalid Cards:
+test(`should return string ("Invalid card rank.") for invalid inputs`, () => {
+  expect(getCardValue("HelloWorld!")).toEqual("Invalid card rank.");
+  expect(getCardValue("♠")).toEqual("Invalid card rank.");
+  expect(getCardValue("979")).toEqual("Invalid card rank.");
+  expect(getCardValue("1")).toEqual("Invalid card rank.");
+  expect(getCardValue([1, 2, 3])).toEqual("Invalid card rank.");
+});
```
cjyuan: The function is not expected to validate the suit character or the type of the parameter.

However, the tests on lines 26-29 and 36-37 seem to suggest the function should check also for the suit character.  
The test on line 38 seems to suggest the function should also check the data type of the parameter.

A function that implements stricter validation rules would also pass the same tests, but the implementer reading the tests may end up implementing more than they need to.

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
M-Abdoon: Yes, I actually like doing that, I do that when solving most of the Codewars tests. but since these are exercises, I though they should be easy to read and very clear.
I converted it into one line:
`return Math.abs(numerator) < Math.abs(denominator) && denominator !== 0;`
it will return false if Math.abs(numerator) not smaller that Math.abs(denominator) and if denominator isn't 0; otherwise it will return true.

PR #746, opened 2025-10-10
Glasgow | 25-ITP-SEP | Mohammed Abdoon | Sprint 3 | Coursework/2-practice-tdd

M-Abdoon:
## Learners, PR Template

Self checklist

## Changelist

I modified the function countChar() so it can find a character withing a string and returns its occurrence times if it exists.
also the function getOrdinalNumber() to return "1st" when the input is 1 or 0 when the input number is not 1.
also modified the function repeat() to repeat the str count times, and returns string error if count is negative.
the test files are also updated with the new test cases.

## Questions

No questions.

### DIFF
--- a/Sprint-3/2-practice-tdd/repeat.js
+++ b/Sprint-3/2-practice-tdd/repeat.js
@@ -1,5 +1,5 @@
 function repeat(str, count) {
-  if (typeof count !== "number" || isNaN(count) || count < 0) {
+  if (!Number.isInteger(count) || count < 0) {
     throw new Error("Count must be a valid number");
   }
   return str.repeat(count);


Sprint-3/2-practice-tdd/get-ordinal-number.js
cjyuan: The function should work for any positive integers.

Can you lookup the rules for representing ordinal numbers?

Sprint-3/2-practice-tdd/repeat.js
```diff
@@ -1,5 +1,8 @@
-function repeat() {
-  return "hellohellohello";
+function repeat(str, count) {
+  if(count < 0) {
+    return "Error: Negative counts are not valid.";
+  }
```
cjyuan: How would the caller distinguish the result of the following two function calls?
1. `repeat("Error: Negative counts are not valid.", 1)`
2. `repeat("", -1)`

Sprint-3/2-practice-tdd/repeat.test.js
```diff
@@ -20,13 +20,31 @@ test("should repeat the string count times", () => {
...
 // case: Negative Count:
 // Given a target string str and a negative integer count,
 // When the repeat function is called with these inputs,
 // Then it should throw an error or return an appropriate error message, as negative counts are not valid.
+test("should repeat the string 0 times ( means the output will be an empty string)", () => {
+  const str = "hello";
+  const count = -3;
+  const repeatedStr = repeat(str, count);
+  expect(repeatedStr).toEqual("Error: Negative counts are not valid.");
+});
```
cjyuan: If you modified `repeat()` to throw an error when `count` is negative, and you wanted to test if the function can throw an error as expected, you can use `.toThrow()`. You can find out more about how to use `.toThrow()` here: https://jestjs.io/docs/expect#tothrowerror (Note: Pay close attention to the syntax of the example)

Sprint-3/2-practice-tdd/get-ordinal-number.js
M-Abdoon: Thank you for the feedback. I set the tests for the function and  modified the function getOrdinalNumber() to meet the requirements and to handle all possible cases.

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,7 +1,28 @@
 function getOrdinalNumber(num) {
-  if(num === 1)
-    return "1st";
-  return 0;
+  //
+  if (typeof num !== "number" || !Number.isInteger(num) || num <= 0) {
```
cjyuan: This condition can be shorten. Can you figure out how?

cjyuan (commented):
Changes look good. I have some follow-ups.

Sprint-3/2-practice-tdd/repeat.js
```diff
@@ -1,6 +1,6 @@
 function repeat(str, count) {
-  if(count < 0) {
-    return "Error: Negative counts are not valid.";
+  if (typeof count !== "number" || isNaN(count) || count < 0) {
```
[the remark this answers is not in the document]
cjyuan: Why define a condition that rejects `NaN` but not `Infinity`?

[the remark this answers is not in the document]
M-Abdoon: I didn't know what really the difference was. but I changed the rule and used `Number.isInteger()` instead of isNaN.
`Number.isInteger()` is more general and can detect both NaN and Infinity.

M-Abdoon:
Changes have been made.

cjyuan:
Changes look good. Well done.

PR #758, opened 2025-10-12
Glasgow | 25-ITP-SEP | Alaa Tagi | Sprint 3 | Coursework/sprint 3 practice tdd

Alaa-Tagi:
## Learners, PR Template

Self checklist

## Changelist

I have created a PR with completed tasks as required.

## Questions

No questions

### DIFF
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.js
@@ -1,19 +1,34 @@
 function getOrdinalNumber(num) {
-  num = num.toString();
-  const lastTwo = num.slice(-2);
-  const lastOne = num.slice(-1);
+  if (typeof num !== "number" || isNaN(num)) {
+    return "Invalid input";
+  }
+  const sign = num < 0 ? "-" : "";
+  num = Math.abs(num);
 
-  if (['11', '12', '13'].includes(lastTwo)) {
-    return num + 'th';
+  // Handle decimals
+  if (!Number.isInteger(num)) {
+    return sign + num + "th";
   }
 
+  const lastTwo = num % 100;
+  const lastOne = num % 10;
+
+  // Handle special cases 11th, 12th, 13th
+  if (lastTwo >= 11 && lastTwo <= 13) {
+    return sign + num + "th";
+  }
+
+  // Normal ordinal rules
   switch (lastOne) {
-    case '1': return num + 'st';
-    case '2': return num + 'nd';
-    case '3': return num + 'rd';
-    default:  return num + 'th';
+    case 1:
+      return sign + num + "st";
+    case 2:
+      return sign + num + "nd";
+    case 3:
+      return sign + num + "rd";
+    default:
+      return sign + num + "th";
   }
 }
 
 module.exports = getOrdinalNumber;
-


Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,9 @@
...
-  return 5
-}
+    return stringOfCharacters.split(findCharacter).length - 1;
+  }
 
 module.exports = countChar;
+console.log(countChar('hello', 'l'));
+console.log(countChar('hello', 'z')); 
+console.log(countChar('alaaaaa', 'a'));
+
```
jennethydyrova: Let's keep the PR clean and focused. Could you please remove all console logs from your PR?

Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -17,8 +18,16 @@ test("should count multiple occurrences of a character", () => {
...
 });
 
+  
 // Scenario: No Occurrences
 // Given the input string str,
 // And a character char that does not exist within the case-sensitive str,
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of the char were found in the case-sensitive str.
+
+test("should return 0 for no occurrences", () => {
```
jennethydyrova: This test is good but it doesn't test different possible function inputs. What if the inputs' types are not strings, the `char `string is longer than the `str` string, or you have an empty string for either of the inputs, etc.? Testing different inputs will show whether your function accounts for different scenarios, which is what makes a function robust.

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,23 @@
 function getOrdinalNumber(num) {
-  return "1st";
+    num = num.toString();
+    if (num.slice(-2) === '11' || num.slice(-2) === '12' || num.slice(-2) === '13') {
```
jennethydyrova: Is there way not to repeat `num.slice(-1)` and `num.slice(-2)`?

Sprint-3/2-practice-tdd/repeat.test.js
```diff
@@ -11,22 +11,39 @@ const repeat = require("./repeat");
...
-  expect(repeatedStr).toEqual("hellohellohello");
+  expect(repeatedStr).toEqual("hellohellohellohello");
 });
 
 // case: handle Count of 1:
 // Given a target string str and a count equal to 1,
 // When the repeat function is called with these inputs,
 // Then it should return the original str without repetition, ensuring that a count of 1 results in no repetition.
 
+test("should return the original string when count is 1", () => {
```
jennethydyrova: This test suit are more comprehensive but can you think of other test cases?

jennethydyrova (changes requested):
Hi @Alaa-Tagi! Great job on translating instructions into functions. It was quite easy to read your PR, which is always a good thing for a reviewer!

A couple of things to note:

1. You should try to remove all your testing and scratch work when pushing a PR (mostly console logs in your case).
2. Test cases need to be more comprehensive. Think of it this way: when you write different test cases for your functions, you help yourself catch bugs early. For example, currently your `getOrdinalNumber` function will most likely return `"trueth"` for `num=true`, which I don't think is the expected behavior.

Sprint-3/2-practice-tdd/repeat.test.js
```diff
@@ -11,22 +11,82 @@ const repeat = require("./repeat");
...
+  expect(repeatedStr).toEqual("");
+});
+
+// case: Undefined or missing arguments :
+// When arguments are missing or undefined, it should throw an error.
+test("should handle undefined inputs gracefully", () => {
+  expect(() => repeat(undefined, 3)).toThrow();
+  expect(() => repeat("hello")).toThrow();
+  expect(() => repeat()).toThrow();
+});
```
jennethydyrova: This doesn't look correct to me. Your function doesn't throw any errors in the cases you wrote here.

jennethydyrova (changes requested):

Sprint-3/2-practice-tdd/repeat.test.js
Alaa-Tagi: i fixed the code

Sprint-3/2-practice-tdd/repeat.test.js
jennethydyrova: Have you tried running the tests you wrote?

Alaa-Tagi: Yes, I have and it passed all the tests.

jennethydyrova: How do you run tests?

jennethydyrova: The unit tests in your screenshots do not match those in the PR. Take a look at `Sprint-3/2-practice-tdd/get-ordinal-number.test.js` file, for example. Your last test in this PR is 

```
test("should handle decimal numbers", () => {
  expect(getOrdinalNumber(21.5)).toEqual("21.5th");
});
```

and on the screenshot you sent the last test is

```
test("should return '33rd' for 33", () => {
  expect(getOrdinalNumber(33)).toEqual("33rd");
});
```

Alaa-Tagi: @jennethydyrova,
I have edited this code if you could check down

jennethydyrova: No worries. Okay, I can see these changes and I can see from the function implementation, that this particular unit test on the screenshot you sent will fail but you can understand this by just running unit tests. You have these changes locally, right? If you have same changes locally and you run unit tests, some of them will fail in `Sprint-3/2-practice-tdd` folder indicating that some more work needs to be done.

Alaa-Tagi:
i have editing both get ordinal and repeat code . also i have tried both tests with different laptop it showed failed but in my laptop showed passed i dont know why but i will figure it out later. so, i editing the two codes and tried to run them on another laptop it show passed to all test after editing .

jennethydyrova:
Looks good to me, good job! Regarding not having same code on both computers, double check if you pull all changes from branch. If you work on one laptop and then continue on another one, your changes will not be automatically pulled.

PR #776, opened 2025-10-16
London | 25-ITP-September | Alexandru Pocovnicu | Coursework/sprint 3 | 2 practice tdd

alexandru-pocovnicu:
## Learners, PR Template

Self checklist

## Changelist

In this PR  I have implemented test driven development by first writing the test cases and then implementing the functions to make those tests pass.

### DIFF
--- a/Sprint-3/2-practice-tdd/repeat.test.js
+++ b/Sprint-3/2-practice-tdd/repeat.test.js
@@ -40,6 +40,18 @@ test("should return an empty string", ()=>{
   expect(repeatedStr).toEqual("");
 })
 
+// case: Handle str not a string:
+// Given a number and a count equal to 2,
+// When the repeat function is called with these inputs,
+// Then it should return an error message.
+
+test("should return an error message", ()=>{
+  const str = 5;
+  const count = 5;
+  const repeatedStr = repeat(str, count);
+  expect(repeatedStr).toEqual("Error: str must be a string");
+})
+
 // case: Handle Count of Undefined:
 // Given a target string str and a count equal to Undefined,
 // When the repeat function is called with these inputs,
@@ -48,6 +60,18 @@ test("should return an empty string", ()=>{
 test("should return an error message", ()=>{
   const str = "hello";
   const count = undefined;
+  const repeatedStr = repeat(str, count);
+  expect(repeatedStr).toEqual("Error : count must be a positive integer");
+})
+
+// case: Handle Count of array:
+// Given a target string str and a count equal to [],
+// When the repeat function is called with these inputs,
+// Then it should return an error message.
+
+test("should return an error message", ()=>{
+  const str = "hello";
+  const count = [];
   const repeatedStr = repeat(str, count);
   expect(repeatedStr).toEqual("Error : count must be a positive integer");
 })


Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,26 @@
...
+
+  if(lastTwoDigits>=10 && lastTwoDigits<=13){
+    return `${num}th`;
+  }
+
+  if (lastDigit === 1) {
+    return `${num}st`;
+  } else if (lastDigit === 2) {
+    return `${num}nd`;
+  } else if (num === 3) {
```
jennethydyrova: Are you sure this is the right condition? Looks like a bug.

Sprint-3/2-practice-tdd/repeat.test.js
```diff
@@ -21,12 +21,33 @@ test("should repeat the string count times", () => {
 // When the repeat function is called with these inputs,
 // Then it should return the original str without repetition, ensuring that a count of 1 results in no repetition.
 
+test("should not repeat the string", ()=>{
```
jennethydyrova: Same as above, these test cases are good but they are a bit slim. You could add couple of more tests to test different variations of count and try to break your function by adding some tests for edge cases, like in the case you pass str="".

Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -22,3 +22,44 @@ test("should count multiple occurrences of a character", () => {
...
+});
+
+test("should return 0 if the character doesn't appear in the string", () => {
+  const str = "aaca";
+  const char = "ab";
+  const count = countChar(str, char);
+  expect(count).toEqual(0);
+});
+
+test("should return 0 if the character doesn't appear in the string", () => {
```
jennethydyrova: This is fine if this is the behaviour you designed for your function but if `str` is empty, can there be any character? Should the function have a guardrail to protect against this corner case?

It's fine to leave this as is, you did a great job on adding the tests. It's just something to think about.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -8,6 +8,63 @@ const getOrdinalNumber = require("./get-ordinal-number");
 // When the number is 1,
 // Then the function should return "1st"
 
+test("should return NaN for 1.5", () => {
+  expect(getOrdinalNumber(1.5)).toEqual(NaN);
+});
+
+test("should return NaN for 0", () => {
```
jennethydyrova: This is not critical but it doesn't really make sense. 0 is a number and returning NaN for it is a bit confusing. Same for 1.5. You might want to return something else if you don't support those numbers.

Sprint-3/2-practice-tdd/repeat.test.js
alexandru-pocovnicu: I added a few more tests for str as for count the requirement says it's a positive integer , no point to test for NaN etc, I also commented out the test for count = 0 , 0 not being a positive integer.

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,15 @@
...
+    }
+    if(stringOfCharacters.substring(i, i + findCharacter.length ) === findCharacter){
+      char++;
+    }
+  }
+  return char
 }
 
 module.exports = countChar;
+console.log(countChar("aaaaa","a"));
```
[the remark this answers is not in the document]
alexandru-pocovnicu: 🫣 done

Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -22,3 +22,9 @@ test("should count multiple occurrences of a character", () => {
 // And a character char that does not exist within the case-sensitive str,
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of the char were found in the case-sensitive str.
+test("should return 0 if the character doesn't appear in the string",()=>{
```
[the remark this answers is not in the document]
jennethydyrova: I would suggest discussing this with your classmates as it's quite important to understand why we write this kind of test cases and how we write them.

jennethydyrova:
@alexandru-pocovnicu I set the label to complete but just noticed that the check is failing, can you please fix it?

Screenshot 2025-10-28 at 10 45 09

PR #777, opened 2025-10-17
West Midlands | ITP-Sept-25 | Mustaf Asani | Sprint 2 | coursework/sprint-2

asaniDev:
## Learners, PR Template

Self checklist

## Changelist

finished all the tasks for coursework sprint 2

### DIFF
--- a/Sprint-2/5-stretch-extend/format-time.js
+++ b/Sprint-2/5-stretch-extend/format-time.js
@@ -18,11 +18,18 @@ function formatAs12HourClock(time) {
     return `12:${minutes} am`;
   }
 
+  if (time.charAt(0) === "0") {
+    return `${time.slice(1)} am`;
+  }
+
   return `${time} am`;
 }
 
-const currentOutput = formatAs12HourClock("00:30");
-const targetOutput = "12:30 am";
+console.log(`the 12 hr format for 01:01 is ${formatAs12HourClock("05:01")}`);
+console.log(`the 12 hr format for 13:01 is ${formatAs12HourClock("13:01")}`);
+
+const currentOutput = formatAs12HourClock("01:01");
+const targetOutput = "1:01 am";
 console.assert(
   currentOutput === targetOutput,
   `current output: ${currentOutput}, target output: ${targetOutput}`


Sprint-2/1-key-errors/1.js
```diff
@@ -2,19 +2,21 @@
...
 
 function convertToPercentage(decimalNumber) {
-  const decimalNumber = 0.5;
+  //const decimalNumber = 0.5;
   const percentage = `${decimalNumber * 100}%`;
 
   return percentage;
 }
 
-console.log(decimalNumber);
```
cjyuan: We could also move the declaration at line 10 outside the function without modifying the `console.log` statement.

Variables in different **scopes** can have the same name and the JS language has a set of rules to resolve "conflicting" names. 

Do look up these two concepts, ***identifier scope*** and ***identifier resolution***, in the context of JavaScript programming. ChatGPT can give a good explanation.

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,14 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+  // return the BMI of someone based off their weight and height
+  const bmi = weight / height ** 2;
+  return bmi.toFixed(1);
```
cjyuan: What **type** of value do you expect the function to return? A number or a string?
Does your function return the type of value you expect?

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,8 +4,14 @@
...
+  if (hours == 12) {
+    return `${hours}:${minutes} pm`;
+  }
+
   if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+    return `${hours - 12}:${minutes} pm`;
   }
   return `${time} am`;
 }
```
cjyuan: What do you expect from the following function calls? 
```javascript
  formatAs12HourClock("00:00");
  formatAs12HourClock("01:01");
  formatAs12HourClock("13:01");    
```

Does your function return the value you expected?

Sprint-2/3-mandatory-implement/1-bmi.js
asaniDev: the function will return a string representation of the equivalent number with the specified number of decimal places. I had expected it to produce a number but because i used the toFixed() method to shorten the answer to the required number of decimal places this changed.

Sprint-2/5-stretch-extend/format-time.js
cjyuan: What about making the format of the returned string values from these two function calls more consistent?
```
  formatAs12HourClock("01:01");
  formatAs12HourClock("13:01");
```

asaniDev: thank you i had missed that, made the format for both times match, romved the trailing zero.

PR #785, opened 2025-10-21
London | 25-ITP-September | Ammad Ur Rehman | Sprint 1 | 03 Complete Sprint 1 Coursework

anosidium:
## Learners, PR Template

Self checklist

## Changelist

I answered all the questions in the `.js` and `.md` files

## Questions

According to the console in the Web Inspector (Safari) or JavaScript Console (Chrome), `Number` and `String` and other types are "`function`" which confuses me because I expected them to be classified as `Object`.

In Swift programming language, `String`, `Int`, `Double`, etc. are `struct` type and not `function`.

Why are `function`s that belong to an `Object` are called `function`? Why not called `method`? That's how it is in Swift. Functions are freestanding and belong to no objects, struct, etc.

### DIFF
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -10,6 +10,6 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 
 // num represents a random number in the interval [1,100]
 // First, it starts with 100 - 1 + 1, which is equals to 100
-// Secondly, it multiply with Math.random(), which the function returns any number in the interval [0,1]
+// Secondly, it multiply with Math.random(), which the function returns any number in the interval [0,1)
 // Thirdly, Math.floor() rounds down and returns the largest integer less than or equal to a given number
 // Finally, it adds 1
--- a/Sprint-1/3-mandatory-interpret/1-percentage-change.js
+++ b/Sprint-1/3-mandatory-interpret/1-percentage-change.js
@@ -12,7 +12,7 @@ console.log(`The percentage change is ${percentageChange}`);
 // Read the code and then answer the questions below
 
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
-// Line 4, 5, 10
+// 5 (2 × Number(), 2 × replaceAll(), 1 × log())
 
 // b) Run the code and identify the line where the error is coming from - why is this error occurring? How can you fix this problem?
 // It's missing a comma between the arguments at line 5 in replaceAll() method
--- a/Sprint-1/4-stretch-explore/chrome.md
+++ b/Sprint-1/4-stretch-explore/chrome.md
@@ -26,4 +26,6 @@ It displays an alert with a textfield where you can enter some text, e.g. John.
 
 What is the return value of `prompt`?
 
-The string returned from the textfield what the user has entered otherwise null if the user left it blank.
+If the user has entered text, the return value is a string containing the entered text.
+If the user leaves the text field blank and clicks the OK button, the return value is an empty string.
+If the user enters a name or leaves the text field blank but clicks the Cancel button, the return value is null.


Sprint-1/1-key-exercises/1-count.js
```diff
@@ -4,3 +4,5 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+
+// Line 3 is a reassignment of the variable `count` to its current value plus 1
```

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -7,3 +7,9 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+// num represents a random number in an inclusive range from 1 to 100
+// First, it starts with 100 - 1 + 1, which is equals to 100
+// Secondly, it multiply with Math.random(), which the function returns any number between 0 and 1, inclusive.
```
cjyuan: We can also use the concise and precise ***interval notation*** to describe a range of values. 
- `[`, `]` => inclusion
- `(`, `)` => exclusion

Sprint-1/4-stretch-explore/chrome.md
```diff
@@ -12,7 +12,18 @@ invoke the function `alert` with an input string of `"Hello world!"`;
...
+
 Now try invoking the function `prompt` with a string input of `"What is your name?"` - store the return value of your call to `prompt` in an variable called `myName`.
 
 What effect does calling the `prompt` function have?
+
+It displays an alert with a textfield where you can enter some text, e.g. John.
+
 What is the return value of `prompt`?
+
+The string returned from the textfield what the user has entered otherwise null.
```
cjyuan: When does the function return `null`?

cjyuan (commented):
Regarding the question you asked in the PR description, have you tried asking an AI?

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -12,7 +12,7 @@ console.log(`The percentage change is ${percentageChange}`);
 // Read the code and then answer the questions below
 
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
-// Line 4, 5
+// Line 4, 5, 10
```
cjyuan: That still does not indicate how many function calls? And the answer is not 3.

Sprint-1/4-stretch-explore/chrome.md
```diff
@@ -26,4 +26,4 @@ It displays an alert with a textfield where you can enter some text, e.g. John.
 
 What is the return value of `prompt`?
 
-The string returned from the textfield what the user has entered otherwise null.
+The string returned from the textfield what the user has entered otherwise null if the user left it blank.
```
cjyuan: What does the function return when the user does not enter anything and click the "OK" button? `null` or an empty string `""`?

When does the function return `null`?

Sprint-1/4-stretch-explore/chrome.md
```diff
@@ -26,4 +26,6 @@ It displays an alert with a textfield where you can enter some text, e.g. John.
 
 What is the return value of `prompt`?
 
-The string returned from the textfield what the user has entered otherwise null if the user left it blank.
+If the user has entered text, the return value is a string containing the entered text.
+If the user leaves the text field blank and clicks the OK button, the return value is an empty string.
+If the user enters a name or leaves the text field blank but clicks the Cancel button, the return value is null.
```
cjyuan: This is accurate but could be simplified. Some of the info is not very useful.

Knowing how to explain or describe code is an important skill in software development.

Suggestion: Use AI to help you figure out how to make the description more concise.

cjyuan (commented):
> I was under the impression that AI tools should be avoided and that we should instead ask other trainees and volunteers, as well as use Google.

Not at all. Asking AI is not much different from asking volunteers or using Google search. You just need to **validate** the answer and make sure you truly **understand** the answer (to the point that you can **reproduce** the solution or **explain** the solution).

AI is here to stay, and knowing how to use it as a tool to find answer and solve problem is a critical skill to flourish in the IT field.

Why don't you try your question (in the PR description) on AI. If you present your answer here, I will share my view with you.

---
Changes look good. I will mark this as "Complete" first.

PR #794, opened 2025-10-23
London | 25-ITP-SEP | Imran Mohamed| Sprint 3 | Coursework/sprint 3  practice tdd

i786m:
Self checklist

## Changelist

I have implemented tests, and used tdd to implement the functions set out in the coursework.

### DIFF
--- a/Sprint-3/2-practice-tdd/repeat-str.test.js
+++ b/Sprint-3/2-practice-tdd/repeat-str.test.js
@@ -96,6 +96,5 @@ test("should return error message for string input with undefined count", () => 
 
 test('should have the correct amount of arguments', () => { 
   expect(() => repeatStr('hello')).toThrow(new Error("Function requires exactly two arguments: a string and a count. Received 1 arguments"));
-  expect(() => repeatStr("hello", 3, 3)).toThrow(new Error("Function requires exactly two arguments: a string and a count. Received 3 arguments"));
  })
 


Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -11,3 +11,51 @@ const getOrdinalNumber = require("./get-ordinal-number");
 test("should return '1st' for 1", () => {
   expect(getOrdinalNumber(1)).toEqual("1st");
 });
+
+test("should return '2nd' for 2", () => {
```
jennethydyrova: Same as my previous comment. What if someone passes "3", undefined, or another unexpected value to your function? I understand your assumptions, but users of your functions (both internal within your team and external if you build some sort of API) can pass values you didn't safeguard against, which can break your function.

jennethydyrova (changes requested):
Hi @i786m! Kudos on writing clean and readable functions! You did a good job with the unit tests as well. However, unit tests are designed to catch bugs from unexpected usage, which is why it's important to pass different types of parameters and test various scenarios. Please make your test cases more thorough to ensure your functions handle edge cases properly.

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,39 @@
...
+  }
+  if (typeof num !== "number" || isNaN(num)) {
+    throw new Error("Input must be a number");
+  }
+  if (!Number.isFinite(num)){
+    throw new Error("Input must be a finite number");
+  }
+  if (!Number.isInteger(num) || num < 0) {
+    throw new Error("Input must be a non-negative integer");
+  }
```
jennethydyrova: This is a good validation but I think this could be simplified because at the end your input should be a `Non-negative integer`. For example, `Number.isFinite()` and `isNan()` have some overlap, can you check in the MDN docs? 

I am leaving this as an optional comment, no action needs to be taken to complete this PR but I would like you think if this could be written simpler. You do not need to throw different error for each small case, you can have more general error for anything that is negative non-integer.

Sprint-3/2-practice-tdd/repeat.js
```diff
@@ -1,5 +1,14 @@
...
+function repeat(str, count) {
+  if ( arguments.length !== 2) {
+    return "Function requires exactly 2 arguments";
+  }
+  if (typeof str !== "string") {
+    return "First argument must be a string";
+  }
+  if (!Number.isInteger(count) || count < 0) {
+    return "Second argument must be a non-negative integer";
+  }
```
jennethydyrova: I see you extensively use throw Error in other functions, is there a reason you want to use just return here?

Small point, no action required to complete this PR but you could improve your error messages to be something like `throw Error("Second argument must be a non-negative integer, received: ${count} (${typeof count})")` (same for str). It would be a bit easier to debug.

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,24 @@
...
+  if (typeof findCharacter !== "string") {
+    throw new Error("Second argument must be a string.");
+  }
+  if (findCharacter.length !== 1) {
+    throw new Error("Character to find must be a single character.");
+  } 
+  if (stringOfCharacters.length === 0) {
+    return 0;
+  }
+  console.log(Array.from(stringOfCharacters));
```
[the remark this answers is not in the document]
i786m: I have removed console log, will ensure future PRs are free from debugging logs

Sprint-3/2-practice-tdd/repeat.js
i786m: changed returns to throw errors to match the rest in this sprint

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,22 @@
...
+  if (typeof stringOfCharacters !== 'string'){
+    throw new Error("First argument must be a string.");
+  }
+  if (typeof findCharacter !== "string") {
+    throw new Error("Second argument must be a string.");
+  }
+  if (findCharacter.length !== 1) {
+    throw new Error("Character to find must be a single character.");
+  } 
+  if (stringOfCharacters.length === 0) {
```
wheresdiasd: Would that be the same as 
`(!stringOfCharacters.length)` ?

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,22 @@
...
+    return 0;
+  }  
+  return Array.from(stringOfCharacters).filter(char => char === findCharacter).length;
```
wheresdiasd: Could we use spread operator here ? [...stringOfCharacters] ?

Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -22,3 +22,34 @@ test("should count multiple occurrences of a character", () => {
 // And a character char that does not exist within the case-sensitive str,
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of the char were found in the case-sensitive str.
+test("should return  0 if no occurrences of a character", () => {
```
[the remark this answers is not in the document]
wheresdiasd: 💯 @jennethydyrova

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
wheresdiasd: 💯

Sprint-3/2-practice-tdd/repeat.js
```diff
@@ -1,5 +1,15 @@
-function repeat() {
-  return "hellohellohello";
+function repeat(str, count) {
+  if ( arguments.length !== 2) {
```
wheresdiasd: Same about the args here, could we be more specific ?

PR #795, opened 2025-10-23
West Midlands |  ITP-Sept-25 | Mustaf Asani | Sprint 3 | Coursework/sprint 3 implement and rewrite

asaniDev:
## Learners, PR Template

Self checklist

## Changelist

created functions and practiced test driven development by making tests for the functions.

### DIFF
--- a/Sprint-2/1-key-errors/2.js
+++ b/Sprint-2/1-key-errors/2.js
@@ -17,3 +17,4 @@ function square(3) {
 
 // =============> write your new code here
 
+


Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -8,8 +8,20 @@
 // write one test at a time, and make it pass, build your solution up methodically
 
 function isProperFraction(numerator, denominator) {
+  if (Math.sign(numerator) === -1 || Math.sign(denominator) === -1) {
```
cjyuan: - There is an easier syntax to check if a number is negative.

- And since you are using `Math.abs()`, checking for negatives is optional.

- This function can be further simplified. If you are up to the challenge, go for a solution that uses at most one if-statement.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -8,8 +8,18 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
-  if (rank === "A") {
+  const rank = card[0];
```
cjyuan: The rank is the part without the suit character (the last character). `card[0]` yields only a string with the first character.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -8,8 +8,18 @@
...
+
+  if (card === "A♠") {
+    return 11;
+  } else if (Number(rank) > 1 && Number(rank) < 10) {
+    return Number(rank);
+  } else if (rank === "10" || rank === "J" || rank === "Q" || rank === "K") {
```
cjyuan: Does your function return the value you expected from each of the following function calls?

```
getCardValue("10♠");
getCardValue("100♠");
getCardValue("3.1416♠");
```

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -7,7 +7,16 @@ test("should return true for a proper fraction", () => {
...
 
 // Case 2: Identify Improper Fractions:
+test("should return false for an improper fraction", () => {
+  expect(isProperFraction(7, 4)).toEqual(false);
+});
 
 // Case 3: Identify Negative Fractions:
+test("should return true for a proper fraction based on the absolute values of the numerator and denominator", () => {
+  expect(isProperFraction(-3, 8)).toEqual(true);
+});
```
cjyuan: To make the test more comprehensive, we could test multiple samples like this:

```javascript
test("should return true for a proper fraction based on the absolute values of the numerator and denominator", () => {
  expect(isProperFraction(-3, 8)).toEqual(true);
  expect(isProperFraction(-3, -8)).toEqual(true);
  expect(isProperFraction(3, -8)).toEqual(true);
  expect(isProperFraction(3, 8)).toEqual(true);      
})
```

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
asaniDev: removed the check for negative and took off the first if else block, not quite sure how to reduce it to one if statement without making risky assumptions. since the function only returns true for a proper fraction and false for everything else, would it be appropriate to have an if statement that checks for a proper fraction and else that gives false for any other scenario?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -8,8 +8,18 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
-  if (rank === "A") {
+  const rank = card[0];
+
+  if (card === "A♠") {
+    return 11;
```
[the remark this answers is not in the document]
asaniDev: I thought this was to check specifically for ace of spades, on taking a closer look I realised this did not matter as the other condition returns the desired value as well. removed this condition.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -8,11 +8,13 @@
...
 
-  if (card === "A♠") {
-    return 11;
-  } else if (Number(rank) > 1 && Number(rank) < 10) {
+  for (let i = 0; i < card.length - 1; i++) {
+    rank += card[i];
+  }
+
+  if (Number.isInteger(Number(rank)) && Number(rank) > 1 && Number(rank) < 10) {
     return Number(rank);
```
Do you want to recognize these string values as valid ranks?

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -9,14 +9,19 @@ test("should return true for a proper fraction", () => {
 // Case 2: Identify Improper Fractions:
 test("should return false for an improper fraction", () => {
   expect(isProperFraction(7, 4)).toEqual(false);
+  expect(isProperFraction(35, 8)).toEqual(false);
```
cjyuan: Could also test some negative improper fractions in this test.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -23,7 +18,12 @@ test("should return 11 for Ace", () => {
...
-  const invalidCard = getCardValue("1♥");
+test("should return invalid card rank for cards that are not in the suite", () => {
+  const invalidCard = getCardValue("100♥");
+  expect(invalidCard).toEqual("Invalid card rank.");
+});
+
+test("should return invalid card rank for cards that are not in the suite", () => {
+  const invalidCard = getCardValue("3.1416♥");
   expect(invalidCard).toEqual("Invalid card rank.");
 });
```
cjyuan: They both test "invalid case". Consider combine them into one test but using multiple samples.

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -8,20 +8,14 @@
...
-
-  if (numerator < denominator) {
+  if (Math.abs(numerator) < Math.abs(denominator)) {
     return true;
   } else if (numerator > denominator) {
     return false;
   } else if (numerator === denominator) {
     return false;
+  } else if (numerator === 0) {
+    return false;
```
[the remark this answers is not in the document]
asaniDev: after tracing out the code execution for these pairs I realised I only needed 1 if statement for the exercises and also corrected an error on my fraction with zero as a numerator, this is a proper fraction. I also added an extra test to check for fractions with 0 as the denominator as these are indefinite and cannot be proper or improper.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -2,19 +2,14 @@
...
-test("should return 11 for Ace of Spades", () => {
-  const aceofSpades = getCardValue("A♠");
-  expect(aceofSpades).toEqual(11);
-});
-
 // Case 2: Handle Number Cards (2-10):
-test("should return a number matchig the rank value for given card", () => {
+test("should return a number matchig the rank value for number cards", () => {
   const numberCard = getCardValue("5♠");
   expect(numberCard).toEqual(5);
```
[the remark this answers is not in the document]
asaniDev: added a suite of tests for different values to make the test more robust

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -8,8 +8,31 @@
...
+    return false;
+  }
+
+  let rank = card.slice(0, card.length - 1);
+
+  if (strictNumber(rank)) {
+    if (
+      Number.isInteger(Number(rank)) &&
+      Number(rank) > 1 &&
+      Number(rank) <= 10
```
cjyuan: It is possible to define a regex that matches exactly the valid number ranks so that the checks on lines 22-24 can be omitted. 

Can you figure out how to define such regex?

cjyuan (commented):
- The tests are now very solid. Well done.

- This PR branch still contains unrelated changes in "Sprint-2" folder. Can you revert the changes made in the "Sprint-2" folder to make the PR branch clean?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -8,11 +8,33 @@
...
+    if (pattern.test(Number(rank))) {
+      return Number(rank);
+    }
+  }
+
+  const faceCardPattern = /^[JQK]$/;
+
+  if (faceCardPattern.test(rank)) {
+    return 10;
+  } else if (/^A$/.test(rank)) {
```
cjyuan: Could also just write `rank == 'A'`.

cjyuan (commented):
Changes look good. Well done.

PR #798, opened 2025-10-23
West Midlands | 25 Sep ITP | Iswat Bello | Sprint 3 | 2 Practice TDD

Iswanna:
## Learners, PR Template

Self checklist

## Changelist

This pull request includes modifications to the **2-practice-tdd** directory. I added test cases for the various functions in the files, and I also implemented the functions to ensure all tests pass.

## Questions

Hi. Please could you review my PR? I’d really appreciate your feedback?

### DIFF
--- a/Sprint-3/2-practice-tdd/repeat.test.js
+++ b/Sprint-3/2-practice-tdd/repeat.test.js
@@ -25,7 +25,7 @@ test("should return the original input with no repetition", () => {
   const count = 1;
   const repeatedStr = repeat(str, count);
   expect(repeatedStr).toEqual(str);
-})
+});
 
 // case: Handle Count of 0:
 // Given a target string str and a count equal to 0,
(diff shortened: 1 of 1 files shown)


Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -22,3 +22,9 @@ test("should count multiple occurrences of a character", () => {
 // And a character char that does not exist within the case-sensitive str,
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of the char were found in the case-sensitive str.
+test("should return 0 occurence of a character", () => {
```
jennethydyrova: This is a good start but this is would be considered as a good test suite as it covers just one behavior. Good test suite should cover different scenarios based on what type of arguments you pass to the function.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -11,3 +11,15 @@ const getOrdinalNumber = require("./get-ordinal-number");
 test("should return '1st' for 1", () => {
   expect(getOrdinalNumber(1)).toEqual("1st");
 });
+
+test("should return '11th' for 11", () => {
```
jennethydyrova: Same as a previous comment regarding very slim test suite. Try to think of more test cases.

Sprint-3/2-practice-tdd/repeat.test.js
```diff
@@ -20,13 +20,31 @@ test("should repeat the string count times", () => {
 // Given a target string str and a count equal to 1,
 // When the repeat function is called with these inputs,
 // Then it should return the original str without repetition, ensuring that a count of 1 results in no repetition.
+test("should return the original input with no repetition", () => {
```
jennethydyrova: These unit tests are good, great job on writing them! Now I would like you to think of more test cases, where the values passed to the function are not what you expected. What if you pass integer 3 as `valueToRepeat` or undefined as `numOfTimes`, etc.?

jennethydyrova (changes requested):
hi @Iswanna! Good start on the code in this PR! There are a few things to consider while writing unit tests:

1. You should try to cover more test cases to see how your function behaves in different scenarios.
2. You should test edge cases like empty strings, undefined inputs, different data types, etc. This helps catch bugs early and makes your function more robust.

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,21 @@
 function countChar(stringOfCharacters, findCharacter) {
-  return 5
+  let characterOccurrence = "";
+  if (
```
jennethydyrova: It's usually better to return early if inputs are invalid, it's called `early return`. Can you think of a way to implement this in your function?

jennethydyrova (commented):
Hi @Iswanna! Really great job on writing unit tests and improving your functions. There are couple of common issues in the functions you wrote. I left comments on the `countChar` function but those comments are applicable to other functions as well.

1. You should try to return early from the function if the arguments are invalid. This makes your function much cleaner and there is really no need to execute the rest of the code if function input is invalid.
2. You should be throwing error or at least returning error message without assigning it to a variable. This is semantically incorrect, please try to read the function out loud and think if assigning error message to a variable that should hold something else make your function easier or harder to read. 
3. When you find yourself repeating same code over and over again, try to think how it can be combined into more general structures or variables. Having similar if conditions can be combined into one or having multiple calculations of the same number can be saved in a variable or moved to function.

Something I want to highlight here is how you write commit messages. They are extremely clean and nicely written, really great job on it! If your were working in a team, your teammates would appreciate it a lot as it's very easy to track your changes.

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,23 @@
 function countChar(stringOfCharacters, findCharacter) {
-  return 5
+  // Check both inputs are strings
+  if (
+    typeof stringOfCharacters !== "string" ||
+    typeof findCharacter !== "string"
+  ) {
+    return "Invalid input: input should be a string";
```
jennethydyrova: It's okay to return error message like this as you are just at the beginning of learning js, but there is a better way of doing it. This is optional but can you think of how you can make use of the following https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw in the context of error handling in your functions?

jennethydyrova (changes requested):
Hi @Iswanna! Very good iteration on the last version! I left couple of comments, some of them are optional and you don't need to change anything to get your PR accepted but I think it would be useful for you to try to think of better ways of writing some parts.

Iswanna:
> Hi @Iswanna! Very good iteration on the last version! I left couple of comments, some of them are optional and you don't need to change anything to get your PR accepted but I think it would be useful for you to try to think of better ways of writing some parts.

Hi @jennethydyrova. 
Thank you for the feedback! I’ll carefully utilise all your comments, even the optional ones, as they’ll help me improve my coding skills and think of better ways to write my code.

Sprint-3/2-practice-tdd/count.js
Iswanna: > You don’t need to act on optional comments right now, this one was optional, but feel free to if you’d like!

Thank you for clarifying! I’ll keep the optional comments in mind and may try them out later to improve the code further.

Sprint-3/2-practice-tdd/repeat.test.js
```diff
@@ -20,13 +20,98 @@ test("should repeat the string count times", () => {
...
+// case: Non-integer negative count
+test("should return an error message for non-integer negative count", () => {
+  const str = "banana";
+  const count = -1.7;
+  const repeatedStr = repeat(str, count);
+  expect(repeatedStr).toEqual("Invalid count: count should be an integer");
+});
+
+// case: Object input
+test("should repeat an object count times as a string", () => {
```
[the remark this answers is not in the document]
Iswanna: > Hmm, I am not sure if you want to repeat empty object like this. I don't see value in doing so. I think the requirement of this function is to repeat string n times. If what you pass is not a string, should it perhaps return an error as you did in other functions?

Hi @jennethydyrova.

I am trying to fix my code. Please, I would like your opinion. Should I let the number input (e.g. 2) repeat or return an error message?

PR #799, opened 2025-10-23
West Midlands | 25-ITP-Sep | Baba Yusuf | Sprint 3 | Coursework/sprint-3/2-practice-tdd

Baba05206:
## Learners, PR Template

Self checklist

## Changelist

I completed the three practice exercises in 2-practice-tdd

### DIFF
--- a/Sprint-3/2-practice-tdd/count.js
+++ b/Sprint-3/2-practice-tdd/count.js
@@ -1,24 +1,31 @@
 function countChar(stringOfCharacters, findCharacter) {
-  if (stringOfCharacters == null) {
-    return "Error: The input string cannot be null or undefined.";
+  // 1. Validate 'stringOfCharacters' input
+  if (stringOfCharacters === null || stringOfCharacters === undefined) {
+    throw new Error("The input string cannot be null or undefined.");
+  }
+  if (typeof stringOfCharacters !== "string") {
+    throw new Error("The input string must be a string type.");
   }
 
+  // 2. Validate 'findCharacter' input
+  if (typeof findCharacter !== "string" || findCharacter.length !== 1) {
+    // This handles null, undefined, and strings that aren't a single character (including empty string "")
+    throw new Error(
+      "The character to count must be a single character string."
+    );
+  }
+
+  // 3. Handle the empty string case (returns 0)
   if (stringOfCharacters.length === 0) {
-    return "Error: The string cannot be empty.";
+    return 0;
   }
 
-  if (findCharacter == null) {
-    return "Error: The character to count must be a single character.";
-  }
-
-  const findChars = [...findCharacter];
-  if (findChars.length !== 1) {
-    return "Error: The character to count must be a single character.";
-  }
-
+  // 4. Perform the count
   let count = 0;
   for (let ch of stringOfCharacters) {
-    if (ch === findCharacter) count++;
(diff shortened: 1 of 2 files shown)


Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,42 @@
...
+  }
+
+  if (stringOfCharacters.length === 0) {
+    return "Error: The string cannot be empty.";
+  }
+
+  const chars = [...stringOfCharacters];
+  const findChars = [...findCharacter];
+
+  if (!findCharacter || findChars.length !== 1) {
```
cjyuan: Why check `!findCharacter` when you are also checking `findChars.length !== 1`?

Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -22,3 +21,101 @@ test("should count multiple occurrences of a character", () => {
...
+});
+
+// Scenario: Empty String
+// Given an empty input string str,
+// And any character char,
+// When the function is called with these inputs,
+// Then it should return 0, as there are no characters to count.
+test("should return error if the string is empty", () => {
+  expect(countChar("", "a")).toEqual("Error: The string cannot be empty.");
+});
```
cjyuan: This test description and the behavior of the function do not quite match the description in the comment.

Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -22,3 +21,101 @@ test("should count multiple occurrences of a character", () => {
...
+
+// Scenario: Unicode Characters
+// Given the input string str containing Unicode characters (e.g., emojis),
+// And a character char that is a Unicode character,
+// When the function is called with these inputs,
+// Then it should return the number of times that Unicode character occurs in str.
+test("should count unicode characters correctly", () => {
+  expect(countChar("😀😃😀😄", "😀")).toEqual(2);
+});
```
cjyuan: It is not needed to check strings containing emoji in this exercise but this is a thoughtful consideration.

cjyuan (commented):
The tests are very comprehensive. Nicely done.

Sprint-3/2-practice-tdd/count.test.js
cjyuan: Where do you get the idea the function should return an error when the first parameter is an empty string?

Shouldn't the comments in the original file serve as the spec?

Baba05206: Should be 0. I got myself all confused. I apologize for all these... I should know enough to prevent you having to point out this many issues. I would do better going forward

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,36 @@
 function countChar(stringOfCharacters, findCharacter) {
-  return 5
+  // 1. Validate 'stringOfCharacters' input
+  if (stringOfCharacters === null || stringOfCharacters === undefined) {
+    throw new Error("The input string cannot be null or undefined.");
+  }
+  if (typeof stringOfCharacters !== "string") {
+    throw new Error("The input string must be a string type.");
+  }
```
cjyuan: Why treat `null` and `undefined` differently from other non-string types?

PR #805, opened 2025-10-24
Manchester | 25-ITP-Sep | Mahtem T. Mengstu | Sprint 2 | coursework/sprint-2 

Mahtem:
## Learners, PR Template

Self checklist

## Changelist

Completed course works of Sprint 2 from module structuring and testing data

## Questions

Questions and blockers will be forwarded to slack.

### DIFF
--- a/Sprint-2/1-key-errors/1.js
+++ b/Sprint-2/1-key-errors/1.js
@@ -33,4 +33,4 @@ console.log(convertToPercentage(0.27));
 console.log(convertToPercentage(0.76));
 
 // The new code run smoothly and retuned "50%"
-// DecimalNumber is now taken as a parameter. 
+// decimalNumber is now taken as a parameter. 
--- a/Sprint-2/5-stretch-extend/format-time.js
+++ b/Sprint-2/5-stretch-extend/format-time.js
@@ -3,56 +3,56 @@
 // Your task is to write tests for as many different groups of input data or edge cases as you can, 
 // and fix any bugs you find.
 
-function formatAs12HourClock(time) {
-  const hours = Number(time.slice(0, 2));
-  if (hours > 12) {
-    return `${hours - 12}:00 pm`;
-  }
-  return `${time} am`;
-}
+// function formatAs12HourClock(time) {
+//   const hours = Number(time.slice(0, 2));
+//   if (hours > 12) {
+//     return `${hours - 12}:00 pm`;
+//   }
+//   return `${time} am`;
+//}
 
-const currentOutput = formatAs12HourClock("08:00");
-const targetOutput = "08:00 am";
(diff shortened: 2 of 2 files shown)


Sprint-2/1-key-errors/1.js
```diff
@@ -1,20 +1,34 @@
...
+// Finally, correct the code to fix the problem
+// =============> write your new code here
+
+
+function convertToPercentage() {
   const decimalNumber = 0.5;
   const percentage = `${decimalNumber * 100}%`;
 
   return percentage;
 }
```
cjyuan: Without the parameter, the function would always return "50%", which is not very useful.

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -23,3 +24,63 @@ console.assert(
...
+
+// console.assert(formatAs12HourClock("08:00") === "08:00 am"); 
+// console.assert(formatAs12HourClock("23:00") === "11:00 pm");
+// console.assert(formatAs12HourClock("22:00") === "10:00 pm");
+
+//  Modified Code:
+
+function formatAs12HourClock(time) {
+  let hours = Number(time.slice(0, 2));
+  const minutes = time.slice(3);
```
cjyuan: We could also pass a parameter to .slice() to clearly express the intent to extract the last two characters, 
without needing to calculate the starting position manually. Can you find out what this parameter is?

Sprint-2/1-key-errors/1.js
Mahtem: > Without the parameter, the function would always return "50%", which is not very useful.

Thank you for pointing out the case. "decimalNumber" is now taken as a parameter.

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -23,3 +24,63 @@ console.assert(
...
+
+  let suffix;
+  if (hours >= 12) {
+    suffix = "pm";
+  } else {
+    suffix = "am";
+  }
+
+  hours = hours % 12 || 12; // convert 0 to 12, 13 to 1
```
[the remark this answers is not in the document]
Mahtem: > My previous question might not be very clear. This is what I would like to find out:
> 
> Suppose `expr1` and `expr2` are some JavaScript expressions. Can you explain how the expression `expr1 || expr2` is evaluated in JavaScript?

In Javascript in evaluating expr1 || expr2...  expr1 is first evaluated if, expr1 happens to be false, expr2 gets evaluated. 
In our case hours % 12 || 12,  if hours %12 = 0 it might be midnight or miday so 12 is returned, other wise.. the remainder is returned. as 1.. 2... 3 ..

PR #813, opened 2025-10-27
West Midlands | Sept-ITP-25 | Mustaf Asani | Sprint 3 | Coursework/sprint 3 practice tdd

asaniDev:
## Learners, PR Template

Self checklist

## Changelist

Created tests to check the expected output of functions before they are created.

### DIFF
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.test.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.test.js
@@ -33,10 +33,10 @@ test("should append 'rd' to numbers with 3 at the end except for those ending wi
 });
 
 test("should append 'th' to all other numbers which do not end in 1,2,3,11,12 or 13", () => {
-  expect(getOrdinalNumber(0)).toEqual("0th");
-  expect(getOrdinalNumber(11)).toEqual("11th");
-  expect(getOrdinalNumber(212)).toEqual("212th");
-  expect(getOrdinalNumber(113)).toEqual("113th");
+  expect(getOrdinalNumber(4)).toEqual("4th");
+  expect(getOrdinalNumber(9)).toEqual("9th");
+  expect(getOrdinalNumber(211)).toEqual("211th");
+  expect(getOrdinalNumber(1013)).toEqual("1013th");
   expect(getOrdinalNumber(17)).toEqual("17th");
 });
 
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.js
@@ -3,7 +3,7 @@ function getOrdinalNumber(num) {
     const lastDigit = num % 10;
 
     if (lastDigit === 1) {
-      if (num === 11) {
+      if (num % 100 === 11) {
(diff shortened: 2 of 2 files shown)


Sprint-3/2-practice-tdd/repeat.js
```diff
@@ -1,5 +1,18 @@
...
+  } else if (count === 1) {
+    return (newStr += str);
+  } else if (count > 1) {
+    for (let i = 0; i < count; i++) {
+      newStr += str;
+    }
+    return newStr;
+  } else {
+    return "Error invalid count used, please use integers from 0 upwards.";
+  }
```
cjyuan: How would the caller distinguish the result of the following two function calls?
1. `repeat("Error invalid count used, please use integers from 0 upwards.", 1)`
2. `repeat("", -1)`

wahibkamran: If the input is invalid (negative integers), how can we tell the user that there's an issue instead of returning a response?

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -8,18 +8,34 @@ const getOrdinalNumber = require("./get-ordinal-number");
...
 });
 
-test("should return any number between 4 and 20 with a suffix of 'th' at end of number", () => {
+test("should append 'th' to all other numbers which do not end in 1,2,3,11,12 or 13", () => {
   expect(getOrdinalNumber(10)).toEqual("10th");
+  expect(getOrdinalNumber(11)).toEqual("11th");
+  expect(getOrdinalNumber(212)).toEqual("212th");
+  expect(getOrdinalNumber(113)).toEqual("113th");
+  expect(getOrdinalNumber(17)).toEqual("17th");
 });
```
cjyuan: These tests are quite comprehensive. If you use this test script to test your implementation, you can discover some bug in your code.

Sprint-3/2-practice-tdd/repeat.js
```diff
@@ -4,15 +4,19 @@ function repeat(str, count) {
...
     }
     return newStr;
   } else {
-    return "Error invalid count used, please use integers from 0 upwards.";
+    throw new Error(
+      "Error invalid count used, please use integers from 0 upwards."
+    );
+    //return "Error invalid count used, please use integers from 0 upwards.";
   }
+  return true;
```
cjyuan: - What's the purpose of the statement on line 19? When do you expect it to be executed?

- You can also consider structuring your code in the following way:
```
  if (code < 0)
    throw ...

  if (code == 0) return "";

  if (code == 1) return ...;

  let newStr = "";
  ...
  return newStr;
```

or (if you don't mind trading performance for simpler code)

```
  if (count < 0)
    throw ...;

  // this code will produce the correct result for any count >= 0 anyway.
  let newStr = "";
  ...
  return newstr;

```

cjyuan (commented):
There is a bug in the `getOrdinalNumber()`. Other changes look good.

cjyuan (commented):
Good job. 

That's why a comprehensive test is important.

PR #818, opened 2025-10-28
London | 25-ITP-Sep | Shaghayegh Shirinfar | Sprint 3 | Practice-tdd

shaghayeghfar:
## Learners, PR Template

Self checklist


jennethydyrova (commented):
Hi @shaghayeghfar ! Good start on this PR, I left some comments to address

shaghayeghfar:
@jennethydyrova 

Dear Jennet,

Thank you very much for reviewing my code. I have made the changes you suggested.

I really appreciate your help, time, and effort.

Best regards,

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,21 @@
 function countChar(stringOfCharacters, findCharacter) {
-  return 5
+  if (
+    typeof stringOfCharacters !== "string" ||
```
jennethydyrova: If either or both arguments are not strings, does it really make sense to return 0? Think of it this way, you call the `countChar(2, 3)`, what would you expect your function to return? I would suggest you to check this link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,21 @@
...
+    typeof findCharacter !== "string"
+  ) {
+    return 0;
+  }
+  if (findCharacter.length !== 1) {
```
jennethydyrova: This is a good check! As an extension, what if `stringOfCharacters.length` is 0 and `findCharacter.length` is 1? I see you now return just 0 but would it perhaps make more sense to throw an error because there is nothing to search character in?

Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -17,8 +17,91 @@ test("should count multiple occurrences of a character", () => {
...
+  const count = countChar(str, char);
+  expect(count).toEqual(0);
+});
+
+// Scenario: Find character is longer than the input string
+// Given a string shorter than the findCharacter (e.g., 'a' vs 'abc'),
+// When the function is called,
+// Then it should return 0 because a multi-character string cannot match a single character position.
+
+test("should return 0 when findCharacter is longer than the input string", () => {
```
jennethydyrova: I think it's a bit incorrect. This is related to my earlier comment for the `countChar` function. If str length is shorter than char length, the function should return an error. Returning 0 should be reserved only for the case when character is not found. Otherwise, you function logic becomes convoluted, you wouldn't know if it returned 0 because character is not found or because arguments were invalid. Do you see the problem with it?

shaghayeghfar:
Dear Jennet,

Thank you for your feedback! I’ve made all the changes you requested. However, I’m not entirely sure what you meant regarding removing the unrelated files. Could you please provide a bit more guidance on that?

Thanks so much for your help!

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,42 @@
...
+  ) {
+    return "Values Can not be Numbers";
+  }
+
+  // stringOfCharacters should be longer than findCharacter.
+  if (
+    stringOfCharacters.length !== 0 &&
+    findCharacter.length > stringOfCharacters.length
+  ) {
+    return "the stringOfCharacters MUST be lonegr than findingcharacter";
```
jennethydyrova: typo in `longer` and `findingcharacter`

jennethydyrova (approved):
Hi @shaghayeghfar ! I am marking this PR as complete but consider fixing typos

PR #833, opened 2025-11-01
London | 25-ITP-Sep | KME Blair | Sprint 2 | Coursework

kohanman:
## Learners, PR Template

Self checklist

## Changelist

done all sprint 2 coursework for Structuring and testing data Module

## Questions

n/a

### DIFF
--- a/Sprint-3/2-practice-tdd/README.md
+++ b/Sprint-3/2-practice-tdd/README.md
@@ -9,5 +9,5 @@ Write the tests _before_ the code that will make them pass.
 Recommended order:
 
 1. `count.test.js`
-1. `repeat.test.js`
+1. `repeat-str.test.js`
 1. `get-ordinal-number.test.js`
--- a/Sprint-3/2-practice-tdd/repeat-str.js
+++ b/Sprint-3/2-practice-tdd/repeat-str.js
@@ -1,0 +1,5 @@
+function repeatStr() {
+  return "hellohellohello";
+}
+
+module.exports = repeatStr;
--- a/Sprint-3/2-practice-tdd/repeat-str.test.js
+++ b/Sprint-3/2-practice-tdd/repeat-str.test.js
@@ -1,32 +1,32 @@
-// Implement a function repeat
-const repeat = require("./repeat");
+// Implement a function repeatStr
+const repeatStr = require("./repeat-str");
 // Given a target string str and a positive integer count,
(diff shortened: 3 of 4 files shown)


Sprint-2/1-key-errors/0.js
```diff
@@ -5,9 +5,17 @@
 // interpret the error message and figure out why an error is occurring
 
 function capitalise(str) {
-  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+  var str = `${str[0].toUpperCase()}${str.slice(1)}`;
```
wheresdiasd: This works fine. I suggest trying to reassign the parameter or change the variable name. What option do you think would be more intuitive for who will read the code in the future ?

cjyuan: `str` (the parameter) works like a local variable.
Redeclaring `str` using `var` has no effect on `str` (Note: Redeclaring `str` using `let` or `const` is prohibited)

So 
```
var str = `${str[0].toUpperCase()}${str.slice(1)}`;
```
is identical to
```
str = `${str[0].toUpperCase()}${str.slice(1)}`;
```

In general, avoid using `var` because it can make our code behave in unexpected ways (if not used properly); `let` and `const` are safer.

Sprint-2/3-mandatory-implement/3-to-pounds.js
```diff
@@ -4,3 +4,22 @@
...
+  const pounds = paddedPenceNumberString.substring(
+    0,
+    paddedPenceNumberString.length - 2
+  );
+
+  const pence = paddedPenceNumberString
+    .substring(paddedPenceNumberString.length - 2)
+    .padEnd(2, "0");
+
+  console.log(`£${pounds}.${pence}`);
```
cjyuan: The function is expected to return a string in pound format. Outputting the string to console is not the same as returning the string.

Sprint-2/4-mandatory-interpret/time-format.js
```diff
@@ -18,17 +18,22 @@ function formatTimeDisplay(seconds) {
...
 
 // Call formatTimeDisplay with an input of 61, now answer the following:
 
 // b) What is the value assigned to num when pad is called for the first time?
 // =============> write your answer here
+//0
 
 // c) What is the return value of pad is called for the first time?
 // =============> write your answer here
+//00
```

PR #838, opened 2025-11-02
London | 25-ITP-Sep | Payman Issa Baiglu | Sprint 2 | Coursework/sprint 2

PaymanIB:

in this PR: Sprint 2 Coursework completed. errors fixed and questions answered


Sprint-2/1-key-errors/0.js
```diff
@@ -1,13 +1,19 @@
...
-
+// it gives us an error.
 // call the function capitalise with a string input
 // interpret the error message and figure out why an error is occurring
 
-function capitalise(str) {
-  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
-  return str;
-}
+// function capitalise(str) {
```
jennethydyrova: Let's remove all commented code/console logs/any scratch work that is not relevant to the pull request to keep it clean.

Sprint-2/1-key-errors/1.js
```diff
@@ -1,20 +1,32 @@
 // Predict and explain first...
-
-// Why will an error occur when this program runs?
+// error on line 11
+// error on line 17
+// Why will an error occur when this program runs? yes
 // =============> write your prediction here
+// decimalNumber is declared before and can not be declared again.
+// console.log(decimalNumber) should be console.log(convertToPercentage)
```
jennethydyrova: This is correct but why is that a case? Why we can't do `console.log(decimalNumber)`?

Sprint-2/1-key-errors/1.js
```diff
@@ -1,20 +1,32 @@
...
 
 // Finally, correct the code to fix the problem
 // =============> write your new code here
+function convertToPercentage(decimalNumber) {
+  //const decimalNumber = 0.5;
+  const percentage = `${decimalNumber * 100}%`;
+
+  return percentage;
```
jennethydyrova: Can you think of a way to simplify this?

Sprint-2/2-mandatory-debug/1.js
```diff
@@ -1,13 +1,19 @@
...
-function sum(a, b) {
-  return;
-  a + b;
-}
-
-console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
+// console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
 
 // =============> write your explanation here
+//the return statement is not returning any value. a+b declared after return.
```
jennethydyrova: You are right about the return but be careful with second part, `a + b` is not not a declaration, it's an expression.

Sprint-2/4-mandatory-interpret/time-format.js
```diff
@@ -4,31 +4,35 @@ function pad(num) {
 
 function formatTimeDisplay(seconds) {
   const remainingSeconds = seconds % 60;
-  const totalMinutes = (seconds - remainingSeconds) / 60;
-  const remainingMinutes = totalMinutes % 60;
+  const totalMinutes = (seconds - remainingSeconds) / 60; //1
+  const remainingMinutes = totalMinutes % 60; //1
```
jennethydyrova: Do you need these comments here?

Sprint-2/1-key-errors/0.js
```diff
@@ -1,13 +1,19 @@
...
+// return str;
+// }
 
-// =============> write your explanation here
+// =============> write your explanation here :
+// as the variable in side the function has the same name as the function variable.
 // =============> write your new code here
+function capitalise(str) {
+  let str1 = `${str[0].toUpperCase()}${str.slice(1)}`;
+  return str1;
```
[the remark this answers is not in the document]
PaymanIB: Thanks for pointing that out. It’s not needed.
Since str1 never changes after it’s assigned, const is better than let. But as you mentioned, the simpler approach is to just return the value directly.

Sprint-2/2-mandatory-debug/2.js
```diff
@@ -2,23 +2,35 @@
 
 // Predict the output of the following code:
 // =============> Write your prediction here
+// the result will be 3 times
```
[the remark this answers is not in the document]
PaymanIB: As const num = 103, return was always 3.

Sprint-2/1-key-errors/1.js
PaymanIB: Good point. I fixed it. thanks.

PaymanIB:
Hi @jennethydyrova, I totally forgot that part. Thanks for reminding me. that file is deleted. Would you please check again.

Sprint-2/1-key-errors/1.js
PaymanIB: sorry, I answered it but it stays as pending.
decimalNumber is defined inside the function and if we console.log(decimalNumber); we will get a ReferenceError.

wahibkamran (approved):

Sprint-2/1-key-errors/1.js
wahibkamran: looks good overall @PaymanIB, I'd just suggest cleaning up code by removing unused comments.

PaymanIB:
@wahibkamran Thank you for reviewing my PR. If it is OK. can you please change the label to complete as well.

PaymanIB:
Great! Thank you!

PR #840, opened 2025-11-03
West Midlands | 25-ITP-Sept | Ali Naru | Sprint 2 | Coursework/sprint-2

MohammedNaru:
## Sprint 2 Coursework

Self checklist

## Changelist

Completed all coursework tasks on Sprint 2

## Questions

No questions currently

### DIFF
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -1,41 +1,19 @@
-// Below are the steps for how BMI is calculated
-
-// The BMI calculation divides an adult's weight in kilograms (kg) by their height in metres (m) squared.
-
-// For example, if you weigh 70kg (around 11 stone) and are 1.73m (around 5 feet 8 inches) tall, you work out your BMI by:
-
-// squaring your height: 1.73 x 1.73 = 2.99
-// dividing 70 by 2.99 = 23.41
-// Your result will be displayed to 1 decimal place, for example 23.4.
-
-// You will need to implement a function that calculates the BMI of someone based off their weight and height
-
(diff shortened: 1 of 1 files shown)


Sprint-2/1-key-errors/1.js
```diff
@@ -16,5 +19,11 @@ console.log(decimalNumber);
...
 
-// Finally, correct the code to fix the problem
+// The code will throw a SyntaxError because the variable 'decimalNumber' is being declared twice within the same scope.
 // =============> write your new code here
+
+function convertToPercentage(decimalNumber) {
+  decimalNumber = 0.5;
+  const percentage = `${decimalNumber * 100}%`;
+
+  return percentage;
```
cjyuan: - This is an incomplete implementation of the function.
- This function will always return "50%", which is not very useful. Can you improve its reusability so that it can be used to convert any number to an equivalent string in percentage format?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -16,4 +16,27 @@
...
+// To fix this, we need to perform the BMI calculation: 
+// BMI = weight / (height * height)
+// Then, we round the result to one decimal place using toFixed(1).
+
+// Finally, correct the code to fix the problem
+//  =============> write your new code here
+
+function calculateBMI(weight, height) {
+  const bmi = weight / (height * height);
+  return bmi.toFixed(1);
```

```javascript
  
```

MohammedNaru: All updated now please can you review again.

cjyuan: Can you also answer my questions in this thread?

PR #858, opened 2025-11-08
Refactor/implement rewrite coverage

wheresdiasd:
The pull requests consist in adding coverage for the sprint-3 coursework "IMPLEMENT AND REWRITE".

When users open a PR with the pattern 'coursework/sprint-3-implement-and-rewrite from their forks, it should run the expected tests for the coursework scenarios in a way the Reviewer have a smooth perspective of the written test results.

In addition, introduces a shell script that compares the implementation with the test cases to see if their match in equality of test cases.

The summary looks like below

Screenshot 2025-11-20 at 07 10 38

And that is how the checks tab look like where we can see if the tests are actually passing or not.
Screenshot 2025-11-27 at 10 47 48

### DIFF
--- a/.github/workflows/test-coverage-1-implement-rewrite-tests.yml
+++ b/.github/workflows/test-coverage-1-implement-rewrite-tests.yml
@@ -1,4 +1,4 @@
-name: Test Coverage Report
+name: Test Coverage Report - Implement and rewrite 
 
 on:
   pull_request:


.github/workflows/test-coverage-1-implement-rewrite-tests.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run Inline Assertion Tests & Verify Parity
+        run: node scripts/run-inline-tests.js
+        continue-on-error: true
+
+      - name: Run Jest Tests with Coverage
+        run: npx jest --config=jest.config.jest-tests.js --coverage --ci
```
LonMcGregor: the npm test command is defined in the package.json.Is there a reason you use npx to run the same command here?

.github/workflows/test-coverage-3-stretch.yml
LonMcGregor: This file looks the same as for coverage-2. Could they be combined?

Sprint-3/1-implement-and-rewrite-tests/README.md
LonMcGregor: Test coverage runs for task 2 and task 3 in this sprint. Either the instructions should be replicated in those readmes, or a more general set of instructions should be given in the main sprint3 readme rather than here.

PR #869, opened 2025-11-10
Manchester| 25-ITP-Sep | Fithi Teklom| Sprint 2 | Coursework 

Fithi-Teklom:
Self checklist

### DIFF
--- a/Sprint-2/3-mandatory-implement/3-to-pounds.js
+++ b/Sprint-2/3-mandatory-implement/3-to-pounds.js
@@ -20,7 +20,6 @@ function toPounds(penceString) {
 
   const pence = paddedPenceNumberString
     .substring(paddedPenceNumberString.length - 2)
-    .padEnd(2, "0");
 
   return `£${pounds}.${pence}`;
 }


Sprint-2/3-mandatory-implement/3-to-pounds.js
WeiTsungCheng: Remove .padEnd(2, "0") 
Will it affect the code?

Sprint-2/5-stretch-extend/format-time.js
WeiTsungCheng: Only Test One passed the test, indicating that the function formatAs12HourClock(time) needs to be changed.

Sprint-2/3-mandatory-implement/3-to-pounds.js
Fithi-Teklom: And the important thing is that variables don't start with number.

Sprint-1/1-key-exercises/3-paths.js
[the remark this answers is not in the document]
WeiTsungCheng: Yes, so it can be changed to return a blank value. This is because no filename extension was provided.

Sprint-2/5-stretch-extend/format-time.js
WeiTsungCheng: Sorry, I didn't notice your correct version was in the commented-out area. Your new version does indeed pass the test.

Sprint-2/3-mandatory-implement/3-to-pounds.js
Fithi-Teklom: Ahh, removing .padEnd(2, "0") will not change the output for valid inputs, because paddedPenceNumberString is already guaranteed to be at least 3 characters long from .padStart(3, "0").

PR #872, opened 2025-11-11
London | 25-ITP-SEP | Mariia Serhiienko| Sprint 3 | Coursework/sprint 3 practice tdd

Konvaly:
Self checklist

This PR covers work in the Test-First Development section of the sprint project:

Test-First Approach: 

Wrote Jest tests before implementing the functions, practicing the test-driven development (TDD) process.

Jest Tests: Created and completed test files for:
 -  count.test.js
 -  repeat-str.test.js
 -  get-ordinal-number.test.js

Each test file follows the provided acceptance criteria and uses Jest’s  syntax for clarity.

Functions were then implemented to make all tests pass.

All tests run successfully with the VS Code Test Runner or by using npm test.

### DIFF
--- a/Sprint-3/2-practice-tdd/repeat.js
+++ b/Sprint-3/2-practice-tdd/repeat.js
@@ -2,6 +2,9 @@ function repeat(str, count) {
   if (count === 0) {
     return "";
   }
+  if (count < 0) {
+    return "Error! Negative counts are not valid!";
+  }
 
   let repeatedStrCountTimes = "";
 
@@ -10,7 +13,7 @@ function repeat(str, count) {
     repeatedStrCountTimes += str;
     i++;
   }
-  console.log(repeatedStrCountTimes);
+
   return repeatedStrCountTimes;
 }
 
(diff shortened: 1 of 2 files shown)


Sprint-2/3-mandatory-implement/3-to-pounds.js
```diff
@@ -4,3 +4,26 @@
...
+    paddedPenceNumberString.length - 2
+  );
+  const pence = paddedPenceNumberString
+    .substring(paddedPenceNumberString.length - 2)
+    .padEnd(2, "0");
+
+  return `£${pounds}.${pence}`;
+}
+
+console.log(toPounds("123p")); // £1.23
```
wheresdiasd: I'm curious why you used a string with "p" at the end of the input. 

That is not usual, and it is hard to determine how we could guarantee the input would always be in that format.

I come from the point that a string can be anything; we could, for example, call the function with a string without "p" at the end.

I'd recommend to research what data types would be recommended to work when working with currencies ?

PR #879, opened 2025-11-14
Manchester | 25-ITP-Sep | Khalidbih | Sprint 3 | coursework/sprint-3-implement-and-rewrite

khalidbih:
Self checklist

Changelist: 

 I have completed the coursework/sprint-3-implement-and-rewrite.

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
@@ -12,8 +12,8 @@ test("should return false for Improper fraction", () => {
 });
 
 // Case 3: Identify Negative Fractions:
-test("should return false for Negative fraction", () => {
-  expect(isProperFraction(-4, 7)).toEqual(false);
+test("should return true for Negative fraction", () => {
+  expect(isProperFraction(-4, 7)).toEqual(true);
 });
 
 // Case 4: Identify Equal Numerator and Denominator:


Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -8,9 +8,20 @@
 // write one test at a time, and make it pass, build your solution up methodically
 // just make one change at a time -- don't rush -- programmers are deep and careful thinkers
 function getCardValue(card) {
+  const chars = [...card];
+  const rank = chars.slice(0, -1).join('');
```
cjyuan: String's `.slice()` method is probably more efficient.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -8,9 +8,20 @@
...
+
   if (rank === "A") {
     return 11;
   }
+  const parsed = Number(rank);
+  if (!Number.isNaN(parsed) && parsed >= 2 && parsed <= 9) {
+    return parsed;
+  }
```
Do you want to recognize these string values as valid ranks?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
khalidbih: Thanks for the feedback, I understand now that string.slice(0, -1) is simpler and more efficient. I’ll be more careful next time and try to avoid doing extra steps when a shorter method works.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
khalidbih: I understand that Number(rank) accepts many numeric formats like '02', '2e0', '2.0', etc. For this exercise, I only want plain card ranks ('2'–'9'), not those other formats. Thanks for the feedback.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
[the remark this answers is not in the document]
khalidbih: No, because it will fail the negative fraction test (Case 3).
I run the code, this is the output:  FAIL  ../1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
  ✓ should return true for a proper fraction (4 ms)
  ✓ should return false for Improper fraction (4 ms)
  ✕ should return false for Negative fraction (2 ms)
  ✓ should return false for Equal numerator and denominator (1 ms)

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -3,11 +3,46 @@
...
 // Case 2: Handle Number Cards (2-10):
+test("should return 2 for 2 of hearts", () => {
+  const twoOfHearts = getCardValue("2♥");
+  expect(twoOfHearts).toEqual(2)
+})
+
+test("should return 10 for 10 of Diamonds", () => {
+  const tenOfDiamonds = getCardValue("10♦");
+  expect(tenOfDiamonds).toEqual(10)
+});
```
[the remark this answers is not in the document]
khalidbih: Thank you for the feedback. I understood. I’ll group cards by category (numbers, face cards, Ace) to cover all cases without testing each individually.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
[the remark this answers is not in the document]
cjyuan: One test fails. So how would you fix the issue?

PR #911, opened 2026-01-25
London | 26-ITP-Jan | Abdul Moiz | Sprint 2 | Sprint 2

A-Moiz:
## Learners, PR Template

Self checklist

## Changelist

Completed all the exercises in Sprint 2 directory

## Questions

N/A

### DIFF
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -4,10 +4,21 @@
 // call the function capitalise with a string input
 // interpret the error message and figure out why an error is occurring
 
-function capitalise(str) {
-  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
-  return str;
-}
+// function capitalise(str) {
+//   let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+//   return str;
+// }
+
+// console.log(capitalise("lowercase"));
 
 // =============> write your explanation here
+// The error states str has already been declared. This is because str is the parameter and within the
+// function it is trying to declare a new variable with the same name using let.
+
 // =============> write your new code here
+function capitalise(str) {
+  let newStr = `${str[0].toUpperCase()}${str.slice(1)}`;
+  return newStr;
+}
+
+console.log(capitalise("lowercase"));
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -17,8 +17,7 @@
 
 // =============> write your new code here
 function capitalise(str) {
-  let newStr = `${str[0].toUpperCase()}${str.slice(1)}`;
-  return newStr;
+  return `${str[0].toUpperCase()}${str.slice(1)}`;
 }
 
 console.log(capitalise("lowercase"));
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -15,5 +15,11 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
+  let heightSquared = height * height;
+  let bmi = weight / heightSquared;
+  return bmi.toFixed(1);
 }
+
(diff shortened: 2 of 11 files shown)


Sprint-2/1-key-errors/0.js
```diff
@@ -4,10 +4,21 @@
...
+
+// console.log(capitalise("lowercase"));
 
 // =============> write your explanation here
+// The error states str has already been declared. This is because str is the parameter and within the
+// function it is trying to declare a new variable with the same name using let.
+
 // =============> write your new code here
+function capitalise(str) {
+  let newStr = `${str[0].toUpperCase()}${str.slice(1)}`;
```
cjyuan: For a variable not needed to be reassigned a value, a better practice is to use `const` instead of `let`.

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,11 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+  let heightSquared = height * height;
+  let bmi = weight / heightSquared;
+  return bmi.toFixed(1);
```

```javascript
  
```

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -23,3 +43,20 @@ console.assert(
...
+// Morning times
+console.assert(formatAs12HourClock("01:00") === "01:00 am");
+console.assert(formatAs12HourClock("11:59") === "11:59 am");
+
+// Noon
+console.assert(formatAs12HourClock("12:00") === "12:00 pm");
+console.assert(formatAs12HourClock("12:30") === "12:30 pm");
+
+// Afternoon / evening
+console.assert(formatAs12HourClock("13:00") === "1:00 pm");
```
cjyuan, line 56: Output on lines 56 and 48 have slightly different format.
Can you modify the function to return a string in a **consistent** format?

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -2,11 +2,31 @@
...
+//   if (hours > 12) {
+//     return `${hours - 12}:00 pm`;
+//   }
+//   return `${time} am`;
+// }
+
+// Updated function:
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
+  const minutes = time.slice(3, 5);
```
cjyuan: Can also consider `time.slice(-2)` (a bit more expressive).

cjyuan (commented):
Code is generally good.

Sprint-2/3-mandatory-implement/1-bmi.js
A-Moiz: It was a string because the toFixed() converts it to String. I added in parseFloat so it is converted back into a number.

Sprint-2/5-stretch-extend/format-time.js
A-Moiz: Hi I'm unsure of the question you're asking. Line 56 tests if 13:00 is same as 1:00pm and line 48 tests if 01:00 is same as 01:00 am. All the assertions are passing currently.

cjyuan: If "01:00" is converted to "01:00 am", it is reasonable for the caller to expect "13:00" to be converted to "01:00 pm".

What could go wrong? Here are two examples,
- When the strings are all centered within a table column on a webpage, "01:00 pm" and "1:00 pm" would not align as nicely.

<div align=center>01:00 am</div>
<div align=center>1:00 pm</div>
<div align=center>12:00 am</div>
<div align=center>01:00 pm</div>

- When the times are compared in the program, `"1:00 pm" < "11:00 pm"` and `01:00 pm" < "11:00 pm"` produce different results.

Did you choose the format "1:00 pm" by design (before you implement the function), or did you set the expected value in your tests because you knew that's what your function will return?

cjyuan:
Changes look good.

Stretch exercise is optional so I will mark this PR as complete.

illicitonion:

PR #914, opened 2026-01-27
Configure npm to not create lockfile

textbook:
Rather than ignore it or have huge commit diffs, [tell npm](https://docs.npmjs.com/cli/v11/using-npm/config#package-lock) not to use lock files.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false


illicitonion (approved):
Makes sense! Thanks :)

PR #915, opened 2026-01-27
London | 26-ITP-Jan | Abdul Moiz | Sprint 3 | implement-and-rewrite

A-Moiz:
## Learners, PR Template

Self checklist

## Changelist

Completed all the exercises in implement-and-rewrite directory

## Questions

N/A


Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -11,7 +11,13 @@
...
 
 function isProperFraction(numerator, denominator) {
-  // TODO: Implement this function
+  if (denominator === 0) {
+    return false;
+  } else if (Math.abs(numerator) < Math.abs(denominator)) {
+    return true;
+  } else {
+    return false;
+  }
```
cjyuan: This works. 
Optional challenge: Simplify the code.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -22,7 +22,16 @@
 // execute the code to ensure all tests pass.
 
 function getCardValue(card) {
-  // TODO: Implement this function
+  const rank = card.slice(0, card.length - 1);
```
cjyuan: The requirements has changed a bit. 
To ensure `card` is valid, you should also check if the last character is a valid suit character.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -22,7 +22,16 @@
...
+  if (rank === "A") {
+    return 11;
+  } else if (Number(rank) >= 2 && Number(rank) < 10) {
+    return Number(rank);
```
Do you want to recognize these string values as valid ranks? If not, you will need to enforce stricter check.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -8,3 +8,19 @@ const isProperFraction = require("../implement/2-is-proper-fraction");
...
 });
+
+test(`should return true for proper fraction with positive numerator and denominator`, () => {
+  expect(isProperFraction(1, 2)).toEqual(true);
+});
+
+test(`should return false for improper fraction with positive numerator and denominator`, () => {
+  expect(isProperFraction(3, 2)).toEqual(false);
+});
+
```
cjyuan, line 19: What about negative numbers and different combinations of  positive/negative/zero?

Think also what edge case to check?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -22,7 +22,16 @@
...
+  } else if (rank === "10" || rank === "J" || rank === "Q" || rank === "K") {
+    return 10;
+  } else {
+    return "Invalid card rank";
+  }
```
cjyuan: The new requirement is:

>// When the card string is invalid (not following the above format), the function should
>// throw an error.

Throwing an error is not the same as returning an error message.
Could you look up "How to throw an error in JS" and update your code accordingly?

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -14,7 +14,27 @@ test(`Should return 11 when given an ace card`, () => {
...
+  expect(getCardValue("K♥")).toEqual(10);
+});
+
+// Case 4: Invalid Cards
+test(`Should return Invalid card rank for invalid cards`, () => {
+  expect(getCardValue("1♠")).toEqual("Invalid card rank");
+  expect(getCardValue("11♥")).toEqual("Invalid card rank");
+  expect(getCardValue("Z♦")).toEqual("Invalid card rank");
+});
+
```
cjyuan, line 46: - Could also include invalid cases that involves invalid suit character.

- To test if a function can throw an error as expected, we could use `.toThrow()`. You can find out more about how to use `.toThrow()` here: https://jestjs.io/docs/expect#tothrowerror (Note: Pay close attention to the syntax of the example)

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
A-Moiz: I added an if statement beforehand to check typeof for the variable. If it fails it will throw an error.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
A-Moiz: Added extra tests for negative numerators/denominators and all are passing.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
A-Moiz: Modified function and test case to throw error.

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -11,7 +11,7 @@
 // execute the code to ensure all tests pass.
 
 function isProperFraction(numerator, denominator) {
-  // TODO: Implement this function
+  return Math.abs(numerator) < Math.abs(denominator) ? true : false;
```
cjyuan: When an expression evaluates to a Boolean value, we can also just return the value of the expression:
`return Math.abs(numerator) < Math.abs(denominator);`

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -22,7 +22,23 @@
...
+    throw new Error("Invalid card suit");
+  }
+
+  if (rank === "A") return 11;
+  if (rank === "10" || rank === "J" || rank === "Q" || rank === "K") return 10;
+
+  const rankNum = Number(rank);
+  if (!Number.isNaN(rankNum) && rankNum >= 2 && rankNum < 10) {
+    return rankNum;
+  }
```
cjyuan, line 45: > I added an if statement beforehand to check typeof for the variable. If it fails it will throw an error.

I don't see any statement involving `typeof`. 

Currently, a card value like ""0x02♠" or "3.1416♠" can still pass the check on line 37.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -8,3 +8,35 @@ const isProperFraction = require("../implement/2-is-proper-fraction");
...
+  expect(isProperFraction(5, -2)).toEqual(false);
+});
+
+test(`should return false for negative numerator`, () => {
+  expect(isProperFraction(-5, 2)).toEqual(false);
+});
+
+test(`should return false for negative numerator`, () => {
+  expect(isProperFraction(-5, 2)).toEqual(false);
+});
```
cjyuan: Same test sample.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
A-Moiz: Apologies, I got it mixed up with a different task. It is resolved now.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -34,7 +34,13 @@ function getCardValue(card) {
   if (rank === "10" || rank === "J" || rank === "Q" || rank === "K") return 10;
 
   const rankNum = Number(rank);
-  if (!Number.isNaN(rankNum) && rankNum >= 2 && rankNum < 10) {
+  if (
+    !Number.isNaN(rankNum) &&
+    Number.isInteger(rankNum) &&
+    rankNum >= 2 &&
+    rankNum < 10 &&
+    rank === rankNum.toString()
```
cjyuan, line 42: This works, but an approach similar to that at line 29 is probably cleaner.

cjyuan (commented):
Changes look good. Well done.

illicitonion:

PR #925, opened 2026-02-01
London | 26-ITP-January | Karla Grajales | Sprint 2 | Coursework/sprint 2

Grajales-K:
## Learners, PR Template

Self checklist

## Changelist

In this project, I implemented and refactored functions to handle data conversion (Time and Currency). I also documented the execution flow of nested functions and how parameters are processed when functions are invoked sequentially.

## Questions

What is the most effective learning path for mastering functions: should I focus on core logic like loops first, or move directly to built-in array methods (like .forEach or .map)?


Sprint-2/1-key-errors/1.js
```diff
@@ -1,20 +1,41 @@
 // Predict and explain first...
 
+const { t } = require("tar");
```
LonMcGregor: Why are you requiring this "tar"?

Sprint-2/1-key-errors/1.js
```diff
@@ -1,20 +1,41 @@
...
+// parameter dynamically and invoke the function correctly.
+
 // Finally, correct the code to fix the problem
 // =============> write your new code here
+
+
+
+function convertToPercentage(num) {
+
+  const percentage = `${num * 100}%`;
```
LonMcGregor: Could you simplify this to one line?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,19 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+  // return the BMI of someone based off their weight and height
+  const squareHeight = height * height;
+  const operation = weight / squareHeight;
+  console.log(operation);
```
LonMcGregor: Good solution - do you need to keep this console.log if the solution is finished?

Sprint-2/3-mandatory-implement/3-to-pounds.js
```diff
@@ -4,3 +4,78 @@
...
+// Empty String (''): Replaces the "bad" characters with nothing (deleting them).
+// Example: "£1.50p" -> "150"
+// const onlyNumbers = str.replace(/\D/g, '');
+
+// =============================== Reusable Code ========================
+
+
+function toPounds(str){
+
+    const onlyNumbers = str.replace(/[^\d]/g, '');
```
LonMcGregor, line 60: This is a really interesting approach!

LonMcGregor (commented):
Well done on this task so far, you have given good explanations to the debugging tasks.

I have left a couple of comments where you could improve further.

For learning, I would suggest understanding the basics before starting on using built-in functions. Something that might be interesting is seeing if you can write your own forEach or map function, and once you've got that down, then just using the built in functions.

Sprint-2/1-key-errors/1.js
Grajales-K: Hi @LonMcGregor, thanks for raising this.
The require("tar") line was already there when I cloned the repository. Initially, I assumed it was a dependency related to how GitHub or the project handles files, but after investigating, I found it wasn't being used anywhere. It might have been included intentionally as a 'red herring' (a distraction) to test if I could identify unnecessary code, or it could be a leftover auto-import error. I've removed it because it's not needed for the function to work.

Sprint-2/1-key-errors/1.js
Grajales-K: Hi @LonMcGregor,

I did one returning the result, then made an arrow function. thank you

Sprint-2/3-mandatory-implement/1-bmi.js
Grajales-K: I removed it. I usually have them because they help me to read the code when I come back to recap what I did.
Thank you.

Sprint-2/3-mandatory-implement/3-to-pounds.js
Grajales-K: Thank you
Yes, it helped me with one Codewars and I'm really happy to use this instead of looping all characters.

Grajales-K:
> Well done on this task so far, you have given good explanations to the debugging tasks.
> 
> I have left a couple of comments where you could improve further.
> 
> For learning, I would suggest understanding the basics before starting on using built-in functions. Something that might be interesting is seeing if you can write your own forEach or map function, and once you've got that down, then just using the built in functions.


Thank you very much, @LonMcGregor, for your review and your suggestions. I really appreciate your time, and yes, I want to learn and am happy to read your suggestion.

I'm actually trying to use for loops in Codewars right now to learn how the data really moves. Built-in methods are easy, but I want to understand what's happening 'under the hood' before I use them all the time.

LonMcGregor (commented):
Great work!

I'm honestly not sure what the `require tar`, but good to remove it if not needed.

You don't always need to make things a one-liner, but I mention it as a way to get some practice. Typically i find if the function is very short it can make things more readable.

This task is done now.

illicitonion:

PR #927, opened 2026-02-03
London | 26-ITP-January | Eugenie Ahangama | Sprint 1 | coursework/sprint-1 

Eugenie-A:
## Learners, PR Template

Self checklist

## Changelist

In this project, I completed both the coding exercises and practical exploration of browser based functions using the Chome DevTools console. I tested how different JavaScript features behave, documented my findings and wrote clear explanations in Markdown.

I referred to MDN documentation throughout this project:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charAt
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder

## Questions

Are there habits I should start building now to avoid problems later?

### DIFF
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -4,3 +4,6 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+
+// The = operator assigns a new value to 'count'. It takes the current value of 'count'.
+// adds 1 to it, and then stores that new value back into the 'count' variable.
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -1,9 +1,7 @@
 let count = 0;
 
-count = count + 1;
+count = count + 1;  // increment: adds 1 and assigns the new value back
 
-// Line 1 is a variable declaration, creating the count variable with an initial value of 0
-// Describe what line 3 is doing, in particular focus on what = is doing
-
-// The = operator assigns a new value to 'count'. It takes the current value of 'count'.
-// adds 1 to it, and then stores that new value back into the 'count' variable.
+// MDN Increment operator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Increment
+// MDN Assignment operators: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#Assignment_operators
+// Alternatives: count += 1 or count++
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -1,9 +1,20 @@
 const minimum = 1;
 const maximum = 100;
 
-const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
+const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum; // Generates a random whole number between minimum and maximum
 
 // In this exercise, you will need to work out what num represents?
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+// Generate a random decimal number between 0 and 1
+const decimal = Math.random();
+// Calculate the range size
+const range = maximum - minimum + 1;
+// Scale the decimal to the range
+const scaled = decimal * range;
+// Round down to get a whole number
+const floored = Math.floor(scaled);
+// Shift the number up to start at the minimum
+const result = floored + minimum;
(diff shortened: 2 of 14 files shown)


Sprint-1/1-key-exercises/1-count.js
```diff
@@ -4,3 +4,6 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+
+// The = operator assigns a new value to 'count'. It takes the current value of 'count'.
+// adds 1 to it, and then stores that new value back into the 'count' variable.
```

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -1,9 +1,26 @@
...
-const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
+const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum; // Generates a random whole number between minimum and maximum
 
 // In this exercise, you will need to work out what num represents?
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+// Generate a random decimal number between 0 and 1
+const decimal = Math.random();
```

We can use the concise and precise ***interval notation*** to describe a range of values. 
- `[`, `]` => inclusion
- `(`, `)` => exclusion

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -25,3 +25,28 @@ console.log(`£${pounds}.${pence}`);
...
+// Removes the final "p" character so only the numeric part of the price remains
+
+// 3. const paddedPenceNumberString = penceStringWithoutTrailingP.padStart(3, "0"): 
+// Ensures the number is atleast 3 digits long by adding leading zeros when necessary, which helps with consistent formatting
+
+// 4. const pounds = paddedPenceNumberString.substring(0, paddedPenceNumberString.length - 2): 
+// Extracts everything except the last two digits to form the pounds portion of the price
+
+// 5. const pence = paddedPenceNumberString .substring(paddedPenceNumberString.length - 2) .padEnd(2, "0"):
+// Takes the final two digits of the padded string to form the pence portion
```
cjyuan, line 39: Could we expect this program to work as intended for any valid `penceString` if we deleted `.padEnd(2, "0")` from the code?

Sprint-1/4-stretch-explore/objects.md
```diff
@@ -5,12 +5,19 @@ In this activity, we'll explore some additional concepts that you'll encounter i
...
+- console {debug: ƒ, error: ƒ, info: ƒ, log: ƒ, warn: ƒ, …}
 
 Try also entering `typeof console`
+- 'object'
 
 Answer the following questions:
 
 What does `console` store?
+- The console stores a history of messages your code prints while it runs.
+
```
cjyuan: I think this question is asking, "what properties can be found in the object assigned to the variable `console`.

I don' think the stores the text that has already been output.

cjyuan (commented):
Code looks good. I just have a few suggestions and questions.

Sprint-1/1-key-exercises/1-count.js
Eugenie-A: I have updated the comment on line 3 to be more concise and included the one word programming term you were asking for: **increment**.

Sprint-1/1-key-exercises/4-random.js
Eugenie-A: I've updated the comment to use precise interval notation. It now matches the official Math.random() behaviour (≥ 0 and < 1).

Sprint-1/3-mandatory-interpret/3-to-pounds.js
Eugenie-A: After testing: removing `.padEnd(2, "0")` doesn't change the output for valid inputs (e.g. "399p" → "£3.99") because `.padStart(3, "0")` already gurantees the last two characters are correctly padded. 

So no, you don't really need `.padEnd(2, "0")` in this code, the output remains correct without it.

Sprint-1/4-stretch-explore/objects.md
Eugenie-A: You are totally right, the question is about the properties/methods of the `console` object itself, not the output history. Thank you for the feedback!

Eugenie-A:
Hi @cjyuan Thank you for the feedback, I was hoping you could check my changes to the files.

cjyuan:
Can you add "Needs review" label?

Eugenie-A:
> Can you add "Needs review" label?

I have changed the label to 'Needs review'

cjyuan:
Thanks for addressing each inline comment clearly! Well done!

illicitonion:

PR #931, opened 2026-02-04
London | 26-ITP-January | Eugenie Ahangama | Sprint 2 | coursework/sprint 2

Eugenie-A:
## Learners, PR Template

Self checklist

## Changelist

I worked through the Sprint 2 exercises by going through the starter code, identifying the main errors and making sure I understood why the logic wasn't working. I then wrote tests in the debug section to clearly show the issues before fixing anything. After that, I implemented the correct JavaScript behaviour so the function produced the right results. Finally, in the interpret task, I explained what the original code was doing, why it failed and how my fixes solved those problems.

MDN References I used:
[String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice)
[Number() constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/Number)
[Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)
[console.assert()](https://developer.mozilla.org/en-US/docs/Web/API/console/assert)
[String.prototype.split()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split)
[Array.prototype.map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map)
[Remainder operator (%)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder)
[String.prototype.padStart()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart)
[String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring)
[String.prototype.replace()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace)
[String.prototype.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)
[String.prototype.toUpperCase()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase)
[Function parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions#function_parameters)
[Return](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/return)

## Questions

What's the best way to structure a debugging process?

### DIFF
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -16,4 +16,8 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
+    return (weight / (height * height)).toFixed(1);  // toFixed(1) rounds to 1 decimal place.
 }
+
+// Test case
+console.log(calculateBMI(70, 1.73)) // Output: 23.4
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -16,8 +16,10 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
-    return (weight / (height * height)).toFixed(1);  // toFixed(1) rounds to 1 decimal place.
+    // Number(...) converts the string from toFixed back into a number
+    return Number( (weight / (height * height)).toFixed(1) );  // toFixed(1) rounds to 1 decimal place.
 }
 
 // Test case
 console.log(calculateBMI(70, 1.73)) // Output: 23.4
+console.log(typeof calculateBMI(70, 1.73)); // number
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -1,13 +1,29 @@
 // Predict and explain first...
 //  =============> write your prediction here
+// Prediction: The code will throw a SyntaxError because the variable 'str' has already been declared as a parameter of the function.
 
 // call the function capitalise with a string input
 // interpret the error message and figure out why an error is occurring
 
+// Original code with error:
+
+/*
 function capitalise(str) {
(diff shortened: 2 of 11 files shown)


Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -16,4 +16,8 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    return (weight / (height * height)).toFixed(1);  // toFixed(1) rounds to 1 decimal place.
```

```javascript
  
```

Eugenie-A: I expected the function to return a number, not a string.
No, my function didn't return the type of value I was expecting it to, but rather it was returned as a string due to `.toFixed(1)`.
I updated it to `return Number( (weight / (height * height)).toFixed(1) );` - This keeps the rounding to 1 decimal place but converts back to a number.

cjyuan:
> What's the best way to structure a debugging process?

I do not understand what you mean. Can you elaborate?

Eugenie-A:
> > What's the best way to structure a debugging process?
> 
> I do not understand what you mean. Can you elaborate?

I'm trying to get better at debugging JavaScript code in a consistent, step by step way especially when fixing broken functions or understanding why something fails.

So what I'm asking is what are some reliable structured debugging habits or checklists that experienced developers use?

cjyuan:
The key of debugging is to find out which expression does not evaluate to the expected value or which variable does not hold the expected value.

For small amount of code, we can output the value of suspected expressions or variables.
For larger code base, we can use a debugger. We can use it to set break points, inspect value of variables, and trace the execution of the the code line by line.

illicitonion:

PR #939, opened 2026-02-07
London | ITP-Jan-26 | Mohsen Zamani | Sprint 3 | stretch

mohsenzamanist:
## Learners, PR Template

Self checklist

## Changelist

Comelete exercises in directory called 4-stretch

### DIFF
--- a/Sprint-3/4-stretch/card-validator.js
+++ b/Sprint-3/4-stretch/card-validator.js
@@ -1,0 +1,30 @@
+function creditCardValidator(cardNum) {
+  // Splits the cardNum into an array of characters
+  const cardNumArray = cardNum.split("");
+
+  // Checks if length is 16, otherwise return false
+  if (cardNumArray.length !== 16) return false;
+
+  // checks if all digits are numbers
+  if (
+    !cardNumArray.every(
+      (num) => num.charCodeAt(0) >= 48 && num.charCodeAt(0) <= 57
+    )
+  )
+    return false;
+
+  // Checks if there are at least two different digits
+  const count = cardNumArray.reduce((acc, curr) => {
+    acc[curr] = acc[curr] ? acc[curr] + 1 : 1;
+    return acc;
+  }, {});
+  if (Object.keys(count).length < 2) return false;
+
+  // Checks if the sum of all digits is greater than 16
+  const sumOfDigits = cardNumArray.reduce((acc, curr) => acc + Number(curr), 0);
+  if (sumOfDigits <= 16) return false;
+
+  //checks if last digit is even
+  if (cardNumArray[cardNumArray.length - 1] % 2 !== 0) return false;
+  return true;
+}
--- a/Sprint-3/4-stretch/card-validator.js
+++ b/Sprint-3/4-stretch/card-validator.js
@@ -6,7 +6,7 @@ function creditCardValidator(cardNum) {
   if (cardNumArray.length !== 16) return false;
 
   // checks if all digits are numbers
-  if (!cardNumArray.every((num) => num >= 0 && num <= 9)) return false;
+  if (!cardNumArray.every((num) => num >= "0" && num <= "9")) return false;
 
   // Checks if there are at least two different digits
   const count = new Set(cardNumArray);
--- a/Sprint-3/4-stretch/card-validator.js
+++ b/Sprint-3/4-stretch/card-validator.js
@@ -6,19 +6,11 @@ function creditCardValidator(cardNum) {
   if (cardNumArray.length !== 16) return false;
 
   // checks if all digits are numbers
-  if (
-    !cardNumArray.every(
-      (num) => num.charCodeAt(0) >= 48 && num.charCodeAt(0) <= 57
-    )
-  )
-    return false;
+  if (!cardNumArray.every((num) => num >= 0 && num <= 9)) return false;
 
   // Checks if there are at least two different digits
-  const count = cardNumArray.reduce((acc, curr) => {
-    acc[curr] = acc[curr] ? acc[curr] + 1 : 1;
-    return acc;
-  }, {});
-  if (Object.keys(count).length < 2) return false;
+  const count = new Set(cardNumArray);
+  if (count.size < 2) return false;
 
   // Checks if the sum of all digits is greater than 16
   const sumOfDigits = cardNumArray.reduce((acc, curr) => acc + Number(curr), 0);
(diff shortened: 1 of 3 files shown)


Sprint-3/4-stretch/card-validator.js
```diff
@@ -0,0 +1,30 @@
...
+  const cardNumArray = cardNum.split("");
+
+  // Checks if length is 16, otherwise return false
+  if (cardNumArray.length !== 16) return false;
+
+  // checks if all digits are numbers
+  if (
+    !cardNumArray.every(
+      (num) => num.charCodeAt(0) >= 48 && num.charCodeAt(0) <= 57
+    )
```
cjyuan: Note: We could also check if `num` is a digit as `num >= '0' && num <= '9'`.

Sprint-3/4-stretch/card-validator.js
```diff
@@ -0,0 +1,30 @@
...
+  )
+    return false;
+
+  // Checks if there are at least two different digits
+  const count = cardNumArray.reduce((acc, curr) => {
+    acc[curr] = acc[curr] ? acc[curr] + 1 : 1;
+    return acc;
+  }, {});
+  if (Object.keys(count).length < 2) return false;
```
cjyuan: This works.
Challenge: Implement a simpler and more efficient way to check distinct digit count with `Set`.

Sprint-3/4-stretch/password-validator.js
```diff
@@ -1,6 +1,12 @@
+const previousPasswords = ["5B43n21"];
 function passwordValidator(password) {
```
cjyuan: If we design the function as
```
function passwordValidator(password, previousPasswords=[]) {
  ...
}
```
we could let the caller specify what the previous passwords are.

Sprint-3/4-stretch/password-validator.js
```diff
@@ -1,6 +1,12 @@
+const previousPasswords = ["5B43n21"];
```
cjyuan: An invalid password is not a good candidate for a previous password.

Sprint-3/4-stretch/password-validator.test.js
```diff
@@ -16,11 +16,64 @@ You must breakdown this problem in order to solve it. Find one test case first a
...
-    const password = "12345";
-    // Act
-    const result = isValidPassword(password);
-    // Assert
-    expect(result).toEqual(true);
-}
-);
\ No newline at end of file
+  // Arrange
+  const password = "A1b2";
```
cjyuan: The function can return `false` for multiple reasons.

For example, the function might return `false` for "A1b2" not only because it is fewer than 5 characters, but also because it lacks a special letter.
As a result, we can't be certain that the function correctly handles the case of passwords shorter than 5 characters, since multiple conditions are being violated simultaneously.

Sprint-3/4-stretch/password-validator.test.js
```diff
@@ -16,11 +16,64 @@ You must breakdown this problem in order to solve it. Find one test case first a
...
+});
+
+test("password contains at least one uppercase English letter", () => {
+  // Arrange
+  const password = "1B2345";
+  // Act
+  const result = isValidPassword(password);
+  // Assert
+  expect(result).toEqual(false);
+});
```
cjyuan: - These two tests have the same description but they are testing two different things.

Sprint-3/4-stretch/password-validator.test.js
cjyuan, line 1: Could you make the test descriptions more detailed and descriptive? That way, if a test fails, the person implementing the function can quickly understand what went wrong and why.

Right now, for example, "password contains at least one number (0-9)" does not quite indicate what the test is testing. (Did you mean to say "not containing")

Sprint-3/4-stretch/card-validator.js
```diff
@@ -0,0 +1,23 @@
+function creditCardValidator(cardNum) {
+  // Splits the cardNum into an array of characters
+  const cardNumArray = cardNum.split("");
+
+  // Checks if length is 16, otherwise return false
+  if (cardNumArray.length !== 16) return false;
+
+  // checks if all digits are numbers
+  if (!cardNumArray.every((num) => num >= 0 && num <= 9)) return false;
```
cjyuan:
```javascript
if (!cardNumArray.every((num) => num >= 0 && num <= 9)) return false;
if (!cardNumArray.every((num) => num >= '0' && num <= '9')) return false;
```
Both produce the same result, the latter expression is more a commonly used.

Performance wise, while it is hardly noticeable, the former expression is less efficient 
because it requires the number to be converted to string first before the comparison takes place.

Sprint-3/4-stretch/password-validator.test.js
```diff
@@ -15,12 +15,65 @@ To be valid, a password must:
...
+test("should return false if password does not contain an lowercase English letter", () => {
+  // Arrange
+  const password = "1B2345%";
+  // Act
+  const result = isValidPassword(password);
+  // Assert
+  expect(result).toEqual(false);
+});
+
+test("should return false if password has no numbers(0-9)", () => {
```
cjyuan: Could shorten the description by using the term **digit**.

cjyuan (commented):
- Did you run Jest test after you made the changes? It seems one of the tests in `password-validator.test.js` has a bug and will fail.

mohsenzamanist:
I ran jest and password-validator.test.js works perfectly, all the tests passed.

Sprint-3/4-stretch/password-validator.test.js
```diff
@@ -15,12 +15,65 @@ To be valid, a password must:
...
+  const result = isValidPassword(password);
+  // Assert
+  expect(result).toEqual(false);
+});
+
+test("should return false if password was previously used", () => {
+  // Arrange
+  const password = "5B43n21!";
+  // Act
+  const result = isValidPassword(password, "5B43n21!");
```
cjyuan, line 31: I expected this test to fail because the second parameter isn't an array of strings.

I didn't realise string also has a method `.includes()` and your function manages to pass this test. 

However, other implementations that expect the second parameter to be an array may fail this test unexpectedly.

cjyuan (commented):
Great work!

You have all the info I intend to pass to you. So I will go ahead and mark this PR as complete first.

Sprint-3/4-stretch/password-validator.test.js
mohsenzamanist: Appreciate all the comments. 
I missed that it should be an array and it's surprising the code is working.
But if function explicitly checks for an array then it'll fail, so I'll add the check to the function and change the test as well.
Thanks.

illicitonion:

PR #951, opened 2026-02-10
London|26-ITP-January|Alexandru Pocovnicu|Sprint 3 |stretch

alexandru-pocovnicu:
## Learners, PR Template

Self checklist

## Changelist

set up script and test files for card validator, done tdd to check if a password and credit card are valid

##Questions
got stuck on how to check if a password was used before , don't really know where or what to look for

### DIFF
--- a/Sprint-3/4-stretch/card-validator.js
+++ b/Sprint-3/4-stretch/card-validator.js
@@ -1,16 +1,13 @@
 function validateNumber(number) {
   let arrNumber = [...number.toString()];
 
-  
-
   return arrNumber.length === 16 &&
     arrNumber.every((x) => x >= "0" && x <= "9") &&
     new Set(arrNumber).size > 1 &&
-    arrNumber[arrNumber.length-1]%2===0 &&
-    arrNumber.reduce((acc,cur)=>+acc+(+cur),0)>16
+    arrNumber[arrNumber.length - 1] % 2 === 0 &&
+    arrNumber.reduce((acc, cur) => +acc + +cur, 0) > 16
     ? true
     : false;
 }
-console.log(validateNumber(11111111111112));
 
 module.exports = validateNumber;
--- a/Sprint-3/4-stretch/card-validator.js
+++ b/Sprint-3/4-stretch/card-validator.js
@@ -1,5 +1,5 @@
 function validateNumber(number) {
-  let arrNumber = [...number.toString()];
+  const arrNumber = [...number.toString()];
 
   return arrNumber.length === 16 &&
     arrNumber.every((x) => x >= "0" && x <= "9") &&
--- a/Sprint-3/4-stretch/card-validator.js
+++ b/Sprint-3/4-stretch/card-validator.js
@@ -1,4 +1,16 @@
-function validateNumber(){
+function validateNumber(number) {
+  let arrNumber = [...number.toString()];
 
+  
+
+  return arrNumber.length === 16 &&
+    arrNumber.every((x) => x >= "0" && x <= "9") &&
+    new Set(arrNumber).size > 1 &&
+    arrNumber[arrNumber.length-1]%2===0 &&
+    arrNumber.reduce((acc,cur)=>+acc+(+cur),0)>16
+    ? true
+    : false;
 }
-module.exports=validateNumber
+console.log(validateNumber(11111111111112));
+
+module.exports = validateNumber;
--- a/Sprint-3/4-stretch/card-validator.js
+++ b/Sprint-3/4-stretch/card-validator.js
@@ -1,0 +1,4 @@
+function validateNumber(){
+
+}
+module.exports=validateNumber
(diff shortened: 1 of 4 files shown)


Sprint-3/4-stretch/card-validator.js
```diff
@@ -0,0 +1,13 @@
+function validateNumber(number) {
+  let arrNumber = [...number.toString()];
```
cjyuan: When a variable is not going to be reassigned a value, best practice is to declare it using `const`.

Sprint-3/4-stretch/card-validator.test.js
```diff
@@ -0,0 +1,50 @@
...
+test("number should be 16 digits long",()=>{
+    expect(validateNumber(1029384756820562)).toEqual(true)
+})
+test("number should be 16 digits long", () => {
+  expect(validateNumber(10293847568202)).toEqual(false);
+});
+test("number should be 16 digits long", () => {
+  expect(validateNumber(1029384756820512348)).toEqual(false);
+});
+
```
cjyuan: When one of these tests fails, these test descriptions do not quite tell the developer why their function failed.

Can you make the test descriptions on this file more informative?

Sprint-3/4-stretch/password-validator.js
```diff
@@ -1,6 +1,9 @@
 function passwordValidator(password) {
-    return password.length < 5 ? false : true
+  if (password.match(/[A-Z]/) && password.match(/[a-z]/) &&
```
cjyuan: `string.match(regex)` would also return an array of the matched substring if there is a match, whereas `regex.test(string)` returns a Boolean value if there is a match.

The latter is slightly more performant, and consume less memory.

Sprint-3/4-stretch/password-validator.test.js
cjyuan, line 1: Can you also make the test descriptions on this file more informative?

Sprint-3/4-stretch/password-validator.test.js
```diff
@@ -10,17 +10,73 @@ To be valid, a password must:
...
     expect(result).toEqual(true);
 }
-);
\ No newline at end of file
+);
+test("password has at least 5 characters", () => {
+  const password = "1234";
+  const result = isValidPassword(password);
+  expect(result).toEqual(false);
+});
```
cjyuan: The function can return `false` for multiple reasons.

For example, the function might return `false` for "1234" not only because it is fewer than 5 characters, but also because it lacks a lowercase letter.
As a result, we can't be certain that the function correctly handles the case of passwords shorter than 5 characters, since multiple conditions are being violated simultaneously.

Sprint-3/4-stretch/password-validator.js
alexandru-pocovnicu: thank you

alexandru-pocovnicu:
thank you

Sprint-3/4-stretch/card-validator.test.js
cjyuan: These three test descriptions are greatly improved! Why not also update the remaining test descriptions?

Sprint-3/4-stretch/password-validator.test.js
```diff
@@ -10,17 +10,72 @@ To be valid, a password must:
...
+  const password = "1aS!";
+  const result = isValidPassword(password);
+  expect(result).toEqual(false);
+});
+
+test("should require at least one uppercase letter", () => {
+  const password = "12345Aaoe$";
+  const result = isValidPassword(password);
+  expect(result).toEqual(true);
+});
```
cjyuan: When this test fails, it only means the function fails to recognise a valid password -- which may or may not related to uppercase letters.

I think you can consider grouping all valid cases into the same category.

Sprint-3/4-stretch/password-validator.test.js
```diff
@@ -10,17 +10,72 @@ To be valid, a password must:
...
+test("should reject password without uppercase letter", () => {
+  const password = "12345";
```
cjyuan: This sample does not satisfy the criterium: it meets all other conditions except the one you're targeting.

There are a few more test samples in this file that fit this description.

Can you update them accordingly?

alexandru-pocovnicu:
thank you, I'm still not sure about checking if the password has been used before, at the moment it's just checking against an empty array

Sprint-3/4-stretch/card-validator.test.js
```diff
@@ -0,0 +1,50 @@
...
+test("should return true if at least one of the digits isn't the same as the others",()=>{
+    expect(validateNumber(3636363636363636)).toEqual(true)
+})
+test("should return true if at least one of the digits isn't the same as the others", () => {
+  expect(validateNumber(3333333333333336)).toEqual(true);
+});
+test("should return false if at least one of the digits isn't different from all the others", () => {
+  expect(validateNumber(2222222222222222)).toEqual(false);
+});
```
cjyuan: Suggestion: Ask an AI tool how these test descriptions can be made more concise.

No change needed. 3

cjyuan (commented):
Changes look good. Well done!

>  I'm still not sure about checking if the password has been used before, at the moment it's just checking against an empty array

What are you unsure about?

illicitonion:

PR #952, opened 2026-02-10
Cape Town | 2026-ITP-Jan | Pretty Taruvinga | Sprint 1 | Mandatory errors

Pretty548:
## Learners, PR Template

Self checklist

## Changelist

1. Used const where the value needed to change → changed to let
2. Declared a variable after using it → moved declaration above usage
3. Tried to use .slice() on a number → converted to string first
4. Duplicate variable declaration → removed the extra one
5. Invalid variable name starting with a number → renamed it

## Questions

Is there a clearer or more consistent way I should decide when to use const vs let in future exercises?


A-O-Emmanuel (commented):
good job with this task, I don't have any other thing to add, other reviewers might have.

PR #956, opened 2026-02-10
 Cape Town | 2026-ITP-Jan | Isaac Abodunrin | Sprint 3 | Implement and Rewrite Tests Coursework

bytesandroses:
## Self checklist

## Changelist
**Implemented solutions to the following, using console.assert() for basic testing:**
- 1-get-angle-type.js
- 2-is-proper-fraction.js
- 3-get-card-value.js

**Rewrite tests for the solutions above using Jest**

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -1,24 +1,11 @@
-// Implement a function isProperFraction,
-// when given two numbers, a numerator and a denominator, it should return true if
-// the given numbers form a proper fraction, and false otherwise.
-
-// Assumption: The parameters are valid numbers (not NaN or Infinity).
-
-// Note: If you are unfamiliar with proper fractions, please look up its mathematical definition.
-
-// Acceptance criteria:
-// After you have implemented the function, write tests to cover all the cases, and
-// execute the code to ensure all tests pass.
-
 function isProperFraction(numerator, denominator) {
-  // TODO: Implement this function
+  return Math.abs(numerator) < Math.abs(denominator);
 }
 
 // The line below allows us to load the isProperFraction function into tests in other files.
 // This will be useful in the "rewrite tests with jest" step.
 module.exports = isProperFraction;
 
-// Here's our helper again
 function assertEquals(actualOutput, targetOutput) {
   console.assert(
     actualOutput === targetOutput,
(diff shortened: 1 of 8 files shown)


Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -1,33 +1,30 @@
...
-// Note: If you are unfamiliar with proper fractions, please look up its mathematical definition.
+function isProperFraction(numerator, denominator) {
+  // Return non numeric numerators and denominators as false
+  if (typeof numerator != "number" || typeof denominator != "number")
+    return false;
 
-// Acceptance criteria:
-// After you have implemented the function, write tests to cover all the cases, and
-// execute the code to ensure all tests pass.
+  if (numerator === 0 || denominator === 0) return false;
```
cjyuan: Isn't 0/1 a proper fraction?

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -1,20 +1,42 @@
...
+// Case 3: Number Cards (2-10)
+const cardsNumbers = [2, 3, 4, 5, 6, 7, 8, 9, 10];
+for (const rank of cardsNumbers) {
+  for (const suit of suits) {
+    const cardFace = `${rank}${suit}`;
+    test(`Should return ${rank} when given a card ${cardFace}`, () => {
+      expect(getCardValue(cardFace)).toEqual(rank);
+    });
+  }
+}
```
cjyuan: With this way of expressing the test, the test description
 `Should return ${rank} when given a card ${cardFace}`
would appear 36 times on the console (with different ranks and suits).

Should you need to reduce the number of messages on the console, you can consider grouping similar test cases under one category as 
```
test(`Should return the numerical value of number cards`, () => {
  const cardsNumbers = [2, 3, 4, 5, 6, 7, 8, 9, 10];
  for (const rank of cardsNumbers) {
    for (const suit of suits) {
      const cardFace = `${rank}${suit}`;
      expect(getCardValue(cardFace)).toEqual(rank);
    }
  }
}
```

cjyuan (commented):
Very thorough tests! Well done!

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
bytesandroses: Good catch -- you're absolutely right since the definition of a proper fraction is where the magnitude of the denominator is great than that of the numerator. It doesn't say the numerator can't be zero, so I've updated the code to correct this

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
bytesandroses: Thanks for the tip. I've refactored it so that similar test cases use one category -- there are a lot fewer printouts in the console now

illicitonion:

PR #958, opened 2026-02-11
Birmingham| ITP-Jan | Ahmad Roman Sanaye| Sprint 1| Structuring and Testing Data

RomanSanaye:
### Self checklist

- [x] I have tested my changes using Node

### Changelist

- Completed Sprint 1 exercises
- Ran all files using Node to identify and understand errors
- Added explanations and comments to interpret the code
- Explored stretch tasks to reinforce understanding

### DIFF
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -11,18 +11,24 @@
 
 const filePath = "/Users/mitch/cyf/Module-JS1/week-1/interpret/file.txt";
 const lastSlashIndex = filePath.lastIndexOf("/");
-const base = filePath.slice(-8);
-console.log(`The base part of ${filePath} is ${base}`);
+const base = filePath.substring(lastSlashIndex + 1);
+console.log(`Base part:  ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
 
-const dir = filePath.slice(0, -8);
-const ext = filePath.slice(filePath.length-3);
+const dir = filePath.substring(0, lastSlashIndex + 1);
 
-console.log(`the dir part of filePath variable is ${dir}`);
-console.log(ext);
+const dotIndex = filePath.lastIndexOf(".");
+const ext = filePath.substring(dotIndex + 1);
+
+console.log(`Dir part: ${dir}`);
+console.log(`Ext part: ${ext}`);
 
 // https://www.google.com/search?q=slice+mdn
 
 // After studying the slice() method I was able to do this task easily.
+// The method I used is not working with all, for example if we change the name of our file from file.txt to best-file-ever.txt then out put is not what we want. ===> ever.txt; so we should find an other way ...
+// to solve that we better use substring(lastSlashIndex + 1)method. this method works with any name we add.
+// Before I used slice() which was working for the specific name or path but not for all type. and now this substring() method does work well almost on any type of path.
+// thanks to "hkavalikas" because of his feedback I learned this useful method.
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -11,13 +11,18 @@
 
 const filePath = "/Users/mitch/cyf/Module-JS1/week-1/interpret/file.txt";
 const lastSlashIndex = filePath.lastIndexOf("/");
-const base = filePath.slice(lastSlashIndex + 1);
+const base = filePath.slice(-8);
 console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
 
-const dir = ;
-const ext = ;
+const dir = filePath.slice(0, -8);
+const ext = filePath.slice(filePath.length-3);
+
+console.log(`the dir part of filePath variable is ${dir}`);
+console.log(ext);
 
 // https://www.google.com/search?q=slice+mdn
+
+// After studying the slice() method I was able to do this task easily.
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -2,8 +2,14 @@ const minimum = 1;
 const maximum = 100;
 
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
+console.log(num);
 
 // In this exercise, you will need to work out what num represents?
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+// num is a variable that will be assigned an integer number after the operators are done.
+// Math.floor() rounds a decimal to its nearest number to make it an integer. ex = 1.2 => 1.
+// Math.random() is a method used to create numbers between (0-1) and usually it creates a decimal like (0,2) or etc.
+// each time I run it, I got different number as Math.random() generates different number each time we run it.
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -13,3 +13,11 @@ console.log(num);
 // Math.floor() rounds a decimal to its nearest number to make it an integer. ex = 1.2 => 1.
 // Math.random() is a method used to create numbers between (0-1) and usually it creates a decimal like (0,2) or etc.
 // each time I run it, I got different number as Math.random() generates different number each time we run it.
+
+//to respond to the question "What would happen if we did Math.floor(1.9);?"
+// ===> the method Math.floor() rounds the decimal number down always, in this case the final output would be ==> 2 as our 1.9 would be 1 after our method round it and add it to our minimum number which is 1 so ==> 1 + 1 = 2.
+
+// In addition our formula works we wont have 191, if we had Math.ceil() then the result would exceed our maximum number.
+
+// Response to question "What does the + minimum do here?"
+// ===> + minimum shifts the range so that the smallest possible number is minimum instead of 0. so if we don't have it our range would be (0, 99) and this as we have 1 as minimum and 100 as maximum, if we don't add + minimum then our formula wont work as we are expecting.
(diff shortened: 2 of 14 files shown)


Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -11,13 +11,18 @@
 
 const filePath = "/Users/mitch/cyf/Module-JS1/week-1/interpret/file.txt";
 const lastSlashIndex = filePath.lastIndexOf("/");
-const base = filePath.slice(lastSlashIndex + 1);
+const base = filePath.slice(-8);
```
hkavalikas: We have updated the provided `lastSlashIndex` with a magic number like `-8`, what would happen if we now renamed our `file.txt` to `best-file-ever.txt`? Would this solution still work?

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -2,8 +2,14 @@ const minimum = 1;
...
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
+console.log(num);
 
 // In this exercise, you will need to work out what num represents?
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+// num is a variable that will be assigned an integer number after the operators are done.
+// Math.floor() rounds a decimal to its nearest number to make it an integer. ex = 1.2 => 1.
```
hkavalikas, line 13: What would happen if we did ```Math.floor(1.9);```?

Sprint-1/1-key-exercises/4-random.js
hkavalikas, line 1: What does the `+ minimum` do here?

Sprint-1/2-mandatory-errors/0.js
```diff
@@ -1,2 +1,8 @@
+
 This is just an instruction for the first activity - but it is just for human consumption
-We don't want the computer to run these 2 lines - how can we solve this problem?
\ No newline at end of file
+We don't want the computer to run these 2 lines - how can we solve this problem?
+
+//solution: Simply comment the lines.
+// 1- single line comment like: // example ...
+// 2- multiline comment like : /* a, b, c */
+// comments are for instruction of guidance of who will read our codes, but the computer wont read them.
```
hkavalikas, line 8: Explanation is spot on, let's take it a step further and execute it! I want to run this file without a syntax error. Can we make it possible?

Sprint-1/2-mandatory-errors/1.js
```diff
@@ -1,4 +1,10 @@
 // trying to create an age variable and then reassign the value by 1
 
 const age = 33;
-age = age + 1;
+//age = age + 1;
+
+// we can't do that with const variable, because JS locks the reference.
```
hkavalikas, line 7: Also spot on, let's take the approach and fix the broken code over declaring additional variables.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -12,11 +12,13 @@ console.log(`The percentage change is ${percentageChange}`);
 // Read the code and then answer the questions below
 
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
-
+    // line: 4 and line 5 and it is replaceAll().
```
hkavalikas: There are a few more. Can you try to spot them?

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -12,11 +12,13 @@ console.log(`The percentage change is ${percentageChange}`);
...
+    // error is coming from line 5. The error is because of not putting comma between the inputs of replaceAll function. 
+    // To fix the problem just add a comma between the inputs of replaceAll function.
 // c) Identify all the lines that are variable reassignment statements
-
+    // Line 4 and 5.
 // d) Identify all the lines that are variable declarations
-
+    // we have variable declarations at lines 1, 2, 7 and 8.
 // e) Describe what the expression Number(carPrice.replaceAll(",","")) is doing - what is the purpose of this expression?
+    // the purpose of that expression is to turn into number the string and replace all commas  with an empty space.
```
hkavalikas: This comment is mostly correct, but I do want to highlight the second part of your answer. Is there anything we can do to improve it slightly?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,17 @@ console.log(result);
...
-
+    // here we have only one which is console.log().
 // c) Using documentation, explain what the expression movieLength % 60 represents
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators
-
+    // Remainder operator (%), returns the leftover when one operand divided by the other operand.
+    // movieLength % 60 expression represents the remaining seconds.
 // d) Interpret line 4, what does the expression assigned to totalMinutes mean?
-
+    // that expression turns the seconds into minutes.
```
hkavalikas: You are right, but let's try to break it down step by step, how is the expression doing that?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,17 @@ console.log(result);
...
 // e) What do you think the variable result represents? Can you think of a better name for this variable?
-
+    // the result variable represents the movie length in Hour, minutes and seconds.
+    // a better name can be "movieLength".
```
hkavalikas: Can you see any problem if we were to name it `movieLength`?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,17 @@ console.log(result);
...
 // f) Try experimenting with different values of movieLength. Will this code work for all values of movieLength? Explain your answer
+    // as I changed the value of the movieLength the results has changed too.It means this code works with any value.
```
hkavalikas: There are a few cases that might break this. Can you spot and mention a few?

Sprint-1/4-stretch-explore/objects.md
hkavalikas, line 1: Optional: There are a few questions before the "Answer the following questions" too, if you want to add an answer to those too for completeness.

hkavalikas (changes requested):

Sprint-1/1-key-exercises/3-paths.js
RomanSanaye: Thank you for the feedback!  @hkavalikas 🙏
I removed the magic number and updated the logic to use lastIndexOf("/") to dynamically locate the last slash, combined with substring() to safely extract the file name.
With this approach, the code will continue to work correctly even if the file name changes .

Sprint-1/1-key-exercises/4-random.js
RomanSanaye: Thanks for the feedback!
Math.floor() always rounds a number down to the nearest integer by removing the decimal part. Here it will return 1.

Sprint-1/1-key-exercises/4-random.js
RomanSanaye: Thank you for the feedback! 
The + minimum shifts the random value into the desired range.
Without it, the formula would generate numbers between 0 and (maximum - minimum).
By adding minimum, the result is shifted to fall within the range of minimum to maximum (for example, 1 – 100).

Sprint-1/2-mandatory-errors/0.js
RomanSanaye: Thank you for the feedback! 
I have updated the file by commenting out the instructional lines so they are no longer executed by the interpreter.

Sprint-1/2-mandatory-errors/1.js
RomanSanaye: Thank you for the feedback! 
I have updated the code by fixing the original variable instead of introducing additional variables.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
RomanSanaye: Thank you for the feedback! 
You’re right — in addition to replaceAll(), there are also function calls to Number() on lines 4 and 5, and console.log() on line 10.
I’ve updated the answer to reflect all function calls.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
RomanSanaye: Thank you for the feedback! 
We improved the explanation: the expression Number(carPrice.replaceAll(",", "")) first removes all commas from the string (for example, "1,000" becomes "1000"), and then converts the cleaned string into a number (1000).

Sprint-1/3-mandatory-interpret/2-time-format.js
RomanSanaye: Thank you for the feedback! 
I broke it down step by step:

movieLength % 60 gives the leftover seconds that don’t fit into a full minute.

movieLength - remainingSeconds removes those leftover seconds.

Dividing by 60 converts the remaining seconds into total minutes.

This shows clearly how the expression turns the total seconds into minutes while keeping track of leftover seconds.

Sprint-1/3-mandatory-interpret/2-time-format.js
RomanSanaye: Thank you for the feedback! 
If we name the final result variable movieLength, it could be confusing because we already have a variable movieLength holding the total seconds.
Using a different name, like movieDuration, makes the code clearer and avoids accidentally overwriting the original value.

Sprint-1/3-mandatory-interpret/2-time-format.js
RomanSanaye: Thank you for the feedback! 
Some cases that could break the code include:

Negative numbers (a movie cannot have negative seconds)

Decimal numbers (like 8777.22)

Invalid inputs (like strings, null, or undefined)

Handling these cases ensures the program works safely and produces correct results.

Sprint-1/4-stretch-explore/objects.md
RomanSanaye: Thank you for the feedback! 
I have added answers to the optional questions for completeness.

illicitonion:

PR #959, opened 2026-02-11
Glasgow | Jan-26 | Elisabeth Matulian | Sprint 2 | Coursework

Elisabeth-Matulian:
Self checklist

## Changelist

Completed all exercises

## Questions

No questions

### DIFF
--- a/Sprint-2/4-mandatory-interpret/time-format.js
+++ b/Sprint-2/4-mandatory-interpret/time-format.js
@@ -26,7 +26,7 @@ function formatTimeDisplay(seconds) {
 // =============> 0
 
 // c) What is the return value of pad is called for the first time?
-// =============> 00
+// =============> "00"
 
 // d) What is the value assigned to num when pad is called for the last time in this program?  Explain your answer
 // =============> 1. The remainder operator (%) returns the remainder left after dividing one number by another number. 61 % 60
--- a/Sprint-2/4-mandatory-interpret/time-format.js
+++ b/Sprint-2/4-mandatory-interpret/time-format.js
@@ -11,24 +11,25 @@ function formatTimeDisplay(seconds) {
   return `${pad(totalHours)}:${pad(remainingMinutes)}:${pad(remainingSeconds)}`;
 }
 
+
 // You will need to play computer with this example - use the Python Visualiser https://pythontutor.com/visualize.html#mode=edit
 // to help you answer these questions
 
 // Questions
 
 // a) When formatTimeDisplay is called how many times will pad be called?
-// =============> write your answer here
+// =============> 3 times
 
 // Call formatTimeDisplay with an input of 61, now answer the following:
 
 // b) What is the value assigned to num when pad is called for the first time?
-// =============> write your answer here
+// =============> 0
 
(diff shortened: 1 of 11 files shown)


Sprint-2/4-mandatory-interpret/time-format.js
```diff
@@ -11,24 +11,25 @@ function formatTimeDisplay(seconds) {
...
 
 // Call formatTimeDisplay with an input of 61, now answer the following:
 
 // b) What is the value assigned to num when pad is called for the first time?
-// =============> write your answer here
+// =============> 0
 
 // c) What is the return value of pad is called for the first time?
-// =============> write your answer here
+// =============> 00
```

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,8 +4,15 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
+  const mins = time.slice(3, 5);
```
cjyuan: Could also use `time.slice(-2)` to extract two characters from the right (or last two characters).

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,8 +4,15 @@
...
   if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+    return `${hours - 12}:${mins} pm`;
+  }
+  else if (hours == 0) {
+    return `${hours + 12}:${mins} am`;
```
cjyuan: 0 +12 is 12. Why not just write `12` directly?

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,8 +4,15 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
+  const mins = time.slice(3, 5);
   if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+    return `${hours - 12}:${mins} pm`;
+  }
```
cjyuan, line 10: If "01:00" is converted to "01:00 am", it is probably reasonable for the caller to expect "13:00" to be converted to "01:00 pm".

```
01:00 am
1:00 pm
12:00 am
01:00 pm
```

Elisabeth-Matulian: Thanks for the feedback, I've followed all comments. Could you take another look when you have a moment?

Sprint-2/4-mandatory-interpret/time-format.js
Elisabeth-Matulian: quotes added

Sprint-2/5-stretch-extend/format-time.js
Elisabeth-Matulian: fixed

Sprint-2/5-stretch-extend/format-time.js
Elisabeth-Matulian: Indeed. I didn't even notice.

cjyuan:
Changes look good. Well done.

illicitonion:

PR #960, opened 2026-02-12
London | 26-ITP-January | Laura C | Sprint 2 | Structuring and Testing Data

lauracs24:
### Self checklist

### Changelist

I worked through all Sprint 2 exercises by predicting errors, running the code with node, and explaining why issues occurred before fixing them. In the debug section, I identified logic problems and corrected the functions so they returned the correct values. I also implemented the required functions (calculateBMI, toUpperSnakeCase, and toPounds) and completed the interpret task by analysing the code step by step and answering the questions.

### Questions

What is a reliable way of learning javascript easily and testing yourself?

### DIFF
--- a/Sprint-2/2-mandatory-debug/2.js
+++ b/Sprint-2/2-mandatory-debug/2.js
@@ -3,22 +3,56 @@
 // Predict the output of the following code:
 // =============> Write your prediction here
 
-const num = 103;
+// I think the program will print 3 every time.
+// Even though different numbers are passed into the function,
+// the function is using the variable 'num', which is 103.
 
-function getLastDigit() {
-  return num.toString().slice(-1);
+
+// function getLastDigit() {
+//   return num.toString().slice(-1);
+// }
+
+
+// console.log(`The last digit of 42 is ${getLastDigit(42)}`);
+// console.log(`The last digit of 105 is ${getLastDigit(105)}`);
+// console.log(`The last digit of 806 is ${getLastDigit(806)}`);
+
+// Now run the code and compare the output to your prediction
+// =============> write the output here
+// The last digit of 42 is 3
+// The last digit of 105 is 3
+// The last digit of 806 is 3
+
+// Explain why the output is the way it is
+// =============> write your explanation here
+
+// The function is not using the number that is passed into it.
+// Instead, it always uses the global variable 'num'
+// which is set to 103.
+// The last digit of 103 is 3
+// so the function always returns 3.
+
+// Finally, correct the code to fix the problem
+// =============> write your new code here
+
+function getLastDigit(number) {
+  return number.toString().slice(-1);
 }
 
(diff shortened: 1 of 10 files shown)


Sprint-2/2-mandatory-debug/2.js
```diff
@@ -3,22 +3,56 @@
 // Predict the output of the following code:
 // =============> Write your prediction here
 
-const num = 103;
+// I think the program will print 3 every time.
+// Even though different numbers are passed into the function,
+// the function is using the variable 'num', which is 103.
```
cjyuan, line 8: It is possible to have a global variable and a parameter named `num` in a script. For example,

```javascript
const num = 103;

function getLastDigit(num) {
  return num.toString().slice(-1);
}
```

Can you find out why?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -14,6 +14,9 @@
 // Then when we call this function with the weight and height
 // It should return their Body Mass Index to 1 decimal place
 
-function calculateBMI(weight, height) {
+function calculateBMI(weight, height) {const bmi = weight / (height * height);
+  return bmi.toFixed(1);
```

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -14,6 +14,9 @@
 // Then when we call this function with the weight and height
 // It should return their Body Mass Index to 1 decimal place
 
-function calculateBMI(weight, height) {
+function calculateBMI(weight, height) {const bmi = weight / (height * height);
```
cjyuan: Can you improve the indentation of the code in this script?

cjyuan (commented):
Both the code and the explanations are good. 

I only have a question and some suggestions.

Sprint-2/3-mandatory-implement/1-bmi.js
lauracs24: I have corrected the indentation to improve readability

Sprint-2/3-mandatory-implement/1-bmi.js
lauracs24: Thank you for pointing this out @cjyuan . I updated the function to return a number instead of a string by wrapping toFixed(1) with Number()

Sprint-2/2-mandatory-debug/2.js
lauracs24: Ah I think the issue was not the shared variable name, but that the function did not define a parameter. I have updated my explanation to reflect this

cjyuan: Your change is good, and you are right about the function not having a parameter.

I meant to point out these two concepts, "***identifier scope*** and ***identifier resolution***, in the context of JavaScript programming". ChatGPT can give a good explanation.

cjyuan:
Changes look good. Well done.

illicitonion:

PR #963, opened 2026-02-13
Glasgow | 26-ITP-Jan| Martin McLean | Sprint 2 | coursework

mjm-git185:
Changelist

my coursework

### DIFF
--- a/Sprint-2/1-key-errors/2.js
+++ b/Sprint-2/1-key-errors/2.js
@@ -2,14 +2,21 @@
 // Predict and explain first BEFORE you run any code...
 
 // this function should square any number but instead we're going to get an error
+// There is no "num" veriabal defined. num should be in the beackets
+
 
 // =============> write your prediction of the error here
 
-function square(3) {
-    return num * num;
-}
+// function square(3) {
+//     return num * num;
+// }
 
-// =============> write the error message here
+// // =============> write the error message here
+// /home/martinmclean158/Module-Structuring-and-Testing-Data/Sprint-2/1-key-errors/0.js:8
+//   let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+//       ^
+
+// SyntaxError: Identifier 'str' has already been declared
 
 // =============> explain this error message here
 
@@ -17,4 +24,7 @@ function square(3) {
 
 // =============> write your new code here
 
-
+function square(num) {
+    return num * num;
+}
+console.log(square(3)); // should print 9
--- a/Sprint-2/1-key-errors/2.js
+++ b/Sprint-2/1-key-errors/2.js
@@ -2,19 +2,22 @@
 // Predict and explain first BEFORE you run any code...
 
 // this function should square any number but instead we're going to get an error
-// There is no "num" veriabal defined. num should be in the beackets
+// // There is no "num" veriabal defined. num should be in the beackets
 
 
-// =============> write your prediction of the error here
+// // =============> write your prediction of the error here
 
 // function square(3) {
 //     return num * num;
 // }
 
-// // =============> write the error message here
-// /home/martinmclean158/Module-Structuring-and-Testing-Data/Sprint-2/1-key-errors/0.js:8
-//   let str = `${str[0].toUpperCase()}${str.slice(1)}`;
-//       ^
+// // // // =============> write the error message here
+// // // /home/martinmclean158/Module-Structuring-and-Testing-Data/Sprint-2/1-key-errors/0.js:8
+// // /home/martinmclean158/Module-Structuring-and-Testing-Data/Sprint-2/1-key-errors/2.js:10
+// // function square(3) {
+//                 ^
+
+// SyntaxError: Unexpected number
 
 // SyntaxError: Identifier 'str' has already been declared
 
--- a/Sprint-2/2-mandatory-debug/0.js
+++ b/Sprint-2/2-mandatory-debug/0.js
@@ -17,7 +17,6 @@
 // //  =============> write your new code here
 function multiply(a, b) {
   const result = a * b;
-  console.log(result);
   return result;
 }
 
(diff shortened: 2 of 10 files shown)


Sprint-2/1-key-errors/2.js
```diff
@@ -2,19 +2,29 @@
...
-function square(3) {
-    return num * num;
-}
+// function square(3) {
+//     return num * num;
+// }
+
+// // =============> write the error message here
+// /home/martinmclean158/Module-Structuring-and-Testing-Data/Sprint-2/1-key-errors/0.js:8
+//   let str = `${str[0].toUpperCase()}${str.slice(1)}`;
```
Edu-Vin: I think the error message is wrong. Try running it again

Sprint-2/2-mandatory-debug/0.js
```diff
@@ -2,13 +2,23 @@
...
+// console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
 
-// =============> write your explanation here
+// // =============> write your explanation here
+
+// // Finally, correct the code to fix the problem
+// //  =============> write your new code here
+function multiply(a, b) {
+  const result = a * b;
+  console.log(result);
```
Edu-Vin: Since there is a console.log of the result outside the function, is there a need for the console.log in the function?

Sprint-2/2-mandatory-debug/2.js
```diff
@@ -1,24 +1,33 @@
...
+// console.log(`The last digit of 42 is ${getLastDigit(42)}`);
+// console.log(`The last digit of 105 is ${getLastDigit(105)}`);
+// console.log(`The last digit of 806 is ${getLastDigit(806)}`);
 
 // Now run the code and compare the output to your prediction
-// =============> write the output here
+// =============> yes thay all give 3 
 // Explain why the output is the way it is
-// =============> write your explanation here
+// because the variable is a consrent
```
Edu-Vin: Based on your explanation. 

It means if I add `const num = 103;` to your solution before `getLastDigit()`, will it return 3 for all the console logs?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,7 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    let bmi = weight/(height*height)
```
Edu-Vin: Can you explain why you used let instead of const?

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,9 @@
 // You will need to come up with an appropriate name for the function
 // Use the MDN string documentation to help you find a solution
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
+function SNAKE_CASE (mesasge){
```
Edu-Vin: Does the name of the function follow the appropriate naming convention? Does it explain what the function does?

Sprint-2/4-mandatory-interpret/time-format.js
```diff
@@ -11,24 +11,30 @@ function formatTimeDisplay(seconds) {
...
 
 // a) When formatTimeDisplay is called how many times will pad be called?
 // =============> write your answer here
+// Pad willo be called 3 times
 
 // Call formatTimeDisplay with an input of 61, now answer the following:
 
 // b) What is the value assigned to num when pad is called for the first time?
 // =============> write your answer here
+// "null"
```
Edu-Vin: Since totalHours is the value used to call pad for the first time. Try doing a console.log to be sure it is null

Edu-Vin (changes requested):

mjm-git185:
i have made the changes needed

Edu-Vin:
> i have made the changes needed

I can't find the changes you made.

Edu-Vin:
You are supposed to make your changes to this pull request. @mjm-git185

mjm-git185:
i edited them on vs code and pushed them to git-hub

Edu-Vin:
> i edited them on vs code and pushed them to git-hub

From your VS Code, you are supposed to checkout to this branch `mjm-git185:Coureswork_Sprint-2`, make your changes and push

mjm-git185:
i have pushed it up and the version i can see is the most recent one #7bc1235

Sprint-2/2-mandatory-debug/2.js
Edu-Vin: You didn't answer this.

Edu-Vin:
> i have pushed it up and the version i can see is the most recent one #7bc1235

Seen

mjm-git185:
that one has the revisions

Sprint-2/2-mandatory-debug/2.js
Edu-Vin: @mjm-git185 look at this

mjm-git185: // because the variable is a const before the code block when num is called it can't be changed

Edu-Vin: I think you should study global and local variables and how scope works. It will help you understand this code better

mjm-git185: my understanding was that the const out side of the code block was in the global scope and because of that would over ride the local scope of the code block, so when it is run the console.log would pick up the global and ignore the local "num"

Edu-Vin: If a new variable of the same name is declared within the code block, the code will use that new variable and ignore the global variable. Reason why when you passed num into that function, it used what was passed and not the global num.

Edu-Vin (approved):

illicitonion:

PR #976, opened 2026-02-17
Sheffield | 26-ITP-Jan | Mahmoud Shaabo | Sprint 1 | Module Structuring and Testing Data

mahmoudshaabo1984:
I have completed the coursework for Sprint 1.

**Work done:**
- Fixed all bugs in the `2-mandatory-errors` directory.
- Provided detailed interpretations and answered all questions in the `3-mandatory-interpret` directory.
- Ensured all variable naming and scoping issues were resolved.

**Learners, PR Template**

### DIFF
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -4,3 +4,4 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+// The code adds 1 to the count variable.
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -4,4 +4,5 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
-// The code adds 1 to the count variable.
+// Line 3 is an increment operation. The = operator is the assignment operator:
+// it evaluates the expression on the right (count + 1), then assigns the result back to count.
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -9,4 +9,11 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
 
 console.log(num);
-// num represents a random integer between 1 and 100.
+
+// num represents a random integer in the range [1, 100].
+// Breakdown:
+// 1. Math.random() returns a decimal number in the range [0, 1).
+// 2. (maximum - minimum + 1) calculates the size of the range, which is 100.
+// 3. Math.random() * 100 gives a decimal in the range [0, 100).
+// 4. Math.floor() rounds down to the nearest integer, giving a value in the range [0, 99].
+// 5. Adding minimum (1) shifts the range to [1, 100].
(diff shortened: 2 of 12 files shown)


Sprint-1/1-key-exercises/1-count.js
```diff
@@ -4,3 +4,4 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+// The code adds 1 to the count variable.
```

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -7,3 +7,6 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+console.log(num);
+// num represents a random integer between 1 and 100.
```

We can also use the concise and precise ***interval notation*** to describe a range of values. 
- `[`, `]` => inclusion
- `(`, `)` => exclusion

Sprint-1/2-mandatory-errors/3.js
```diff
@@ -1,4 +1,4 @@
-const cardNumber = 4533787178994213;
+const cardNumber = "4533787178994213";
 const last4Digits = cardNumber.slice(-4);
```

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -12,11 +11,12 @@ console.log(`The percentage change is ${percentageChange}`);
 // Read the code and then answer the questions below
 
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
-
+// There are 5 function calls: Number() twice, replaceAll() twice, and console.log().
 // b) Run the code and identify the line where the error is coming from - why is this error occurring? How can you fix this problem?
-
+// replaceAll might not work in older Node versions
```
cjyuan: Can you describe why the original code on line 5 does not work? Here is the code:
`priceAfterOneYear = Number(priceAfterOneYear.replaceAll("," ""));`

Note: There is an easier way to fix the original code than replacing `replaceAll()` by `replace()`.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -17,11 +17,22 @@ const pence = paddedPenceNumberString
...
+// It removes the last character ('p') from the string to keep only the numbers.
+
+// 3. const paddedPenceNumberString = ...
+// It ensures the string has at least 3 characters by adding a "0" at the beginning if needed (e.g., "50" becomes "050").
+
+// 4. const pounds = ...
+// It extracts the pounds part by taking all characters except the last two.
+
+// 5. const pence = ...
+// It extracts the last two characters to represent the pence part and ensures it has two digits.
```
cjyuan: Could we expect this program to work as intended for any valid `penceString` if we deleted `.padEnd(2, "0")` from the code?

cjyuan (commented):
Code is quite well written. I mainly have some questions and suggestions.

mahmoudshaabo1984:
Hi @cjyuan,

Thank you for reviewing my code and for your detailed feedback. I've updated all five files based on your suggestions:

1. **1-count.js**: Updated the comment to use the correct programming term — this is an **increment** operation.

2. **4-random.js**: Replaced "between 1 and 100" with proper interval notation: **[1, 100]**. I also added a step-by-step breakdown showing how the expression evaluates to this range.

3. **3.js (mandatory-errors)**: Reverted `cardNumber` back to a number (without quotes) and used `String(cardNumber).slice(-4)` for type conversion instead of modifying the original value.

4. **1-percentage-change.js**: Corrected my explanation — the real error was a **missing comma** between the two arguments in `replaceAll("," "")`. The fix is simply adding the comma: `replaceAll(",", "")`.

5. **3-to-pounds.js**: After testing with different inputs like "5p", "50p", and "399p", I concluded that `.padEnd(2, "0")` is actually **not needed** in this script because `.padStart(3, "0")` already guarantees at least 3 characters, so `.substring()` will always extract exactly 2 characters for the pence part.

Please let me know if there is anything else I should improve. I appreciate the feedback — it helped me understand these concepts more deeply.

Best regards,
Mahmoud

cjyuan:
Thanks for answering all my questions, and they are all spot on! Excellent job!

cjyuan:
Sorry, I beat you on the label. :P

illicitonion:

PR #984, opened 2026-02-18
London | 26-ITP-JAN | Shuheda Begum | Sprint 2 | Structuring and Testing Data

codebyshay:
Self checklist

Changelist

i have fixed errors on 0.js file.


Sprint-2/2-mandatory-debug/2.js
```diff
@@ -1,24 +1,43 @@
 // Predict and explain first...
 
 // Predict the output of the following code:
-// =============> Write your prediction here
+// Write your prediction here
+// I think the output will be 'The last digit of 42 is 2', 'The last digit of 105 is 5', and 'The last digit of 806 is 6'.
+// I think this will happen because the getLastDigit function is supposed to take a number, convert it to a string, and then return the last charecter of that string, which should be the last digit. 
 
-const num = 103;
+//const num = 103;
```
cjyuan: Note: Deleting the global `num` is optional. If we declared a parameter that's also named `num`, then within the function block, the identifier `num` is resolved to the parameter `num`.

If you are interested in the topic, you can looking up these two concepts, ***identifier scope*** and ***identifier resolution***, in the context of JavaScript programming. ChatGPT can give a good explanation.

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -14,6 +14,30 @@
...
+
+// I need to:
+// 1. Square the height
+// 2. Divide weight by squared height
+// 3. Round the result to 1 decimal place
+// 4. Return the result 
+
+function calculateBMI(weight, height) {
+    const bmi = weight / (height * height);
+    return bmi.toFixed(1);
```

```javascript
  
```

cjyuan (commented):
Code looks good. Just have a question.

cjyuan:
The Cangelist section of the PR description does not quite describe what you have changed. Can you update it accordingly to better reflect what you have changed on this PR branch?

PR #990, opened 2026-02-19
London | 26-Jan-ITP | Angela McLeary | Sprint 1|Coursework/sprint 1

AngelaMcLeary:
## Learners, PR Template

Self checklist

## Changelist

This PR is focused on understanding the existing codebase and deepening my knowledge of JavaScript concepts.
It involves interpreting the current implementation, analyzing how different parts interact, identifying and understanding errors, and using Chrome DevTools to inspect runtime behavior and debug issues.

## Questions

What is the difference between a function and a methods?


Sprint-1/1-key-exercises/4-random.js
```diff
@@ -4,6 +4,23 @@ const maximum = 100;
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 
 // In this exercise, you will need to work out what num represents?
+//*Answer:
+//Number returns a random whole number between 1 and 100.
```
LonMcGregor: Can you figure out the limits more precisely? Think about if it includes or excludes those numbers?

Sprint-1/2-mandatory-errors/4.js
```diff
@@ -1,2 +1,11 @@
...
+//const 12HourClockTime = "20:53";
+//const 24hourClockTime = "08:53";
+
+//What's wrong with this code?
+//*Answer
+//The variable is not named in line with the Javascript naming convention. 
+// When declaring a variable it must not start with a number, space or reserved word.
+// to fix this, we can rename them correctly.
+
+const twelve_HourClockTime = "20:53";
```
LonMcGregor: here you have mixed variable naming using snake_case and PascalCase. Can you see the difference?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,37 @@ console.log(result);
...
+//  movieDuration
+
+// f) Try experimenting with different values of movieLength. 
+// Will this code work for all values of movieLength? Explain your answer,
+// *Answer
+// I changed the value to;
+// 1200 returned 0:20:00
+// 92383.7 returned 25:39:43.69999999999709
+// -8784 returned -2:-26:-24
+// it worked with all numbers including decimals and negative numbers.
```
LonMcGregor: Just note that when you tested these, though it ran without errors, does the output look OK, or are there values where it looks incorrect?

LonMcGregor (commented):
Good work on these tasks.

To figure out the difference between a function and a method,you need to consider objects (you may or may not have come across these yet in your study). Functions are written by you, methods are written as part of objects. This is a good site with an exaplantion: https://www.geeksforgeeks.org/javascript/difference-between-methods-and-functions-in-javascript/
Does that help you understand the difference?

Sprint-1/1-key-exercises/4-random.js
AngelaMcLeary: @LonMcGregor  Thanks for your question.
I can see it is not very clear. "num" returns a random whole number between 1 and 100 **inclusive**. It can never go below 1 or over 100.

Sprint-1/2-mandatory-errors/4.js
AngelaMcLeary: @LonMcGregor thanks for your feedback. 
I have mixed them up, I see that I should use camelCase for functions and variables. I will update the code.
Thank you
Angela.

Sprint-1/3-mandatory-interpret/2-time-format.js
AngelaMcLeary: @LonMcGregor  another good question. 
Line 46: Milliseconds appeared in the output because they were part of the input and were not explicitly removed, even though they were not required and it does make the output not what is expected.

Line 47: Negative numbers passed the test. In reality there should not be negative time in the movieLength and it should throw an error.

In both cases there was no validation to stop this from occurring.

Thank you
Angela.

LonMcGregor (commented):
Good work, complete!

illicitonion:

PR #998, opened 2026-02-20
London | 26-ITP-January | Eugenie Ahangama | Sprint 3 | implement and rewrite

Eugenie-A:
## Learners, PR Template

Self checklist

## Changelist

Completed all the exercises in implement and rewrite with jest directory.

## Questions

What are the most common mistakes beginners make with Jest?

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -12,6 +12,13 @@
 
 function isProperFraction(numerator, denominator) {
   // TODO: Implement this function
+  if (denominator <= 0) {
+    return false; // not a valid fraction
+  }
+  if (numerator < denominator && numerator >= 0) {
+    return true;
+  }
+  return false;
 }
 
 // The line below allows us to load the isProperFraction function into tests in other files.
@@ -31,3 +38,42 @@ function assertEquals(actualOutput, targetOutput) {
 
 // Example: 1/2 is a proper fraction
 assertEquals(isProperFraction(1, 2), true);
+
+// Example: 3/5 is a postive proper fraction
+const properFraction = isProperFraction(3, 5);
+assertEquals(properFraction, true);
+
+// Example: 0/4 is a proper fraction with zero numerator
+// Zero numerator is allowed in proper fractions
+const zeroNumerator = isProperFraction(0, 4);
+assertEquals(zeroNumerator, true);
+
+// Example: 7/7 is an improper fraction,
+// numerator equals denominator
+const equalFraction = isProperFraction(7, 7);
+assertEquals(equalFraction, false);
+
+// Example: 8/3 is an improper fraction,
+// numerator is larger
+const numeratorLarger = isProperFraction(8, 3);
+assertEquals(numeratorLarger, false);
+
+// Example: 5/0 is an invalid fraction
+// Division by zero is not a valid fraction
+const zeroDenominator = isProperFraction(5, 0);
+assertEquals(zeroDenominator, false);
+
+// Example: 2/-6 is an invalid fraction
+// Negative denominator makes fraction invalid
+const negativeDenominator = isProperFraction(2, -6);
+assertEquals(negativeDenominator, false);
+
+// Example: -1/4 is an invalid fraction
+// Negative numerator makes fraction negative
+const negativeNumerator = isProperFraction(-1, 4);
+assertEquals(negativeNumerator, false);
+
+// Example: 1/100 is a positive fraction
+// Small numerator with large denominator
+const smallFraction = isProperFraction(1, 100);
+assertEquals(smallFraction, true);
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -11,14 +11,13 @@
 // execute the code to ensure all tests pass.
 
 function isProperFraction(numerator, denominator) {
-  // TODO: Implement this function
-  if (denominator <= 0) {
-    return false; // not a valid fraction
+  // Denominator cannot be zero
+  if (denominator === 0) {
+    return false; 
   }
-  if (numerator < denominator && numerator >= 0) {
-    return true;
-  }
-  return false;
+  // A proper fraction satisfies: numerator < denominator 
+  // This correctly handles all sign combinations
+  return Math.abs(numerator) < Math.abs(denominator);
 }
 
 // The line below allows us to load the isProperFraction function into tests in other files.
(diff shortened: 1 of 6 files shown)


Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -12,6 +12,13 @@
 
 function isProperFraction(numerator, denominator) {
   // TODO: Implement this function
+  if (denominator <= 0) {
+    return false; // not a valid fraction
+  }
```
cjyuan: In math, 2/-5 has the same value as -(2/5), so we could also consider 2/-5 or even -2/-5 as proper fraction.

Can you update your implementation to also accept negative denominator (and then also update your test accordingly)?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -23,6 +23,31 @@
...
+  // Special case: Ace is worth 11
+  if (rank === 'A') return 11;
+  // Special case: Face cards (Jack, Queen, King) are worth 10
+  if (rank === 'J' || rank === 'Q' || rank === 'K') return 10;
+  // For number cards (2-10): converts rank string into number
+  const value = Number(rank);
+  // Makes sure it's actually a valid number between 2 and 10
+  if (isNaN(value) || value < 2 || value > 10) {
+    throw new Error('Invalid rank');
+  }
```
cjyuan: Does your function return the value you expected from each of the following function call?

```
getCardValue("2.♠");
```

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -4,7 +4,47 @@ const isProperFraction = require("../implement/2-is-proper-fraction");
...
-// Special case: numerator is zero
+// Special case: denominator is zero
 test(`should return false when denominator is zero`, () => {
   expect(isProperFraction(1, 0)).toEqual(false);
 });
+
+// Special case: numerator is zero
+test(`should return true when numerator is zero`, () => {
+  expect(isProperFraction(0, 4)).toEqual(true);
+});
```
cjyuan: 0/0 is not a proper fraction. Can you rephrase the test description?

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -4,7 +4,47 @@ const isProperFraction = require("../implement/2-is-proper-fraction");
...
+
+// Special case: numerator is larger than denominator
+test(`should return false when numerator is larger than denominator`, () => {
+  expect(isProperFraction(8, 3)).toEqual(false);
+});
+
+// Special case: negative denominator
+test(`should return false when denominator is negative`, () => {
+  expect(isProperFraction(2, -6)).toEqual(false);
+});
```
cjyuan: 2/-6 should be considered a proper fraction.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -14,7 +14,35 @@ test(`Should return 11 when given an ace card`, () => {
 //   Face Cards (J, Q, K)
 //   Invalid Cards
 
+// Case 2: Number Cards (2-10)
+test(`Should return correct value for single digit number card`, () => {
```
cjyuan: 10 is not single digit. Perhaps we don't need to specify number of digits?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
Eugenie-A: Yes, `getCardValue("2♠")` returns 2 as expected.
The rank `"2"` is correctly parsed to the number 2, passes the validation check
`isNaN(value) || value < 2 || value > 10` is false, and the function returns that value.

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
Eugenie-A: Updated to use `Math.abs(numerator) < Math.abs(denominator)` so negative denominators and numerators are accepted when the fraction is smaller than 1 in size.
Tests updated → all passing. Matches the equivalence you pointed out (2/-5 = -2/5).

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
Eugenie-A: Thank you for the feedback.
I've rephrased the test description to: 
`should return true when numerator is zero and denominator is non-zero`
This makes it clear we're only asserting true for valid denominators.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
Eugenie-A: Already fixed, changed to expect true for 2/-6 and -1/4 using Math.abs comparison.
Matches the equivalence 2/-6 = -1/4 < 1 in absolute value.
Thanks!

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
Eugenie-A: Yep, fixed, I removed "single digit" since 10 is included.
Now just: `Should return correct value for number cards (2-10)`
Thanks!

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
cjyuan: I mean, `getCardValue("2.♠")` (With a dot between 2 and the ♠)
or
`getCardValue("+2♠")`

They both return `2` but do you want to consider "+2" or "2." as valid ranks?

cjyuan:
Just the card rank case needs follow up. Other changes look good.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
Eugenie-A: No, I don't want "+2" or "2." to be considered valid ranks, they should throw 'Invalid rank'.
Made the rank check strict with exact string matching. 
Now those throw as expected.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -40,14 +40,13 @@ function getCardValue(card) {
...
+  // Reject anything with extra characters (dots, +, spaces, leading zeros, etc.)
+  const allowedNumberRanks = ["2", "3", "4", "5", "6", "7", "8", "9", "10"];
+  if (!allowedNumberRanks.includes(rank)) {
     throw new Error('Invalid rank');
-  }
-  // Return the numeric value for 2-10
-  return value;
+}
+return Number(rank);
 }
```
cjyuan: Indentation is off.

Consider install **prettier** VSCode extension and enable formatting on save/paste on VSCode.
https://github.com/Eugenie-A/Module-Structuring-and-Testing-Data/blob/main/readme.md

cjyuan (commented):
Changes look good. Well done.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
Eugenie-A: Will do. Thank you very much for the feedback!

Eugenie-A: Thanks for the review!
Fixed the indentation around `return Number(rank);`,  it was misaligned.
Installed Prettier extension and enabled format-on-save as suggested in the readme.
Code should look clean now. Let me know if anything else needs adjusting.

illicitonion:

PR #1005, opened 2026-02-21
Glasgow | Jan-26 | Elisabeth Matulian | Sprint 3 | Coursework practice tdd

Elisabeth-Matulian:
## Learners, PR Template

Self checklist

- [x]  I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title

## Changelist

Completed Coursework practice tdd tasks

## Questions

No questions

### DIFF
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -16,6 +16,12 @@ test("should count multiple occurrences of a character", () => {
   const count = countChar(str, char);
   expect(count).toEqual(5);
 });
+test("No Occurrences", () => {
+  const str = "qwerty";
+  const char = "k";
+  const count = countChar(str, char);
+  expect(count).toEqual(0);
+})
 
 // Scenario: No Occurrences
 // Given the input string `str`,
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -16,15 +16,16 @@ test("should count multiple occurrences of a character", () => {
   const count = countChar(str, char);
   expect(count).toEqual(5);
 });
-test("No Occurrences", () => {
-  const str = "qwerty";
-  const char = "k";
-  const count = countChar(str, char);
-  expect(count).toEqual(0);
-})
 
 // Scenario: No Occurrences
 // Given the input string `str`,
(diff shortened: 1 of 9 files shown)


Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -16,6 +16,12 @@ test("should count multiple occurrences of a character", () => {
   const count = countChar(str, char);
   expect(count).toEqual(5);
 });
+test("No Occurrences", () => {
```
hkavalikas: Functionality-wise, the test is correct. But it is a tiny bit misplaced in the file, and the name could be a bit more descriptive, based on other tests in this file. Can you try to update it?

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,6 @@
-function repeatStr() {
-  return "hellohellohello";
+function repeatStr(str, count) {
+  if (count >= 0) {return str.repeat(count)}
```
hkavalikas: One little thing, for readability purposes, it's worth putting return and throw on their own lines with braces.

Sprint-3/2-practice-tdd/repeat-str.test.js
hkavalikas, line 1: Functionality-wise, the tests are correct, but it's going to be hard to distinguish which one fails, given that they are all named the same. How can we improve that?

hkavalikas (changes requested):

Sprint-3/2-practice-tdd/count.test.js
Elisabeth-Matulian: the peace of code is moved, name is updated

Sprint-3/2-practice-tdd/repeat-str.js
Elisabeth-Matulian: Thanks for mentioning, I have changed the position on the line

Sprint-3/2-practice-tdd/repeat-str.test.js
Elisabeth-Matulian: That's true! Didn't think about that. Names's updated:)

Could you please take another look when you have a moment?

illicitonion:

PR #1007, opened 2026-02-21
Cape Town | 2026-ITP-Jan | Pretty Taruvinga | Sprint 2 | Structuring And Testing Data 

Pretty548:
## Learners, PR Template

Self checklist

## Changelist

In this sprint I worked on JavaScript challenges focused on functions, arrays, and string manipulation. I wrote pure functions that return values so they can be tested and used unit tests to confirm my solutions work correctly.

I practiced key concepts like variable scope, avoiding variable shadowing, and using methods such as padStart for formatting. One issue I faced was redeclaring a variable inside a function, which I fixed by using the parameter instead.

This sprint improved my problem-solving skills and my confidence in debugging and writing cleaner, more readable code.

### DIFF
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -15,6 +15,6 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    const bmi = weight / (height * height);
-    return Math.round(bmi * 10) / 10;
+  const bmi = weight / (height * height);
+  return Number(bmi.toFixed(1));
 }
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -15,6 +15,6 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-  const bmi = weight / (height * height);
-  return Number(bmi.toFixed(1));
+    const bmi = weight / (height * height);
+    return Math.round(bmi * 10) / 10;
 }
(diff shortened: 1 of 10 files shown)


Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,6 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
+    const bmi = weight / (height * height);
+    return Math.round(bmi * 10) / 10;
```
LonMcGregor: Why do you multiply by 10 and then divide by 10? Is there an easier way to get 1 decimal place?

LonMcGregor (commented):
Good work, just one function you could improve a bit further

Sprint-2/3-mandatory-implement/1-bmi.js
Pretty548: I multiplied by 10 and divided by 10 to round the BMI to one decimal place using Math.round(), since it only rounds whole numbers.
Alternatively, I could use Number(bmi.toFixed(1)) which is a bit cleaner and more readable.

LonMcGregor: Good suggestion - I think toFixed makes more sense here. Can you make that change in the code?

Pretty548: Thanks for the feedback! I've updated the function to use `toFixed(1)` for rounding to one decimal place.

LonMcGregor (commented):
Good work!

illicitonion:

PR #1010, opened 2026-02-21
London | 26-ITP-Jan | Alex Okorefe | Sprint 1 | Coursework

Alex-Os-Dev-Lab:
## Learners, PR Template

Self checklist

## Changelist

Completed the JavaScript tasks

## Questions

No questions for now

### DIFF
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -3,7 +3,32 @@ const maximum = 100;
 
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 
+console.log(num);
+
 // In this exercise, you will need to work out what num represents?
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+// when first evaluating the full num expression i got the answer 99
+// The second time i got 50 (which means .floor and .random are doing something to change the value of the expression each run)
+
+// The order of operations is as follows:
+
+// 1. The expression inside the parentheses is evaluated first: (maximum - minimum + 1) which gives us 100 - 1 + 1 = 100
+// 2. Then Math.random() is called, which generates a random decimal number between 0 and 1.
+// 3. The result of Math.random() is multiplied by the result of step 1, giving us a random decimal number between 0 and 100.
+// 4. Math.floor() is then called on the result of step 3, which rounds it down to the nearest whole number, giving us a random integer between 0 and 99.
+// 5. Finally, the minimum value (1) is added to the result of step 4, giving us a random integer between 1 and 100.
+
+// When num = 99
+// math.random gives us 0.98 (or something close to it)
+// 0.98 * 100 = 98
+// math.floor(98) = 98
+// 98 + 1 = 99
+
+// When num = 50
+// math.random gives us 0.49 (or something close to it)
+// 0.49 * 100 = 49
+// math.floor(49) = 49
+// 49 + 1 = 50
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -16,10 +16,11 @@ console.log(num);
 // The order of operations is as follows:
 
 // 1. The expression inside the parentheses is evaluated first: (maximum - minimum + 1) which gives us 100 - 1 + 1 = 100
-// 2. Then Math.random() is called, which generates a random decimal number between 0 and 1.
-// 3. The result of Math.random() is multiplied by the result of step 1, giving us a random decimal number between 0 and 100.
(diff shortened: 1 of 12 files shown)


Sprint-1/1-key-exercises/4-random.js
```diff
@@ -3,7 +3,32 @@ const maximum = 100;
...
+// when first evaluating the full num expression i got the answer 99
+// The second time i got 50 (which means .floor and .random are doing something to change the value of the expression each run)
+
+// The order of operations is as follows:
+
+// 1. The expression inside the parentheses is evaluated first: (maximum - minimum + 1) which gives us 100 - 1 + 1 = 100
+// 2. Then Math.random() is called, which generates a random decimal number between 0 and 1.
+// 3. The result of Math.random() is multiplied by the result of step 1, giving us a random decimal number between 0 and 100.
+// 4. Math.floor() is then called on the result of step 3, which rounds it down to the nearest whole number, giving us a random integer between 0 and 99.
+// 5. Finally, the minimum value (1) is added to the result of step 4, giving us a random integer between 1 and 100.
```

We can use the concise and precise ***interval notation*** to describe a range of values. 
- `[`, `]` => inclusion
- `(`, `)` => exclusion

Sprint-1/2-mandatory-errors/3.js
```diff
@@ -1,9 +1,19 @@
-const cardNumber = 4533787178994213;
+const cardNumber = "4533787178994213";
 const last4Digits = cardNumber.slice(-4);
```

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -13,10 +13,33 @@ console.log(`The percentage change is ${percentageChange}`);
 
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
 
+// Function calls are determined by the presence of parentheses "()" after a function name.
+// Therefore in this file, there are 6 function calls:
```
cjyuan: Which 6? There should be less than 6 function calls in the code.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -25,3 +25,98 @@ console.log(`£${pounds}.${pence}`);
...
+//    to the end of the string until it reaches a total length of 2 characters.
+//    In this case, since paddedPenceNumberString is "399" which has a length of 3 characters,
+//    the substring method will extract the last two characters "99"
+//    and then the padEnd method will not add any padding since the length of the resulting string is already 2 characters.
+//    therefore pence will be "99"
+
+// g. if penceString = "9p" ; const penceStringWithoutTrailingP = "9" ; const paddedPenceNumberString = "009" ;
+//    const pounds = "0" ; const pence = "09" because the substring method will extract the last two characters "09"
+//    and then the padEnd method will not add any padding since the length of the resulting string is already 2 characters.
+//    therefore pence will be "09"
```
cjyuan: Very thorough analysis.

Could we expect this program to work as intended for any valid `penceString` if we deleted `.padEnd(2, "0")` from the code? In other words, do we really need `.padEnd(2, "0")` in this script?

Sprint-1/2-mandatory-errors/3.js
Alex-Os-Dev-Lab: Hey CJ thanks for pointing this out. I believe I have now solved for this.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
Alex-Os-Dev-Lab: I was also including method calls. thanks for flagging. I have updated the comments with the correct answer

Sprint-1/3-mandatory-interpret/3-to-pounds.js
Alex-Os-Dev-Lab: Hey CJ I see where you're coming from. Since paddedPenceNumberString is guaranteed to be at least 3 characters long due to the padStart(3, "0") call in step 3, the substring(length - 2) will always return exactly two characters. Therefore, .padEnd(2, "0") is redundant I will remove it and make the code more efficient.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
cjyuan: Method is a kind of function and method calls should also be counted, but the total number of function calls (including method calls) is not 6.

cjyuan:
Changes look good. Can you clarify the number of function+method calls in https://github.com/CodeYourFuture/Module-Structuring-and-Testing-Data/pull/1010#discussion_r2870604184 ?

And if a PR is ready to be re-reviewed, you should also add the "Needs Review" label to it.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
Alex-Os-Dev-Lab: I have updated this and corrected it. there are 5 function/method calls in the file

cjyuan:
LGTM.

illicitonion:

PR #1020, opened 2026-02-22
London | 25-ITP-May | Kayanat Suleman | Sprint 3 | Implement & Rewrite Tests

KayanatSuleman:
## Learners, PR Template

Self checklist

## Changelist

Implemented getAngleType, isProperFraction, and getCardValue solutions.

Rewrote tests using Jest to cover expected cases, including boundary/invalid inputs.

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -11,7 +11,8 @@
 // execute the code to ensure all tests pass.
 
 function isProperFraction(numerator, denominator) {
-  // TODO: Implement this function
+  if (denominator === 0) return false;
+  return Math.abs(numerator) < Math.abs(denominator);
 }
 
 // The line below allows us to load the isProperFraction function into tests in other files.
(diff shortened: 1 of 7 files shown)


Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -11,7 +11,8 @@
 // execute the code to ensure all tests pass.
 
 function isProperFraction(numerator, denominator) {
-  // TODO: Implement this function
+  if (denominator === 0) return false;
+  return Math.abs(numerator) < Math.abs(denominator);
```
LonMcGregor, line 15: If you are using math.abs here, is there ever going to be a situation where a valid proper fraction could have a numerator less than zero?
Do you need the initial check for the zero denominator?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -41,12 +62,38 @@ function assertEquals(actualOutput, targetOutput) {
...
+
 // Handling invalid cards
 try {
   getCardValue("invalid");
 
   // This line will not be reached if an error is thrown as expected
   console.error("Error was not thrown for invalid card");
-} catch (e) {}
+} catch (e) {
+  console.log("Correctly threw error for invalid card");
```
LonMcGregor: Do you see when you run this test, if it passes it will always log this output? Do any of the other tests use a console.log when they pass without any failures?

LonMcGregor (commented):
Your solutions are all working and good, but I have some suggestions for you to think about which could improve them further

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
KayanatSuleman: Thank you for your comment - For getCardValue, I've removed the console.log from passing tests and rewrote the invalid card tests using jests toThrow() matcher so that passing tests don't log output unnecessarily.

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
KayanatSuleman: Hi @LonMcGregor thank you for your review. For isProperFraction, I used Maths.abs() to correctly handle negative numerators and denominators (e.g. 1/-3 and -2/-5) since proper fractions are defined by comparing magnitudes rather than sign.  You are probably right - that the zero check could be simplified, but it is necessary because mathematically, division by zero is undefined and using this in my code improves clarity by explicitly handling it.

LonMcGregor:
Good explanations for why you wrote the fractions function that way - thanks for updating the console.logs. This sprint task is complete now.

illicitonion:

PR #1022, opened 2026-02-22
London | 25-ITP-May | Kayanat Suleman | Sprint 3 | Dead Code

KayanatSuleman:
## Learners, PR Template

Self checklist

## Changelist

- Removed unreachable/redundant code in Sprint-3/3-dead-code/exercise-1.js (removed unused variables + unreachable log).
- Removed redundant intermediate variables/functions in Sprint-3/3-dead-code/exercise-2.js while keeping behaviour the same.

### DIFF
--- a/Sprint-3/3-dead-code/exercise-1.js
+++ b/Sprint-3/3-dead-code/exercise-1.js
@@ -6,7 +6,7 @@ function sayHello(greeting, name) {
   return `${greeting}, ${name}!`;
 }
 
-testName = "Aman";
+const testName = "Aman";
 
 const greetingMessage = sayHello(greeting, testName);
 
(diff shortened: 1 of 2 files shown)


Sprint-3/3-dead-code/exercise-1.js
```diff
@@ -1,13 +1,9 @@
 // Find the instances of unreachable and redundant code - remove them!
 // The sayHello function should continue to work for any reasonable input it's given.
-
-let testName = "Jerry";
```
hkavalikas, line 4: While this works, it does make testName a global variable now, which could result in unintended behaviour and bugs, especially for multi-file projects. How can we fix that?

hkavalikas (changes requested):

Sprint-3/3-dead-code/exercise-1.js
KayanatSuleman: Hi @hkavalikas  thank you for pointing that out. 'testName' was being created as a global variable, therefore I have now declared it using 'const' to avoid polluting the global scope and to ensure safer behaviour in multi-file projects.

hkavalikas (approved):

hkavalikas:
Awesome work

illicitonion:

PR #1025, opened 2026-02-23
Glasgow | 26-ITP-Jan| Martin McLean |Sprint 3| practice TDD coursework

mjm-git185:
Changelist

 I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
 My changes meet the requirements of the task
 I have tested my changes
 My changes follow the style guide
my coursework


Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,14 @@
 function countChar(stringOfCharacters, findCharacter) {
-  return 5
+totalCount=stringOfCharacters.split("");
```
LonMcGregor: What kind of variables are these intended to be? js lets you declare variables like this, but it is good practice to be a bit more specific.

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,8 @@
-function repeatStr() {
-  return "hellohellohello";
-}
+function repeatStr(times, str) {
+  if (times < 0){ throw ("error")}
```
LonMcGregor: If you ran this program and it just said "error" would that be enough info for you to figure out and solve the problem? Could this message be more specific?

Sprint-3/2-practice-tdd/count.js
LonMcGregor, line 1: Bare careful about being consistent with code style here. Notably capitalisations and line endings.

LonMcGregor (commented):
Good start on this task, there are a few things that you need to take a closer look at.

LonMcGregor (commented):
Almost done here. Just be careful about countChar - did you fully test everything after making changes? When you declare `Count` initially it is still capitalised, and you didn't declare it as either a `const` or a `let`

mjm-git185:
i have now changed them all to lower case and declared the count veritable as a let

LonMcGregor (commented):
Good work on this task

illicitonion:

PR #1029, opened 2026-02-23
London | 26-ITP-May | Oussama Mouggal | Sprint 3 | coursework-practice-tdd

Oussama-Mouggal:
Self checklist

## Changelist

tested edge cases and rewirted functions to pass the npm test


Sprint-3/2-practice-tdd/get-ordinal-number.test.js
Edu-Vin, line 1: I believe there is still room for improvement for this file. You didn't test for the number ending with 11. Also, for the other test cases, there should be a test to verify that the except values will not have the appended string example. using 12 should return 12nd. If you still want to experiment, you can change your except valuese from those ending with  12, 13 to 22 and 33.

Sprint-3/2-practice-tdd/repeat-str.test.js
```diff
@@ -15,7 +15,28 @@ test("should repeat the string count times", () => {
...
+  const count = 0;
+  const repeatedStr = repeatStr(str, count);
+  expect(repeatedStr).toEqual("");
+});
 
+// Case: Handle negative count:
+test("should throw an error when count is negative", () => {
+  const str = "hello";
+  const count = -1;
+  expect(() => repeatStr(str, count)).toThrow();
```
Edu-Vin: To take this further, why not also check if the expected message is returned?

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
Oussama-Mouggal: Thank you @Edu-Vin for your valuable feedback and remarks. I added more cases to my test/code to test all posibilities.

Sprint-3/2-practice-tdd/repeat-str.test.js
Oussama-Mouggal: Thank you for your message. I added a comment which will return in case of an error message.

Edu-Vin (approved):

illicitonion:

PR #1030, opened 2026-02-23
London | 26-ITP-January | Khaliun Baatarkhuu | Sprint 1 | Coursework. 

khaliun-dev:
## Self checklist

## Changelist

Debugged errors, explained some of the answers i have provided with reasoning, and implemented functions. 

## Questions

None at the moment.

### DIFF
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -1,6 +1,7 @@
 let count = 0;
 
-count = count + 1;
+count = count + 1; 
+// line 3 is assigning the variable "count" a new value of "count +1", then storing it back into "count".
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -12,12 +12,13 @@
 const filePath = "/Users/mitch/cyf/Module-JS1/week-1/interpret/file.txt";
 const lastSlashIndex = filePath.lastIndexOf("/");
 const base = filePath.slice(lastSlashIndex + 1);
+
 console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
 
-const dir = ;
-const ext = ;
+const dir = filePath.slice(0,lastSlashIndex);
+const ext = filePath.slice(lastDotIndex+1);
 
 // https://www.google.com/search?q=slice+mdn
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -2,6 +2,11 @@ const minimum = 1;
 const maximum = 100;
 
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
+//num can be any generated random number between 1 and 100, rounded down.
+
+
+
+
 
 // In this exercise, you will need to work out what num represents?
 // Try breaking down the expression and using documentation to explain what it means
--- a/Sprint-1/3-mandatory-interpret/1-percentage-change.js
+++ b/Sprint-1/3-mandatory-interpret/1-percentage-change.js
@@ -1,8 +1,8 @@
 let carPrice = "10,000";
 let priceAfterOneYear = "8,543";
 
-carPrice = Number(carPrice.replaceAll(",", ""));
-priceAfterOneYear = Number(priceAfterOneYear.replaceAll("," ""));
+carPrice = Number(carPrice.replaceAll(",",""));
+priceAfterOneYear = Number(priceAfterOneYear.replaceAll(",",""));
 
 const priceDifference = carPrice - priceAfterOneYear;
 const percentageChange = (priceDifference / carPrice) * 100;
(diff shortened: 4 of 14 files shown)


Sprint-1/1-key-exercises/1-count.js
```diff
@@ -1,6 +1,7 @@
 let count = 0;
 
-count = count + 1;
+count = count + 1; 
+// line 3 is assigning the variable "count" a new value of "count +1", then storing it back into "count".
```
cjyuan, line 4: Operation like `count = count + 1` is very common in programming, and there is a programming term describing such operation.

Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -12,12 +12,13 @@
...
+
 console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
 
-const dir = ;
-const ext = ;
+const dir = filePath.slice(0,lastSlashIndex);
+const ext = filePath.slice(lastDotIndex+1);
```
cjyuan, line 22: What is `lastDotIndex`?

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -2,6 +2,11 @@ const minimum = 1;
 const maximum = 100;
 
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
+//num can be any generated random number between 1 and 100, rounded down.
```
cjyuan, line 5: Can you give a precise description what each of these expressions does, and the range of the numbers it may produce?

1. `Math.random()`

---
Note: To describe a range of numbers, we can use the concise and precise ***interval notation***: 
- `[`, `]` => inclusion
- `(`, `)` => exclusion

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -13,10 +13,27 @@ console.log(`The percentage change is ${percentageChange}`);
...
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
 
+//There are three function calls: replaceAll() - used twice, number() - used twice, console.log() - used once.  
+// carPrice = Number(carPrice.replaceAll("," , ""));
+// priceAfterOneYear = Number(priceAfterOneYear.replaceAll("," , ""));
+// console.log(`The percentage change is ${percentageChange}`);
+
 // b) Run the code and identify the line where the error is coming from - why is this error occurring? How can you fix this problem?
 
+//There was a syntax error on line five, missing comma. Fixed by adding one.
```
cjyuan, line 23: We can more precisely describe "A comma is missing between `","` and `""` in the function call" as:  

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -1,27 +1,27 @@
...
   0,
   paddedPenceNumberString.length - 2
 );
-
+//4. We are extracting the pounds now, ie the first character. 
 const pence = paddedPenceNumberString
   .substring(paddedPenceNumberString.length - 2)
   .padEnd(2, "0");
-
+//4. Here, we are extracting the last two digits, ie pence. Padend insures there are two digits and adds a "0" to compensate, if necessary.
```
cjyuan, line 18: Could we expect this program to work as intended for any valid `penceString` if we deleted `.padEnd(2, "0")` from the code?

Sprint-1/4-stretch-explore/chrome.md
```diff
@@ -12,7 +12,16 @@ invoke the function `alert` with an input string of `"Hello world!"`;
...
 
 What effect does calling the `prompt` function have?
+
+2. It displays a pop-up box with What's your name written on it and an empty dialog box inside, awaiting my input.
+Also, there are two options of OK and Cancel. 
+
 What is the return value of `prompt`?
+
+3. If I enter my name undefined appears, same if i press the cancel button. Fixed this by adding myName at the end.
+Now the value when i enter my name shows "my name", if i press cancel it shows "null".
```
cjyuan, line 27: What you observed were the output produced by `console.log()`

When a value is output on the console, it is converted to a string first. As such, just by looking at the output, we could not tell the actual type of the value being returned by the function. For examples, in both of the following actions, `console.log()` would output "null" on the screen.
- A user entered "null" in the dialog box and then clicked "OK.
- A user entered nothing in the dialog box and then clicked "Cancel"

Can you look up the `Window.prompt()` function at MDN Web Docs and find out what this function returns?

Sprint-1/1-key-exercises/1-count.js
khaliun-dev: Increment.

Sprint-1/1-key-exercises/3-paths.js
khaliun-dev: const ext = base.slice(base.lastIndexOf(".") + 1);

Sprint-1/1-key-exercises/4-random.js
khaliun-dev: - 1 Math.random() , this expression generates a random number between 0 and 1, can be equal to 0 but never exactly 1. 
- 2 Math.random() * (maximum - minimum + 1) , this takes the randomly generated number and scales it out to our given range of 1 and 100, but does not equal 100.
- 3 Math.floor(Math.random() * (maximum - minimum + 1)) , Math.floor rounds down the number generated.
- 4 Math.floor(Math.random() * (maximum - minimum + 1)) + minimum , we are adding the minimum so that the rounded down number can reach the maximum of 100.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
khaliun-dev: arguments

Sprint-1/3-mandatory-interpret/3-to-pounds.js
khaliun-dev: If we delete .padEnd(2, "0") the code will still run, because we are guaranteeing three digits with padStart(3, "0") and then we are extracting the last two with substring(paddedPenceNumberString.length - 2). Since that is the function of  .padEnd(2, "0"), we don't need it in actuallity.

Sprint-1/4-stretch-explore/chrome.md
khaliun-dev: Window.prompt() returns a string containing the text entered by the user, or null.

khaliun-dev: @cjyuan Good morning, 

Thank you for your hard work reviewing everything and your feedback. I think, i have completed and understood all the points you have raised. 

Could you have another look, please?

Khaliun

cjyuan:
All responses are spot on. Well done.

illicitonion:

PR #1034, opened 2026-02-23
London | 26-ITP-Jan | Johnny Vargas | Sprint 2 | Coursework

JohnnyBoyV:
## Learners, PR Template

Self checklist

## Changelist

Answered and completed the questions in the .js files as best I could.

### DIFF
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -15,5 +15,8 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
+    height = height * height;
+    return weight / height;
 }
+
+console.log("Your BMI is: " + calculateBMI(70, 1.73).toFixed(1))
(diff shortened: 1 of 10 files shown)


Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,8 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    height = height * height;
+    return weight / height;
```
cjyuan: - After line 18, the variable `height` no longer stores a value that match its name. It is better to introduce a constant or a variable with a meaningful name to store the value. Or you can combine the code on lines 18 and 19 into one statement.

- The spec on line 15 says the function should return the computed value to 1 decimal place. So that means the rounding need to be done in the function.

Sprint-2/4-mandatory-interpret/time-format.js
```diff
@@ -11,24 +11,26 @@ function formatTimeDisplay(seconds) {
...
 
 // Call formatTimeDisplay with an input of 61, now answer the following:
 
 // b) What is the value assigned to num when pad is called for the first time?
-// =============> write your answer here
+// =============> 0, there are no hours in 61 seconds. (first is totalHours)
 
 // c) What is the return value of pad is called for the first time?
-// =============> write your answer here
+// =============> 00, as there are 0 hours it pads it to 00 at least.
```

cjyuan:
Changes are all good.

illicitonion:

PR #1035, opened 2026-02-23
London | JAN-2026 ITP | Asha Ahmed | Sprint 3 | implement and rewrite tests coursework 

ashaahmed7:
## Learners, PR Template

Self checklist

## Changelist

Implemented getAngleType, isProperFraction, and getCardValue solutions.

Rewrote tests using Jest to cover expected cases, including boundary/invalid inputs.

## Questions

n/a

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
@@ -15,8 +15,28 @@
 // execute the code to ensure all tests pass.
 
 function getAngleType(angle) {
-  // TODO: Implement this function
+  if (angle > 0 && angle < 90) {
+    return "Acute angle";
+  } else if (angle === 90) {
+    return "Right angle";
+  } else if (angle > 90 && angle < 180) {
+    return "Obtuse angle";
+  } else if (angle === 180) {
+    return "Straight angle";
+  } else if (angle > 180 && angle < 360) {
+    return "Reflex angle";
+  } else {
+    return "Invalid angle";
+  }
 }
+
+console.assert(getAngleType(45) === "Acute angle");
+console.assert(getAngleType(90) === "Right angle");
+console.assert(getAngleType(120) === "Obtuse angle");
+console.assert(getAngleType(180) === "Straight angle");
+console.assert(getAngleType(270) === "Reflex angle");
+console.assert(getAngleType(0) === "Invalid angle");
+console.assert(getAngleType(360) === "Invalid angle");
 
 // The line below allows us to load the getAngleType function into tests in other files.
 // This will be useful in the "rewrite tests with jest" step.
@@ -35,3 +55,27 @@ function assertEquals(actualOutput, targetOutput) {
 // Example: Identify Right Angles
 const right = getAngleType(90);
 assertEquals(right, "Right angle");
+
+// Example: Identify Acute Angles
+const acute = getAngleType(45);
+assertEquals(acute, "Acute angle");
+
+// Example: Identify Obtuse Angles
+const obtuse = getAngleType(120);
+assertEquals(obtuse, "Obtuse angle");
+
+// Example: Identify Straight Angles
+const straight = getAngleType(180);
+assertEquals(straight, "Straight angle");
+
+// Example: Identify Reflex Angles
+const reflex = getAngleType(270);
+assertEquals(reflex, "Reflex angle");
+
+// Example: Identify Invalid Angles
+const invalid1 = getAngleType(0);
+assertEquals(invalid1, "Invalid angle");
+
+// Example: Identify Invalid Angles
+const invalid2 = getAngleType(360);
+assertEquals(invalid2, "Invalid angle");
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -12,6 +12,13 @@
 
 function isProperFraction(numerator, denominator) {
   // TODO: Implement this function
+  if (denominator <= 0) {
+    return false;
+  }
+  if (numerator < denominator && numerator >= 0) {
+    return true;
+  }
+  return false;
 }
 
 // The line below allows us to load the isProperFraction function into tests in other files.
@@ -31,3 +38,27 @@ function assertEquals(actualOutput, targetOutput) {
 
 // Example: 1/2 is a proper fraction
 assertEquals(isProperFraction(1, 2), true);
+
+// Example: 3/5 is a proper fraction
+assertEquals(properFraction(3 / 5), true);
+
+// Example: 1/100 is a positive fraction
+assertEquals(smallFraction(1, 100), true);
+
+// Example: 0/4 is a proper fraction with zero numerator
+assertEquals(zeroNumerator(0, 4), true);
+
+// Example: 7/7 is an improper fraction,
+assertEquals(equalFraction(7, 7), false);
+
+// Example: 8/3 is an improper fraction,
+assertEquals(numeratorLarger(8, 3), false);
+
+// Example: 5/0 is an invalid fraction
+assertEquals(zeroDenominator(5, 0), false);
+
+// Example: 2/-6 is an invalid fraction
+assertEquals(negativeDenominator(2, -6), false);
+
+// Example: -1/4 is an invalid fraction
+assertEquals(negativeNumerator(-1, 4), false);
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -40,25 +40,25 @@ function assertEquals(actualOutput, targetOutput) {
 assertEquals(isProperFraction(1, 2), true);
 
 // Example: 3/5 is a proper fraction
-assertEquals(properFraction(3 / 5), true);
+assertEquals(isProperFraction(3, 5), true);
 
-// Example: 1/100 is a positive fraction
-assertEquals(smallFraction(1, 100), true);
+// Example: 1/100 is a proper fraction
+assertEquals(isProperFraction(1, 100), true);
 
-// Example: 0/4 is a proper fraction with zero numerator
-assertEquals(zeroNumerator(0, 4), true);
+// Example: 5/2 is not a proper fraction
+assertEquals(isProperFraction(5, 2), false);
 
-// Example: 7/7 is an improper fraction,
-assertEquals(equalFraction(7, 7), false);
+// Example: 7/7 is an improper fraction
+assertEquals(isProperFraction(7, 7), false);
 
-// Example: 8/3 is an improper fraction,
-assertEquals(numeratorLarger(8, 3), false);
+// Example: 8/3 is an improper fraction
+assertEquals(isProperFraction(8, 3), false);
 
 // Example: 5/0 is an invalid fraction
-assertEquals(zeroDenominator(5, 0), false);
+assertEquals(isProperFraction(5, 0), false);
 
 // Example: 2/-6 is an invalid fraction
-assertEquals(negativeDenominator(2, -6), false);
+assertEquals(isProperFraction(2, -6), false);
 
 // Example: -1/4 is an invalid fraction
-assertEquals(negativeNumerator(-1, 4), false);
+assertEquals(isProperFraction(-1, 4), false);
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
@@ -23,6 +23,14 @@
 
 function getCardValue(card) {
   // TODO: Implement this function
+  if (!["♠", "♣", "♦", "♥"].includes(card.slice(-1)))
+    throw new Error("Invalid card rank.");
+  const rank = card.slice(0, card.length - 1);
+  if (rank === "A") return 11;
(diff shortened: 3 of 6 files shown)


Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -50,3 +64,33 @@ try {
 } catch (e) {}
```
netEmmanuel, line 64: The catch block can be improved. it's currently empty

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -50,3 +64,33 @@ try {
...
 
 // What other invalid card cases can you think of?
+
+try {
```
netEmmanuel, line 68: An empty catch block is not good practice. 

Fix the catch blocks in this file

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -7,6 +7,74 @@ const getCardValue = require("../implement/3-get-card-value");
 // Case 1: Ace (A)
 test(`Should return 11 when given an ace card`, () => {
   expect(getCardValue("A♠")).toEqual(11);
+  expect(getCardValue("A♣")).toEqual(11);
```
netEmmanuel, line 10: These are duplicated test lines. Is there a reason for this?

netEmmanuel (changes requested):

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
ashaahmed7: You are referring to the original CodeYourFuture code example, which I did not write. The question I was asked was “What other invalid card cases can you think of?” (see below), not to modify the code you are commenting on.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
ashaahmed7: My answers are based on the example given above.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
ashaahmed7: I don't think this is a duplication. We have two codes who have different purpose: one for spades and one for clubs. Can you please check again?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
netEmmanuel: Avoid empty catch. At the very least, there should be a comment in the empty block explaining why you're swallowing the exception at that point or output the exception.

Look up the standard on how to write a try and catch block; it would help fix the syntax.

ashaahmed7: I'm still learning JavaScript, so I’ll look into catch blocks as I’m not entirely sure what you mean yet. My understanding was that the code is technically correct, and my answers were based on the example provided by CodeYourFuture above. Given that, it feels like a big change to fix the file itself rather than respond to the questions asked.

Poonam-raj: Hey Asha, so Emmanuel is actually right here. We do want that "catch" block to be more informative.

Let's think of it another way. The goal for this test is to **show** that when an invalid card is passed in, an error is thrown. That's a really interesting problem because when you pass a invalid card into the function, an error will throw and you won't have the chance to "show" that the error behaviour is tested for because the whole programme crashes. Annoying. 

But that's why `try...catch` is so useful. It lets us "try" code in the `try...` block which might fail, and if an error occurs we can "catch" the error in the `catch...` block without crashing the programme 🎉 

The next step which Emmanuel is referencing is that _it isn't enough to just have a empty catch block_ we should use that block to communicate that the expected behaviour has occurred. In this case, the expected behaviour is the error message is thrown. So that catch block has a parameter `e`
E.g. `catch (e) {}` -> can you use this parameter to console log something that SHOWS us that the right error message was thrown?

More helpful reading about catch blocks:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch
https://www.w3schools.com/js/js_errors.asp
https://dev.to/wisdomudo/javascript-trycatch-explained-a-beginners-guide-to-error-handling-2kf6

Poonam-raj: hint: What could that `e` parameter represent? Docs and logging could help here

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
Poonam-raj: Fair to say it is not a duplication here as they are different suites

Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
```diff
@@ -15,9 +15,29 @@
...
 }
 
+console.assert(getAngleType(45) === "Acute angle");
+console.assert(getAngleType(90) === "Right angle");
+console.assert(getAngleType(120) === "Obtuse angle");
+console.assert(getAngleType(180) === "Straight angle");
+console.assert(getAngleType(270) === "Reflex angle");
+console.assert(getAngleType(0) === "Invalid angle");
+console.assert(getAngleType(360) === "Invalid angle");
+
```
Poonam-raj, line 40: In future I'd remove console.asserts as they aren't adding anything to your solution thanks to your actual assertions below on line 60 onwards. Before you submit code make sure you clear those less useful bits like these lines

Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
```diff
@@ -15,9 +15,29 @@
...
+    return "Right angle";
+  } else if (angle > 90 && angle < 180) {
+    return "Obtuse angle";
+  } else if (angle === 180) {
+    return "Straight angle";
+  } else if (angle > 180 && angle < 360) {
+    return "Reflex angle";
+  } else {
+    return "Invalid angle";
+  }
```
Poonam-raj, line 30: great work here - an opportunity for refactor is considering if every if block contains a `return` statement do we need `else if` blocks on lines 20 -26 or can they be just `if` blocks? Try it out, see if you can predict the outcome and whether the code runs like you expect.

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -12,6 +12,13 @@
 
 function isProperFraction(numerator, denominator) {
   // TODO: Implement this function
+  if (denominator <= 0) {
+    return false;
+  }
+  if (numerator < denominator && numerator >= 0) {
+    return true;
+  }
```
Poonam-raj, line 20: This is a great example of what i was suggesting in my previous comment! That a plain `if` block works here, no `else if` needed. Why is that the case?

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -12,6 +12,13 @@
 
 function isProperFraction(numerator, denominator) {
   // TODO: Implement this function
+  if (denominator <= 0) {
+    return false;
+  }
```
Poonam-raj, line 17: You return false on line 21 and you return false here. Can you refactor this function completely to be even more efficient about returning false? Can we return false just once in the solution?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -23,6 +23,14 @@
 
 function getCardValue(card) {
   // TODO: Implement this function
+  if (!["♠", "♣", "♦", "♥"].includes(card.slice(-1)))
```
Poonam-raj, line 26: Great use of methods here (`includes`, `slice`) and bang operator (`!`), how did you feel about nesting these methods like this?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -23,6 +23,14 @@
...
+    throw new Error("Invalid card rank.");
```
Poonam-raj, line 27: Nice throw of an error

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -23,6 +23,14 @@
...
+  const rank = card.slice(0, card.length - 1);
+  if (rank === "A") return 11;
+  if (["10", "J", "Q", "K"].includes(rank)) return 10;
+  if (["2", "3", "4", "5", "6", "7", "8", "9"].includes(rank))
+    return Number(rank);
```
Poonam-raj, line 32: Some super efficient if blocks here, being really accurate and DRY about the return statements.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
Poonam-raj: Edit: this is really a nice-to-have but not an expectation based on how this task is designed. We could argue the console log on line 71 not running is enough evidence in this task that the error was thrown.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -50,3 +64,33 @@ try {
...
+  getCardValue("A");
+
+  console.log("Error was not thrown for invalid card");
+} catch (error) {}
+
+try {
+  getCardValue("JK");
+
+  console.log("Error was not thrown for invalid card");
+} catch (error) {}
```
Poonam-raj, line 96: Super repeated code (breaks the Don't Repeat Yourself rule). Could you refactor this code into a reusable function that you pass your invalid cases to?

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
```diff
@@ -14,7 +14,32 @@ test(`should return "Acute angle" when (0 < angle < 90)`, () => {
...
+test("should identify straight angle (angle = 180°)", () => {
+  expect(getAngleType(180)).toEqual("Straight angle");
+});
+
 // Case 5: Reflex angles
+test("should identify reflex angle (180° < angle < 360)", () => {
+  expect(getAngleType(181)).toEqual("Reflex angle");
+  expect(getAngleType(280)).toEqual("Reflex angle");
+  expect(getAngleType(359)).toEqual("Reflex angle");
+});
```
Poonam-raj, line 38: Across this file, good grouping of similar behaviours.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -7,6 +7,74 @@ const getCardValue = require("../implement/3-get-card-value");
...
+  expect(getCardValue("A♦")).toEqual(11);
+  expect(getCardValue("A♥")).toEqual(11);
+});
+
+// Case 2: Handle Number Cards (2-10):
+test("should return the appropriate number from 2 to 10", () => {
```
Poonam-raj, line 16: I would say this is an extremely excessive jest test block. I'd prefer to take a snapshot of the test cases that we need to check this behaviour for. Doing every combo of number and suite is very detailed but also quite bulky. 
One reason this is a bad idea is it could be computationally heavy, e.g. if the function was running a very complicated algorithm for this behaviour. 
But the tradeoff is that you know it works for every case. 

I would argue that the "suite" doesn't actually matter in the case described; the number is important. So you should focus on this. If the suite mattered, the test case would be including it in the description but it only mentioned number cards.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -7,6 +7,74 @@ const getCardValue = require("../implement/3-get-card-value");
...
+  expect(getCardValue("K♥")).toEqual(10);
+});
+
+// Case 4: Handle Invalid Cards:
+test("Should return 'Invalid card rank.' for invalid cards", () => {
+  expect(() => getCardValue("KJ")).toThrow("Invalid card rank.");
+  expect(() => getCardValue("AK")).toThrow("Invalid card rank.");
+  expect(() => getCardValue("  ")).toThrow("Invalid card rank.");
+  expect(() => getCardValue("S♠")).toThrow("Invalid card rank.");
+  expect(() => getCardValue("J♠♠")).toThrow("Invalid card rank.");
```
Poonam-raj, line 77: Great work

Poonam-raj (commented):
So I've done a quick review of what's happening here. 

Few things I love
- choices of methods and compact if statements to ensure logic is DRY.
- test cases and detail in the jest testing
- error throwing

Few areas to think about (but are not core criteria here):
- else if vs if statements, when do we need a else if?
- how big we make our test blocks, what behaviour do we ACTUALLY care about

illicitonion:

PR #1036, opened 2026-02-23
London | 26-ITP-January | Karla Grajales | Sprint 3 | 1-implement-and-rewrite-tests

Grajales-K:
## Learners, PR Template

Self checklist

## Changelist

In this project, I learned how to implement functions and write comprehensive test cases using assertions and error handling (throw). I also practiced using helper functions to improve test readability and finally transitioned to Jest by exporting the logic as a module.

## Questions

Hi, I have a question regarding the testing strategy. When and why is it better to implement manual assertions versus using a framework like Jest? Are there specific rules or industry standards on when to use each approach? 
Thank you.


Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -22,7 +22,42 @@
 // execute the code to ensure all tests pass.
 
 function getCardValue(card) {
-  // TODO: Implement this function
+  //check the length and if does not match the valid length throw an error
```
illicitonion: This comment is correct, but I'm curious - what value do you think this comment is providing that the variable names and code structure below don't already provide?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -22,7 +22,42 @@
...
+  const suit = card[card.length - 1];
+
+  let suitValid = false;
+
+  for (let i = 0; i < suits.length; i++) {
+    if (suit === suits[i]) {
+      suitValid = true;
+      break;
+    }
+  }
```
illicitonion: This works, but arrays have built-in functions for finding whether the array contains a value - it could be simpler to try using that.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -22,7 +22,42 @@
...
+  // after comparing each suit in our array and does not match throw an error
+  if (suitValid === false) {
```
illicitonion: It's very unusual to compare values to `true` or `false` - we tend to prefer just `if (suitValid)` or `if (!suitValid)` - `if` statements take any expression as their condition, and avoiding the extra comparison operator leads to simpler code.

illicitonion (commented):
This is generally looking really good, well done!

On your question:

> Hi, I have a question regarding the testing strategy. When and why is it better to implement manual assertions versus using a framework like Jest? Are there specific rules or industry standards on when to use each approach?

In real life, we almost always use a testing framework - they tend to have lots of convenience things (e.g. as you've seen here `expect(...).toThrow(...)` is much more convenient to write than having to do a `try-catch` and failing the test in the `try` after the code you're testing - you _can_ build these things up yourself, but frameworks exist to do this for you :)

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
Grajales-K: I included those comments as a guide while I was building the logic. Now that the code structure and variable names are clear, I agree they are redundant and I will remove them. Thanks for the feedback!

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
Grajales-K: Hi @illicitonion, thank you for this suggestion, I refactored the suit validation using` .includes()`. It's much more readable now and less prone to error than the manual loop.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
Grajales-K: Hi, yes, I simplified the code with your suggestions. I replaced the loop in the previous commit. Now looks much better. Thank you.

Grajales-K:
> This is generally looking really good, well done!
> 
> On your question:
> 
> > Hi, I have a question regarding the testing strategy. When and why is it better to implement manual assertions versus using a framework like Jest? Are there specific rules or industry standards on when to use each approach?
> 
> In real life, we almost always use a testing framework - they tend to have lots of convenience things (e.g. as you've seen here `expect(...).toThrow(...)` is much more convenient to write than having to do a `try-catch` and failing the test in the `try` after the code you're testing - you _can_ build these things up yourself, but frameworks exist to do this for you :)

Hi @illicitonion, that makes a lot of sense. Building manual assertions was a great exercise to understand how testing works under the hood, but I can see how `expect(...).toThrow()` indeed is much cleaner and more efficient for real-world projects. 

Thank you very much for your time.

illicitonion (commented):
This looks great, well done!

illicitonion:

PR #1045, opened 2026-02-24
London | 26-ITP-Jan | Kris Oldrini | Sprint 3 | Implement And Rewrite Tests

XiaoQuark:
## Learners, PR Template

Self checklist

### 1 Implement solutions

- Implemented required functions for files:
  - `1-get-angle-type.js`
  - `2-is-proper-fraction.js`
  - `3-get-card-value.js`
- Added tests in the above files to cover all possible cases (including error handling).
- Verified all tests passed.

### 2 Rewrite tests with Jest

- Rewrote equivalent test suites in Jest
- Verified all tests pass via `npm test`

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
@@ -23,42 +23,51 @@
 
 function getCardValue(card) {
   const validSuits = ["♠", "♥", "♦", "♣"];
+  const validRanks = [
+    "A",
+    "1",
+    "2",
+    "3",
+    "4",
+    "5",
+    "6",
+    "7",
+    "8",
+    "9",
+    "10",
+    "J",
+    "Q",
+    "K",
+  ];
   if (typeof card !== "string")
     throw new Error(`Invalid card format: "${card}". Expected a string.`);
 
-  let cardSuit = card.slice(-1);
-  let cardRank = card.slice(0, -1);
+  const cardSuit = card.slice(-1);
+  const cardRank = card.slice(0, -1);
 
-  if (validSuits.includes(cardSuit)) {
-    switch (cardRank) {
-      case "A":
-        return 11;
-
-      case "J":
-      case "Q":
-      case "K":
-        return 10;
-
-      case "2":
-      case "3":
-      case "4":
-      case "5":
-      case "6":
-      case "7":
-      case "8":
-      case "9":
-      case "10":
-        return Number(cardRank);
-
-      default:
-        throw new Error(
-          `Invalid card format: "${card}" has an invalid rank "${cardRank}". Expected one of: A, 2–10, J, Q, K.`
-        );
-    }
-  } else
+  if (!validSuits.includes(cardSuit)) {
     throw new Error(
       `Invalid card format: "${card}" has an invalid suit "${cardSuit}". Expected one of: ♠, ♥, ♦, ♣.`
     );
+  }
+  if (!validRanks.includes(cardRank)) {
+    throw new Error(
+      `Invalid card format: "${card}" has an invalid rank "${cardSuit}". Expected one of: ♠, ♥, ♦, ♣.`
+    );
+  }
+
+  switch (cardRank) {
+    case "A":
+      return 11;
+
+    case "J":
+    case "Q":
+    case "K":
+      return 10;
+
+    default:
+      return Number(cardRank);
+  }
 }
 
 // The line below allows us to load the getCardValue function into tests in other files.
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
@@ -22,37 +22,27 @@
 // execute the code to ensure all tests pass.
 
 function getCardValue(card) {
-  const validSuits = ["♠", "♥", "♦", "♣"];
-  const validRanks = [
-    "A",
-    "1",
-    "2",
-    "3",
-    "4",
-    "5",
-    "6",
-    "7",
-    "8",
-    "9",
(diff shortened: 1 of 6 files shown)


Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -5,6 +5,45 @@ const isProperFraction = require("../implement/2-is-proper-fraction");
 // TODO: Write tests in Jest syntax to cover all combinations of positives, negatives, zeros, and other categories.
 
 // Special case: numerator is zero
-test(`should return false when denominator is zero`, () => {
```
netEmmanuel, line 8: Is there a reason why this test was deleted?

netEmmanuel (changes requested):

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
XiaoQuark: Hi Emmanuel, I deleted it because I wanted the wording to be more consistent with my other tests. If you see below, I do have a test for the same special case: `should return false when the denominator is equal to zero`

XiaoQuark:
image

I'm not sure I understand this github message. Is there something I need to do?

netEmmanuel:
@XiaoQuark, I left a comment on your PR that needs your attention. 

here: https://github.com/CodeYourFuture/Module-Structuring-and-Testing-Data/pull/1045/changes#r2891725763

XiaoQuark:
Hi @netEmmanuel, I mentioned this in my reply yesterday, but just to clarify: I didn’t remove that case. I just rewrote the test so the wording is consistent with my other tests. The “denominator equals 0” scenario is still covered in the test `should return false when the denominator is equal to zero`.

Let me know if you were expecting a different edge case to be tested.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -24,24 +24,19 @@
 function getCardValue(card) {
   const validSuits = ["♠", "♥", "♦", "♣"];
   if (typeof card !== "string")
-    throw new Error(
-      `Invalid card format: "${card}". Expected rank followed by suit, for example A♠, 10♥, K♦, or 7♣.`
-    );
+    throw new Error(`Invalid card format: "${card}". Expected a string.`);
 
   let cardSuit = card.slice(-1);
   let cardRank = card.slice(0, -1);
```
favourO: This looks much cleaner after removing the logs. One small thing to think about: since cardSuit and cardRank aren't reassigned later in the function, is **let** the most appropriate declaration here?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -24,24 +24,19 @@
...
 
-  if (validSuits.find((suit) => suit === cardSuit)) {
-    console.log(cardSuit);
+  if (validSuits.includes(cardSuit)) {
```
favourO: The suit validation works well here. I’m wondering though; if the suit was invalid, what would happen to the rest of the function? Is there a way we could structure the check so that invalid cases exit earlier and the main logic stays flatter?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -53,17 +48,16 @@ function getCardValue(card) {
       case "8":
```
favourO: The numeric ranks (2–10) all follow the same logic. Do you think we need a separate case for each one, or could they be handled with a single rule?

XiaoQuark:
Hi @favourO , thank you for your feedback. I think I have fixed everything you mentioned.
I decided to create a second array of valid ranks to validate them before entering the switch. Not sure if this was the idea you had in mind.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -22,7 +22,52 @@
 // execute the code to ensure all tests pass.
 
 function getCardValue(card) {
-  // TODO: Implement this function
+  const validSuits = ["♠", "♥", "♦", "♣"];
+  const validRanks = [
```
favourO: This is a nice improvement, 
For the numeric cards, do you think you could check if the rank is between 1 and 10 instead of listing each value from 1 to 10? For example, by converting cardRank to a number and checking something like >= 1 && <= 10.

favourO (commented):
Great improvement, just a little more effort
I added a comment.

XiaoQuark:
@favourO 
Sorry for the late reply. I refactored the code as requested.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -22,37 +22,27 @@
...
 
+  const validSuits = ["♠", "♥", "♦", "♣"];
+  const letterRanks = ["A", "J", "Q", "K"];
+
   const cardSuit = card.slice(-1);
   const cardRank = card.slice(0, -1);
+  const numericRank = Number(cardRank);
+
+  const validNumericRank =
+    Number.isInteger(numericRank) && numericRank >= 2 && numericRank <= 10;
```
favourO, line 36: Great job, exactly what I was looking for

favourO (commented):
All done now.

XiaoQuark:
Thank you @favourO ,

Do you think the PR could be marked as completed then?

illicitonion:

PR #1046, opened 2026-02-24
London | 26-ITP-Jan | Kris Oldrini | Sprint 3 | Practice TDD

XiaoQuark:
## Learners, PR Template

Self checklist

## Changelist

- Completed TDD exercises using Jest

  - Wrote tests in files:

    - `count.test.js`
    - `repeat-str.test.js`
    - `get-ordinal-number.test.js`

  - Implemented functions in files:

    - `count.js`
    - `repeat-str.js`
    - `get-ordinal-number.js`

- Followed a test-first workflow.

- Refactored some tests after implementation for clarity and readability.

 - Verified all tests pass using `npm test`.

### DIFF
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -29,3 +29,19 @@ test("should count 0 occurrences of a character", () => {
   const count = countChar(str, char);
   expect(count).toEqual(0);
 });
+
+test("should be case sensitive and return only occurrences of same case characters", () => {
+  const str = "Dodgers";
+  const char = "D";
+  const count = countChar(str, char);
+  expect(count).toEqual(1);
+});
+
+test("should count occurrences of regex metacharacters", () => {
+  const str = "a*b*c";
+  const char = "*";
+  const count = countChar(str, char);
+  expect(count).toEqual(2);
+});
+
+// metacharacters . ^ $ * + ? { } [ ] \ | ( )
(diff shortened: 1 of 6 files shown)


Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -22,3 +22,10 @@ test("should count multiple occurrences of a character", () => {
 // And a character `char` that does not exist within `str`.
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of `char` were found.
+
+test("should count 0 occurrences of a character", () => {
+  const str = "dog";
+  const char = "a";
+  const count = countChar(str, char);
+  expect(count).toEqual(0);
```
cjyuan, line 30: - Could consider testing more samples.

- Could consider test these cases:
    - A case to show that the function should work also for non-alphabets 
      - Currently your function does not work properly when `char` is a certain non-alphabet character that has special meaning in regex.

XiaoQuark: Thank you CJ, I reviewed your comment, and I think I'm now covering all the cases you mentioned.
This was actually really fun to work on. I've been learning a lot about working with regex.

cjyuan: Seems to work ok now.

I am glad you learned something more about regex.

It is ok to use this approach in this course but I wouldn't recommend it in practice for such a task.
Can you find out (maybe from AI) the pros and cons using this approach to count characters?

cjyuan (commented):
Changes look good. Well done!

illicitonion:

PR #1058, opened 2026-02-25
Manchester | 26-ITP-JAN | Abdu Mussa | Sprint 1 | structuring and Testing Data 

Abduhasen:
## Learners, PR Template

Self checklist

## Changelist

this PR is for the Structuring and Testing Data.
-working on errors identifying them and fixing the errors with explanation 
-brief explanation of codes and interpret  
## Questions

N/A

### DIFF
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -16,8 +16,10 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
+const secondCharacterIndex = 1;
+const startingPointForExtension = filePath.lastIndexOf(".");
 
-const dir = ;
-const ext = ;
-
+const dir = filePath.slice(secondCharacterIndex, lastSlashIndex);
+const ext = filePath.slice(startingPointForExtension + 1, filePath.length);
+console.log(`The dir part of ${filePath} is ${dir} and the ext part is ${ext}`);
 // https://www.google.com/search?q=slice+mdn
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -20,6 +20,6 @@ const secondCharacterIndex = 1;
 const startingPointForExtension = filePath.lastIndexOf(".");
 
 const dir = filePath.slice(secondCharacterIndex, lastSlashIndex);
-const ext = filePath.slice(startingPointForExtension + 1, filePath.length);
+const ext = filePath.slice(startingPointForExtension + 1);
 console.log(`The dir part of ${filePath} is ${dir} and the ext part is ${ext}`);
 // https://www.google.com/search?q=slice+mdn
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -7,3 +7,9 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+/*num represent any value between 1 to 100.
+Math.floor is a method that rounds a number down to the nearest whole number.
+Math.random is a method returning any number that is greater than or equal to 0 and less than 1.
+first of all i will try to get a value from 0 to less than 1 from Math.random and multiple it by 100
+and apply Math.floor to the result which will give me the integer number and then add the minimum value i.e. 1   */
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -12,4 +12,4 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 Math.floor is a method that rounds a number down to the nearest whole number.
 Math.random is a method returning any number that is greater than or equal to 0 and less than 1.
 first of all i will try to get a value from 0 to less than 1 from Math.random and multiple it by 100
-and apply Math.floor to the result which will give me the integer number and then add the minimum value i.e. 1   */
+and apply Math.floor to the result which will give me the integer number and then add the minimum value i.e. 1  */
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -1,6 +1,4 @@
-const minimum = 1;
-const maximum = 100;
-
+const [min, max] = [1, 100];
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 
 // In this exercise, you will need to work out what num represents?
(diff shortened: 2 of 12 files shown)


Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -16,8 +16,10 @@ console.log(`The base part of ${filePath} is ${base}`);
...
+const secondCharacterIndex = 1;
+const startingPointForExtension = filePath.lastIndexOf(".");
 
-const dir = ;
-const ext = ;
-
-// https://www.google.com/search?q=slice+mdn
\ No newline at end of file
+const dir = filePath.slice(secondCharacterIndex, lastSlashIndex);
+const ext = filePath.slice(startingPointForExtension + 1, filePath.length);
```
cjyuan: Can you look up why on line 23 the second parameter of `.slice()` is optional?

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -7,3 +7,9 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+/*num represent any value between 1 to 100.
+Math.floor is a method that rounds a number down to the nearest whole number.
+Math.random is a method returning any number that is greater than or equal to 0 and less than 1.
+first of all i will try to get a value from 0 to less than 1 from Math.random and multiple it by 100
+and apply Math.floor to the result which will give me the integer number and then add the minimum value i.e. 1  */
```
cjyuan: We can also use the concise and precise ***interval notation*** to describe a range of values. 
- `[`, `]` => inclusion
- `(`, `)` => exclusion

Sprint-1/2-mandatory-errors/3.js
```diff
@@ -1,4 +1,4 @@
-const cardNumber = 4533787178994213;
+const cardNumber = "4533787178994213";
 const last4Digits = cardNumber.slice(-4);
```

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,23 @@ console.log(result);
...
 // c) Using documentation, explain what the expression movieLength % 60 represents
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators
+/* the expression movieLength % 60 represents the returns of the remainder after dividing one number by another 
+   in this case 8784 % 60 and it return the remainder 24. */
 
 // d) Interpret line 4, what does the expression assigned to totalMinutes mean?
+//This expression calculates the total full minutes in the movie by removing leftover seconds and converting seconds to minutes.
 
 // e) What do you think the variable result represents? Can you think of a better name for this variable?
+//result represent the total movie length using hours,minutes and seconds,the better name can be totalMovieLength.
```
cjyuan: The name `totalMovieLength` does not quite indicate the value the variable stores is a formatted string in the form `"2:12:02"`.
Can you suggest a more descriptive name?

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -24,4 +24,26 @@ console.log(`£${pounds}.${pence}`);
...
+        0,
+        paddedPenceNumberString.length - 2
+      );  : this part of the code assigning const pound with value of string that length is 
+            reduced by -2 . */
+
+/* 5. const pence = paddedPenceNumberString
+        .substring(paddedPenceNumberString.length - 2)
+        .padEnd(2, "0"); : let's break it down this code parts it have two steps first one is .substring(paddedPenceNumberString.length - 2)
+                            Takes the last 2 characters of the string and the second is .padEnd(2, "0") Ensures the result 
+                            is at least length 2 if the length less than 2 it will add 0 at the end of the string. */
```
cjyuan, line 48: Optional challenge: 

cjyuan (commented):
Have you installed **prettier** VSCode extension and enable formatting on save/paste on VSCode as recommended in 
https://github.com/CodeYourFuture/Module-Structuring-and-Testing-Data/blob/main/readme.md ?

Sprint-1/1-key-exercises/3-paths.js
Abduhasen: yes it is optional in this case even if i don't give the second parameter the result will be the same.

Sprint-1/1-key-exercises/4-random.js
Abduhasen: I read and understood the concept of interval notation but i want to ask how we can apply it in our case.

Sprint-1/3-mandatory-interpret/2-time-format.js
Abduhasen: thank you for your feedback i changed it to totalMovieDuration.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
Abduhasen: thank you for the feedback, we don't need the .padEnd(2, "0") in this script because the program will work for any valid penceString even after removing the padEnd.

Sprint-1/1-key-exercises/4-random.js
cjyuan: You can say, "`num` represents a number in [1, 100]" and "`Math.random()` returns a number in [0, 1)". Less wordy and more precise.

Abduhasen: Is the following code right way to do it?

`const [min, max] = [1, 100];
const num = Math.floor(Math.random() * (max - min + 1)) + min;`

cjyuan: I meant in the comment when we need to describe the range in words.

Sprint-1/3-mandatory-interpret/2-time-format.js
cjyuan: `totalMovieDuration` still does not quite describe the value that is a string formatted in form `"02:12:02"`.

Numbers such as 2.2 (hours) and 132.03 (minutes) are also valid total movie duration.

One more try?

Sprint-1/3-mandatory-interpret/3-to-pounds.js
cjyuan: Did you figure out **why** `.padEnd(2, "0")` is optional in this script? It is more important to understand why.

Sprint-1/3-mandatory-interpret/2-time-format.js
Abduhasen: how about `movieRunTime` this one can describe it more.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
Abduhasen: yes i understood the concept before i didn't realise it ( which is even if we didn't put the pading there will be a value of 2 character and we don't need to have extra script. )

i also want to thank you because i learned more by the feedback i got from you.

Sprint-1/1-key-exercises/4-random.js
Abduhasen: I was confused  now it's clear and i did some changes less wordy now.

Sprint-1/3-mandatory-interpret/2-time-format.js
cjyuan: Not much difference from `movieDuration`. You could try asking an AI tool for suggestion.

Sprint-1/3-mandatory-interpret/3-to-pounds.js
cjyuan: I am glad you understand the concept.
Can you explain why `.padEnd(2, "0")` is optional in this script?

cjyuan:
Changes to the code look good. 

Can you explain why `.padEnd(2, "0")` is optional in `Sprint-1/3-mandatory-interpret/3-to-pounds.js`?
You don't have to make any change to the file. Just respond in the comment will do.

---
> this PR is for the Structuring and Testing Data.

This description (In the PR description) is too general. Can you give a more specific description of what you have changed?

Sprint-1/3-mandatory-interpret/3-to-pounds.js
Abduhasen: const pence always take two string and there is no need of adding `.padEnd(2, "0")` and the string never be shorter than 2.

cjyuan:
Changes to the code are good.

Would be better if you can also address the comments in the two unresolved conversations.

I will mark this PR as "Complete" first.

illicitonion:

PR #1059, opened 2026-02-25
London | ITP-JAN-2026 | Said Fayaz Sadat | Sprint 2 | coursework/sprint 2

fayaz551:
## Learners, PR Template

Self checklist

## Changelist

I have completed sprint-2 tasks.

## Questions

n/a

### DIFF
--- a/Sprint-2/1-key-errors/1.js
+++ b/Sprint-2/1-key-errors/1.js
@@ -2,10 +2,11 @@
 
 // Why will an error occur when this program runs?
 // =============> write your prediction here
+// An error will occur because we are trying to declare a variable with the same name as a function parameter. In this case, we are declaring 'decimalNumber' inside the function, which conflicts with the parameter 'decimalNumber'. This will result in a SyntaxError because we cannot have two variables with the same name in the same scope.
 
 // Try playing computer with the example to work out what is going on
 
-function convertToPercentage(decimalNumber) {
+/*function convertToPercentage(decimalNumber) {
   const decimalNumber = 0.5;
   const percentage = `${decimalNumber * 100}%`;
 
@@ -13,8 +14,15 @@ function convertToPercentage(decimalNumber) {
 }
 
 console.log(decimalNumber);
+*/
 
 // =============> write your explanation here
+// The error is occurring because we are trying to declare a variable with the same name as a function parameter. In this case, we are declaring 'decimalNumber' inside the function, which conflicts with the parameter 'decimalNumber'. This will result in a SyntaxError because we cannot have two variables with the same name in the same scope.
 
 // Finally, correct the code to fix the problem
 // =============> write your new code here
+
+function convertToPercentage(decimalNumber) {
+  const percentage = `${decimalNumber * 100}%`; 
+  return percentage;
+}
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -16,4 +16,7 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
+    let bmi = weight / (height * height);
+    return bmi.toFixed(1);
 }
+console.log(calculateBMI(80, 1.80)); // should return 24.7
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -17,6 +17,6 @@
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
     let bmi = weight / (height * height);
-    return bmi.toFixed(1);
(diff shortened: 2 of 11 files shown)


Sprint-2/1-key-errors/1.js
```diff
@@ -2,19 +2,27 @@
...
   const percentage = `${decimalNumber * 100}%`;
 
   return percentage;
 }
 
 console.log(decimalNumber);
+*/
 
 // =============> write your explanation here
+// The error is occurring because we are trying to declare a variable with the same name as a function parameter. In this case, we are declaring 'decimalNumber' inside the function, which conflicts with the parameter 'decimalNumber'. This will result in a SyntaxError because we cannot have two variables with the same name in the same scope.
```
cjyuan, line 20: It is better to break a long line of comment or code into multiple lines so that one doesn't have to scroll horizontally in an editor to read everything.

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -16,4 +16,7 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    let bmi = weight / (height * height);
+    return bmi.toFixed(1);
```

```javascript
  
```

Sprint-2/3-mandatory-implement/3-to-pounds.js
```diff
@@ -4,3 +4,10 @@
 // You will need to declare a function called toPounds with an appropriately named parameter.
 
 // You should call this function a number of times to check it works for different inputs
+
+function toPounds(kilograms) {
+    return +(kilograms * 2.20462).toFixed(1);
+}
```
cjyuan: The goal is to implement the code in `Sprint-1/3-mandatory-interpret/3-to-pounds.js` as a function which, when given a pence string, returns a string representing the equivalent British pounds.

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,9 +4,20 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
-  if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+  const minutes = time.slice(3, 5);
```
cjyuan: We could also extract the last 2 characters as `time.slice(-2)`.

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,9 +4,20 @@
...
+
+  if (hours === 12) {
+    return `12:${minutes} pm`;
+  }
+
+  if (hours > 12) {
+  return `${hours - 12}:${minutes} pm`;
+}
+
   return `${time} am`;
```
cjyuan, line 21: If "01:00" is converted to "01:00 am", it is probably reasonable for the caller to expect "13:00" to be converted to "01:00 pm".

```
01:00 am
1:00 pm
12:00 am
01:00 pm
```

Sprint-2/1-key-errors/1.js
fayaz551: To be honest I didn't thought of that, fixed that one and going forward i will keep that in mind. Thanks.

Sprint-2/3-mandatory-implement/1-bmi.js
fayaz551: Thanks for the feedback, My function expected a number but I noticed that toFixed() returns a string, so my function was returning a string instead of a number. I updated the return to Number(bmi.toFixed(1)) so the function now returns a number while still rounding to one decimal place.

Sprint-2/3-mandatory-implement/3-to-pounds.js
fayaz551: Appreciate it for point me out in right direction, I’ve updated the implementation so the toPounds function now accepts a pence string (e.g. "399p"), converts it to pounds, and returns the correctly formatted pound value as a string. I also tested it with multiple inputs to confirm it works as expected.

Sprint-2/5-stretch-extend/format-time.js
fayaz551: changes it to time.slice(-2) to get the characters.

Sprint-2/5-stretch-extend/format-time.js
fayaz551: Thanks for the feedback Yuan. I see what you mean about the formatting inconsistency. I updated the function so that the hour is always returned with two digits using padStart, ensuring values like "13:00" are converted to "01:00 pm" to match the "01:00 am" format.

cjyuan:
Changes look good.

fayaz551:
> Changes look good.

Appreciate the feedback and pointing out the areas to improve, I will keep them in mind moving forward.

illicitonion:

PR #1072, opened 2026-02-25
Glasgow | 26-ITP-JAN | Prakash Dcosta | Sprint 1 | Structuring and Testing Data

dcostaprakash:
Changelist

I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
My changes meet the requirements of the task
I have tested my changes
My changes follow the style guide
my coursework

### DIFF
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -17,7 +17,10 @@ console.log(`The base part of ${filePath} is ${base}`);
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
 
-const dir = ;
-const ext = ;
+const dir = filePath.slice(0,lastSlashIndex);
+const lastDotIndex = filePath.lastIndexOf(".");
+const ext = filePath.slice(lastDotIndex + 1);
+console.log (`The dir part of the filePath variable is ${dir}`);
+console.log (`The ext part of the variable is ${ext}`);
 
 // https://www.google.com/search?q=slice+mdn
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -8,7 +8,7 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 // Function Math.random() generates a decimal number between 0 to 1 which is 0 to 0.9999
 // Funcion (maximum-minimum + 1). Therefore it is 100 - 1 + 1 = 100
 // The * sign will multiply the random number with 100 which becomes 0 to 99.99
-// The function Math.floor will round the new number to the nearest whole number. Hence it will be 0 to 99
+// The function Math.floor will unconditionally round it down. Hence it will be 0 to 99
 // + minimum will add 1 to the new rounded up number Hence the old rance 0 - 99 will now become 1 to 100
 // It will help to think about the order in which expressions are evaluated
 // Done - Try logging the value of num and running the program several times to build an idea of what the program is doing
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -3,7 +3,12 @@ const maximum = 100;
 
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 
-// In this exercise, you will need to work out what num represents?
-// Try breaking down the expression and using documentation to explain what it means
+// In this exercise, you will need to work out what num represents? Num represents a randomly selected number from >=1 till <=100
+// Try breaking down the expression and using documentation to explain what it means. 
+// Function Math.random() generates a decimal number between 0 to 1 which is 0 to 0.9999
+// Funcion (maximum-minimum + 1). Therefore it is 100 - 1 + 1 = 100
+// The * sign will multiply the random number with 100 which becomes 0 to 99.99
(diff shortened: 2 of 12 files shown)


Sprint-1/1-key-exercises/3-paths.js
WeiTsungCheng, line 1: Correct!
You can think about this further: What happens if the file path doesn't have a file extension?

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -3,7 +3,12 @@ const maximum = 100;
...
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 
-// In this exercise, you will need to work out what num represents?
-// Try breaking down the expression and using documentation to explain what it means
+// In this exercise, you will need to work out what num represents? Num represents a randomly selected number from >=1 till <=100
+// Try breaking down the expression and using documentation to explain what it means. 
+// Function Math.random() generates a decimal number between 0 to 1 which is 0 to 0.9999
+// Funcion (maximum-minimum + 1). Therefore it is 100 - 1 + 1 = 100
+// The * sign will multiply the random number with 100 which becomes 0 to 99.99
+// The function Math.floor will round the new number to the nearest whole number. Hence it will be 0 to 99
```
WeiTsungCheng: "Round to nearest whole number" is inaccurate; it should be unconditionally rounded down.

Sprint-1/2-mandatory-errors/2.js
WeiTsungCheng, line 1: Does any different between use const let  and var?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,17 @@ console.log(result);
...
+// There are 6 variable declarations in this program
 
 // b) How many function calls are there?
+// There is 1 function call
 
 // c) Using documentation, explain what the expression movieLength % 60 represents
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators
-
+// % 60 is a remainder operator which in this case divides the total by 60 and shares the remainder (seconds / minutes)
 // d) Interpret line 4, what does the expression assigned to totalMinutes mean?
```
WeiTsungCheng, line 23: Yes, but it might not be accurate enough.

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,17 @@ console.log(result);
...
-
+// The expression assigned to total minutes is the difference of the total movie time in seconds less the remainder of the seconds divided by 60 which gives the total length of the movie in minutes
 // e) What do you think the variable result represents? Can you think of a better name for this variable?
-
+// The variable result represents the time of the movie in Hours Mintutes and seconds. We could use time instead of result
 // f) Try experimenting with different values of movieLength. Will this code work for all values of movieLength? Explain your answer
+// Yes. This code will work for all values of movie length. I tried with 8400 seconds and 8405 seconds
```
WeiTsungCheng: Do you consider different type value? (NaN, Decimals (e.g., 8784.5)....)

Sprint-1/3-mandatory-interpret/3-to-pounds.js
```diff
@@ -25,3 +25,9 @@ console.log(`£${pounds}.${pence}`);
 
 // To begin, we can start with
 // 1. const penceString = "399p": initialises a string variable with the value "399p"
+// 2. Line 3 Removes the letter p from the end of the string
+// 3. Line 8 keeps the length to 3 and adds a 0 if the length is smaller like 50p -> 050 or 5p -> 005
+// 4. Line 9 displays pounds from the 1st digit of the 3
```
WeiTsungCheng: The phrase "pounds from the first digit of the 3" is not precise enough.

WeiTsungCheng (commented):
Great, but could be better.

Sprint-1/1-key-exercises/4-random.js
dcostaprakash: Understood. Done the change. Thank you

Sprint-1/2-mandatory-errors/2.js
dcostaprakash: Yes. The difference is when we know the value, we use Const, we cannot change it. We use let if the value will change and Var if it is undefined. Thank you

Sprint-1/3-mandatory-interpret/2-time-format.js
dcostaprakash: Understood. I made the change. Thank you

Sprint-1/3-mandatory-interpret/3-to-pounds.js
dcostaprakash: Yes. i made the change. Thank you

Sprint-1/3-mandatory-interpret/2-time-format.js
dcostaprakash: I actually missed that. I have now corrected my comment since i realised that decimal numbers or string/null or undefined could actually break the code. Thank you.

Sprint-1/1-key-exercises/3-paths.js
dcostaprakash: Oh Yes, i see it now. The code will not work currently. It will give the whole path as an extension.
Thank you

WeiTsungCheng:
nice!

illicitonion:

PR #1073, opened 2026-02-26
Sheffield | 26-ITP-jan | Richard Frimpong | Sprint 1 | Structuring and Testing Data

Richiealx:
## Learners, PR Template

Self checklist

## Changelist

Completed Sprint 1 coursework:

- Key exercises completed
- Mandatory errors analysed and explained in answers.md
- Mandatory interpret section completed
- Stretch exploration completed
- Learning log updated

Thanks for the feedback @cjyuan!

### Answers to follow-up questions

**Programming term for `count = count + 1`**

The one-word programming term is **increment**.  
The statement increases the value stored in the variable by 1 and assigns the updated value back to the same variable.  
This pattern is very common and can also be written using shorthand operators such as `count += 1;` or `count++;`.

**Is `.padEnd(2, "0")` necessary?**

Yes,  `.padEnd(2, "0")` is needed to handle cases where the pence value has fewer than two digits.  

For example:
- If `penceString = "5p"`, removing `"p"` gives `"5"`
- `substring(length - 2)` would return `"5"` (only one digit)

Without `.padEnd(2, "0")`, the output would be `£0.5`.  
With `.padEnd(2, "0")`, it becomes `£0.50`.

So `.padEnd` ensures the pence part always contains exactly two digits, which is required for correct currency formatting.


Sprint-1/1-key-exercises/1-count.js
```diff
@@ -2,5 +2,10 @@ let count = 0;
 
 count = count + 1;
 
+console.log(count);
+
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+
+// Line 3 takes the current value of count (0), adds 1 to it,
+// and assigns the result (1) back to the count variable using =.
```

Sprint-1/3-mandatory-interpret/answers.md
```diff
@@ -0,0 +1,155 @@
...
+4) `const pounds = paddedPenceNumberString.substring(0, paddedPenceNumberString.length - 2);`
+- Takes everything except the last 2 digits.
+- Purpose: the part before the last 2 digits is the pounds.
+- For `"399"`, length is 3, so substring(0, 1) → `"3"`.
+
+5) `const pence = paddedPenceNumberString.substring(paddedPenceNumberString.length - 2).padEnd(2, "0");`
+- First, `substring(length - 2)` takes the last 2 digits.
+- Then `padEnd(2, "0")` ensures it’s at least 2 characters (adds zeros on the right if needed).
+- Purpose: get exactly two pence digits.
+- For `"399"`, last two digits are `"99"` → stays `"99"`.
```
cjyuan: Could we expect this program to work as intended for any valid `penceString` if we deleted `.padEnd(2, "0")` from the code?

cjyuan (commented):
Very clear explanation and answers! I just have two follow-up questions.

Sprint-1/1-key-exercises/1-count.js
Richiealx: Hi @cjyuan, 

Thanks for the helpful feedback.

I updated `1-count.js` to include the correct programming term **increment** to describe the operation `count = count + 1`.

Sprint-1/3-mandatory-interpret/answers.md
Richiealx: I also updated my explanation in `answers.md` to clarify that `.padEnd(2, "0")` is not actually necessary, because `paddedPenceNumberString` has already been padded earlier with `padStart(3, "0")`, so `substring(...)` already returns the correct two-digit pence value.

I pushed the changes to the same branch.

cjyuan:
Explanations are spot on. Well done!

illicitonion:

PR #1079, opened 2026-02-26
London | 26-ITP-January | Karla Grajales | Sprint 3 | Practice TDD

Grajales-K:
## Learners, PR Template

Self checklist

## Changelist

This project is about how to implement the test cases first. Once we cover all possibilities, we can implement the function step by step. I learned that if we need to optimize our function, the test cases are much easier to adapt if we avoid naming them 'case 1'. 'case 2', etc. Using `module.export `and `npx jest --watch` command was very helpful to get the real-time feedback on what was happening in our implementation.

## Questions

Is this how the learning approach should be? Should I study the test cases before solving the function, e.g, in Codewars?

Thank you

### DIFF
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -4,12 +4,23 @@ const countChar = require("./count");
 // When the countChar function is called with these inputs,
 // Then it should:
 
+// Scenario: No Occurrences
+// Given the input string `str`,
+// And a character `char` that does not exist within `str`.
+// When the function is called with these inputs,
+// Then it should return 0, indicating that no occurrences of `char` were found.
+test("should return no occurrences if no found character", () => {
+    const str = "jhhfdd";
+    const char = "r";
+    const count = countChar(str, char);
+    expect(count).toEqual(0);
+  });
+
 // Scenario: Multiple Occurrences
 // Given the input string `str`,
 // And a character `char` that occurs one or more times in `str` (e.g., 'a' in 'aaaaa'),
 // When the function is called with these inputs,
 // Then it should correctly count occurrences of `char`.
-
 test("should count multiple occurrences of a character", () => {
   const str = "aaaaa";
   const char = "a";
@@ -17,8 +28,17 @@ test("should count multiple occurrences of a character", () => {
   expect(count).toEqual(5);
 });
 
-// Scenario: No Occurrences
-// Given the input string `str`,
-// And a character `char` that does not exist within `str`.
-// When the function is called with these inputs,
-// Then it should return 0, indicating that no occurrences of `char` were found.
+// case 2
+test("should return 1 when the character appears only once", () => {
+  expect(countChar("hello", "h")).toEqual(1);
+});
+
+// case 3
+test("Should return 0 for empty string", () => {
+  expect(countChar("", "a")).toEqual(0)
+}) 
+
+// case 4
+test("should be case sensitive", () => {
+  expect(countChar("aA", "a")).toEqual(1);
+});
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -42,3 +42,12 @@ test("Should return 0 for empty string", () => {
 test("should be case sensitive", () => {
   expect(countChar("aA", "a")).toEqual(1);
 });
+
+//case 5 Non-alphabet characters
+test(`Should work with numbers and symbols`, () => {
+  expect(countChar("123451234123!&", "1")).toEqual(3);
+  expect(countChar("@1hello@", "@")).toEqual(2);
+  expect(countChar("white space ", " ")).toEqual(2);
+  expect(countChar("-2534!-1322!", "!")).toEqual(2);
+  expect(countChar("-hello ", "-")).toEqual(1);
+})
--- a/Sprint-3/2-practice-tdd/repeat-str.js
+++ b/Sprint-3/2-practice-tdd/repeat-str.js
@@ -1,8 +1,8 @@
 function repeatStr(str, count) {
   if (count < 0) {
-    throw new Error("count cannot be negative");
+    throw new Error("count cannot be negative 😭");
   }
-  
+
   let result = "";
   for (let i = 0; i < count; i++) {
     result = result + str;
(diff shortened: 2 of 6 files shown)


Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -4,21 +4,41 @@ const countChar = require("./count");
...
+
+// case 3
+test("Should return 0 for empty string", () => {
+  expect(countChar("", "a")).toEqual(0)
+}) 
+
+// case 4
+test("should be case sensitive", () => {
+  expect(countChar("aA", "a")).toEqual(1);
+});
```
cjyuan: Including a test to show match needs to be case sensitive is very thoughtful.

Could consider including a case to show that the function is expected to work also for non-alphabets.

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,31 @@
...
-  return "hellohellohello";
+function repeatStr(str, count) {
+  if (count < 0) {
+    throw new Error("count cannot be negative 😭");
+  }
+
+  let result = "";
+  for (let i = 0; i < count; i++) {
+    result = result + str;
+  }
```
cjyuan: No change required, but can you look up a built-in string method that we can use to replace the code on lines 6-9? Built-in method is usually optimised for performance. In practice, we should use them whenever possible.

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,31 @@
...
+// to verify it's working as expected without breaking the test suite.
+test("Should log the specific error message to the console", () => {
+  try {
+    repeatStr("hi", -3);
+  } catch (error) {
+    console.log("Verified Error Message ->", error.message);
+
+    // Check if the message is exactly what we defined
+    expect(error.message).toBe("count cannot be negative 😭");
+  }
```
cjyuan: If we are using Jest, then we should use Jest's approach (line 30) to test.

The `toThrow()` method accepts an optional parameter which you can use to validate the error message. 
See: https://jestjs.io/docs/expect#tothrowerror

cjyuan (commented):
> Is this how the learning approach should be? Should I study the test cases before solving the function, e.g, in Codewars?

Yes. It is not just a learning process; it is "the way".

Even without actually implementing the tests first, it is still the approach to take. That is, think about all possible input and the expected output before implementing the function.

If test cases are available, study them first.

Sprint-3/2-practice-tdd/repeat-str.js
cjyuan: Please note that not all editors can handle emoji characters, and many terminals cannot (or are not setup to) display them properly. It is best to avoid using emoji characters unless they are necessary.

Sprint-3/2-practice-tdd/count.test.js
Grajales-K: @cjyuan Thanks for suggesting this. I have implemented no-alphabet cases.

Sprint-3/2-practice-tdd/repeat-str.js
Grajales-K: Thank you, @cjyuan 
I refactored the function to optimise it and improve readability. I initially implemented a loop to better understand the execution flow.

Sprint-3/2-practice-tdd/repeat-str.js
Grajales-K: @cjyuan, oh, I understand better now. I commented out just to leave for my notes, and see the workflow manually of try/catch.

I removed the emojis for better readability.

Grajales-K:
> > Is this how the learning approach should be? Should I study the test cases before solving the function, e.g, in Codewars?
> 
> Yes. It is not just a learning process; it is "the way".
> 
> Even without actually implementing the tests first, it is still the approach to take. That is, think about all possible input and the expected output before implementing the function.
> 
> If test cases are available, study them first.


Hi @cjyuan, Thank you once again for your time and for the review of my PR. I have made the suggested changes, and I will continue working on improving my confidence and skills when writing test cases.

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,46 @@
...
+//   }
+//   return result;
+// }
+
+// The repeat() method of String values constructs and returns 
+// a new string which contains the specified number of copies 
+// of this string, concatenated together.
+//repeatStr("abc", 3); // "abcabcabc" 
+
+//refactored function using built-in method
```
cjyuan: I noticed some inconsistently formatted comments.

Have you installed **prettier** VSCode extension and enabled formatting on save/paste on VSCode
as recommended in 
?

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,46 @@
...
+//   if (count < 0) {
+//     throw new Error("count cannot be negative");
+//   }
+
+//   let result = "";
+//   for (let i = 0; i < count; i++) {
+//     result = result + str;
+//   }
+//   return result;
+// }
```
cjyuan: I understand that you wanted to keep these code for reference.

In practice, however, it is considered a poor practice to keep any unnecessary code in a PR branch. One of the reasons is that it makes reviewing code harder.

Can you keep the code in this file clean?

Sprint-3/2-practice-tdd/repeat-str.js
Grajales-K: Yes, I have it and checked it, but it is not changing the format. Thank you, I have removed the unnecessary code and function

Sprint-3/2-practice-tdd/repeat-str.js
Grajales-K: Removed thank you @cjyuan

Sprint-3/2-practice-tdd/repeat-str.js
cjyuan: If you have enabled "Format on save" but it is not working, it is likely that you haven't assign a formatter for JS file. This could happen if you have zero or multiple extensions that can format .js file.

(1) Use "Format document" to format the JS file. Sometimes, VSCode will ask you to choose a formatter, and you can manually select "Prettier".
(2) Edit `settings.json` and set Prettier as the default formatter for JS.

cjyuan:
All good now. Well done.

illicitonion:

PR #1099, opened 2026-02-28
London | 26-ITP-Jan  |  Ebrahim Moqbel | Sprint 2 | coursework/sprint-2

Ebrahim-Moqbel:
## Learners, PR Template

Self checklist

- [ x] I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
- [ x] My changes meet the requirements of the task
- [ x] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/)

## Changelist

I completed the following filles :
- key errors 
- mandatory debug 
- mandatory implement
- mandatory interpret

### DIFF
--- a/Sprint-2/2-mandatory-debug/2.js
+++ b/Sprint-2/2-mandatory-debug/2.js
@@ -2,23 +2,41 @@
 
 // Predict the output of the following code:
 // =============> Write your prediction here
+//getLastDigit will convert a number to a string and reverse it 
+// However, since there is no parameter or local variable we might run into an Error 
 
-const num = 103;
+// const num = 103;
 
-function getLastDigit() {
-  return num.toString().slice(-1);
-}
+// function getLastDigit() {
+//   return num.toString().slice(-1);
+// }
 
-console.log(`The last digit of 42 is ${getLastDigit(42)}`);
-console.log(`The last digit of 105 is ${getLastDigit(105)}`);
-console.log(`The last digit of 806 is ${getLastDigit(806)}`);
+// console.log(`The last digit of 42 is ${getLastDigit(42)}`);
+// console.log(`The last digit of 105 is ${getLastDigit(105)}`);
+// console.log(`The last digit of 806 is ${getLastDigit(806)}`);
 
 // Now run the code and compare the output to your prediction
 // =============> write the output here
+// The last digit of 42 is 3
+// The last digit of 105 is 3
+// The last digit of 806 is 3
 // Explain why the output is the way it is
 // =============> write your explanation here
+// As expected, the function has no parameters, so it relies on the global variable "num" as its only variable.
+// Since there are no parameters defined, the function permanently uses the global variable instead.
+// This means even if we pass arguments when calling the function, they will be ignored because there are no parameters to receive them.
+
 // Finally, correct the code to fix the problem
 // =============> write your new code here
+
+function getLastDigit(num) {
+  return num.toString().slice(-1);
+ }
+
+ console.log(`The last digit of 42 is ${getLastDigit(42)}`);
+ console.log(`The last digit of 105 is ${getLastDigit(105)}`);
+ console.log(`The last digit of 806 is ${getLastDigit(806)}`);
+
 
 // This program should tell the user the last digit of each number.
 // Explain why getLastDigit is not working properly - correct the problem
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -15,5 +15,10 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
+    let result = weight / (height*height);
+    return result.toFixed(1);
+    
     // return the BMI of someone based off their weight and height
 }
+
+console.log(calculateBMI(70,168))
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -15,10 +15,11 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    let result = weight / (height*height);
+    let result = Number(weight/(height*height));
     return result.toFixed(1);
     
-    // return the BMI of someone based off their weight and height
+
 }
 
-console.log(calculateBMI(70,168))
+console.log(calculateBMI(70,1.7))
+console.log(calculateBMI(72,1.7))
--- a/Sprint-2/4-mandatory-interpret/time-format.js
+++ b/Sprint-2/4-mandatory-interpret/time-format.js
@@ -11,24 +11,28 @@ function formatTimeDisplay(seconds) {
   return `${pad(totalHours)}:${pad(remainingMinutes)}:${pad(remainingSeconds)}`;
 }
 
+console.log(formatTimeDisplay(60))
 // You will need to play computer with this example - use the Python Visualiser https://pythontutor.com/visualize.html#mode=edit
 // to help you answer these questions
 
 // Questions
 
 // a) When formatTimeDisplay is called how many times will pad be called?
-// =============> write your answer here
+// =============> 3 times
 
 // Call formatTimeDisplay with an input of 61, now answer the following:
 
 // b) What is the value assigned to num when pad is called for the first time?
-// =============> write your answer here
+// =============> the first time pad was called with the totalHours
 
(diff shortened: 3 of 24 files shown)


Sprint-2/2-mandatory-debug/2.js
```diff
@@ -2,23 +2,41 @@
 
 // Predict the output of the following code:
 // =============> Write your prediction here
+//getLastDigit will convert a number to a string and reverse it 
+// However, since there is no parameter or local variable we might run into an Error 
 
-const num = 103;
+// const num = 103;
```
cjyuan, line 8: Deleting the global `num` is optional. Within the function block, `num` is resolved to the parameter `num`.
If you are interested in the topic, you can looking up these two concepts, ***identifier scope*** and ***identifier resolution***, in the context of JavaScript programming. ChatGPT can give a good explanation.

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,11 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    let result = Number(weight/(height*height));
```
cjyuan, line 18: Do you expect different results from these two statements?
```
  let result = Number(weight/(height*height));
```
```
  let result = weight/(height*height);
```

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,11 @@
...
+    return result.toFixed(1);
```
cjyuan, line 19: What **type** of value do you expect your function to return? A number or a string?

```javascript
  
```

Sprint-2/4-mandatory-interpret/time-format.js
```diff
@@ -11,24 +11,28 @@ function formatTimeDisplay(seconds) {
...
 
 // c) What is the return value of pad is called for the first time?
-// =============> write your answer here
+// =============> "00"
+
 
 // d) What is the value assigned to num when pad is called for the last time in this program?  Explain your answer
-// =============> write your answer here
+// =============> "1"
+// the last time, pad() is called with the remaining seconds which is value "1"
```
cjyuan, line 34: What is the data type of this value? Is it a string or a number?

Sprint-2/4-mandatory-interpret/time-format.js
```diff
@@ -11,24 +11,28 @@ function formatTimeDisplay(seconds) {
...
 
 // a) When formatTimeDisplay is called how many times will pad be called?
-// =============> write your answer here
+// =============> 3 times
 
 // Call formatTimeDisplay with an input of 61, now answer the following:
 
 // b) What is the value assigned to num when pad is called for the first time?
-// =============> write your answer here
+// =============> the first time pad was called with the totalHours
```
cjyuan, line 26: What is the value assigned to `num` when `pad()` is called for the first time?

cjyuan (commented):
Code is not consistently formatted.

Consider install **prettier** VSCode extension and enable formatting on save/paste on VSCode
as recommended in 

Then, resave all your files (in Sprint 2) to ensure all code is consistently formatted.

Sprint-2/2-mandatory-debug/2.js
Ebrahim-Moqbel: Thank you Cjyuan

Sprint-2/3-mandatory-implement/1-bmi.js
Ebrahim-Moqbel: I was expecting the same result but I thought I could right Number just in case but it seems unnecessary to use in here. Thanks xx

Sprint-2/3-mandatory-implement/1-bmi.js
Ebrahim-Moqbel: I was actually expecting a number but I was validating the user input if they enter a string. However, using Number here seems unnecessary to use.

Sprint-2/4-mandatory-interpret/time-format.js
Ebrahim-Moqbel: The output of the data type would be a string as I have convert the parameter num to a string with the function pad.

Sprint-2/4-mandatory-interpret/time-format.js
Ebrahim-Moqbel: the value would be any number

Ebrahim-Moqbel:
I have installed prettier to VS code and selected the copy-paste  save  format. I have been reading about reverting and reseting commits in git. I know how to. revert a specific commit but I still need to know how can I select all the commits. 

I was wondering if you need me revert all of them and commit them again. Also After installing prettier I still find the format the same and I think it's consistent.

Sprint-2/3-mandatory-implement/1-bmi.js
cjyuan: Can you elaborate what you meant by "... using Number here seems unnecessary to use"?

I was asking what **type** of value do you expect from this expression, `result.toFixed(1)`. That is the value the function is going to return.

Sprint-2/4-mandatory-interpret/time-format.js
cjyuan: You may have misunderstood the question. 
> (d) What is the value assigned to `num` when pad is called for the last time in this program?

The `num` in the question refers to the **parameter** of the function.

```javascript
function pad(num) {
  return num.toString().padStart(2, "0");
}
```

If the function is called in the following way:
```
  const a = pad(23);
```
then when the function is called, we say the value assigned to `num` is 23.

The converted value, "23", is the return value of the function.


Can you revise your answer to question (d)?

Sprint-2/4-mandatory-interpret/time-format.js
cjyuan: What do you meant by "any number"? I was rephrasing question (b):
> What is the value assigned to `num` when pad is called for the first time?

Sprint-2/3-mandatory-implement/1-bmi.js
cjyuan: What do you mean by "right Number"

Ebrahim-Moqbel: Sorry, plz disregard the comment. What I was trying to say is that I was using Number just for validation but it seems it is  unnecessary.

cjyuan:
If you have enabled "Format on save" but it is not working, it is likely that you haven't assign a formatter for JS file. This could happen if you have **zero** or **multiple extensions** that can format .js file.

(1) Use "Format document" to format the JS file. Sometimes, VSCode will ask you to choose a formatter, and you can manually select "Prettier".
(2) Edit `settings.json` and set Prettier as the default formatter for JS.

--- 
Can you also address all the follow-up questions in the inline comments.

Sprint-2/3-mandatory-implement/1-bmi.js
Ebrahim-Moqbel: .toFixed would return the number to a fixed one decimal place and the value would be a string.

Sprint-2/4-mandatory-interpret/time-format.js
Ebrahim-Moqbel: The value assigned is a number and a return would be string.

Ebrahim-Moqbel:
I have addressed the questions above and regarding prettier I have installed it and selected the the format to be on paste as well but I may still need to Edit settings.json but I am still not sure how to access to this settings whether they are in an external file or in built settings in VS code. 

In the bellow video prettier asked for configuration and I am hoping you could advise me what I expected to do at this stage. Thanks xx

Screenshot 2026-02-26 at 20 57 19 (2)

Sprint-2/3-mandatory-implement/1-bmi.js
cjyuan: Thanks for clarification.

cjyuan:
I don't see any video. I only see an image.

Prettier won't format the code if there is a syntax error in the code. 

The file `settings.json` is located at   
`$HOME/Library/Application\ Support/Code/User/settings.json`.

If you still have problem setting Prettier as your default formatter, try to seek for assistance on Saturday in the workshop?

---
Thanks for the responses to each comment. All good now.

illicitonion:

PR #1101, opened 2026-02-28
Manchester | 26-ITP-Jan | Liban Jama | Sprint 2 | Structuring and Testing Data

libanj0161:
## Learners, PR Template

Self checklist

## Changelist

Completed sprint 2 coursework.

## Questions

N/A

### DIFF
--- a/Sprint-2/1-key-errors/2.js
+++ b/Sprint-2/1-key-errors/2.js
@@ -4,17 +4,20 @@
 // this function should square any number but instead we're going to get an error
 
 // =============> write your prediction of the error here
-
+// You can't put a number value into an argument.
 function square(3) {
     return num * num;
 }
 
 // =============> write the error message here
-
+// SyntaxError: Unexpected number
 // =============> explain this error message here
-
+// It is not expecting a number to be in the function argument.
 // Finally, correct the code to fix the problem
 
 // =============> write your new code here
+function square(num) {
+    return num * num;
+}
 
-
+console.log("square",square(5))
--- a/Sprint-2/1-key-errors/2.js
+++ b/Sprint-2/1-key-errors/2.js
@@ -4,7 +4,7 @@
 // this function should square any number but instead we're going to get an error
 
 // =============> write your prediction of the error here
-// You can't put a number value into an argument.
+// You can't put a number value into a parameter.
 function square(3) {
     return num * num;
 }
@@ -12,7 +12,7 @@ function square(3) {
 // =============> write the error message here
 // SyntaxError: Unexpected number
 // =============> explain this error message here
-// It is not expecting a number to be in the function argument.
+// It is not expecting a number to be in the function parameter.
 // Finally, correct the code to fix the problem
 
 // =============> write your new code here
@@ -20,4 +20,3 @@ function square(num) {
     return num * num;
 }
 
-console.log("square",square(5))
--- a/Sprint-2/2-mandatory-debug/0.js
+++ b/Sprint-2/2-mandatory-debug/0.js
@@ -1,7 +1,9 @@
 // Predict and explain first...
 
 // =============> write your prediction here
-// the function will not return anything.
+
+// the function will not return anything, as there is no return.
+
 function multiply(a, b) {
   console.log(a * b);
 }
@@ -10,7 +12,7 @@ console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
 
 // =============> write your explanation here
 
-// declared a variable called result and stored a*b.
+// declared a variable called result and stored a*b, then return result in the function.
 
 // Finally, correct the code to fix the problem
 //  =============> write your new code here
--- a/Sprint-2/2-mandatory-debug/0.js
+++ b/Sprint-2/2-mandatory-debug/0.js
@@ -1,7 +1,7 @@
 // Predict and explain first...
 
 // =============> write your prediction here
-
+// the function will not return anything.
 function multiply(a, b) {
   console.log(a * b);
 }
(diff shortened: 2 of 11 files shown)


Sprint-2/1-key-errors/2.js
```diff
@@ -4,17 +4,20 @@
 // this function should square any number but instead we're going to get an error
 
 // =============> write your prediction of the error here
-
+// You can't put a number value into an argument.
```
tee4tao: You mentioned 'argument', but the correct term for what goes inside the parentheses when you create a function is a Parameter. So, in this case it's a parameter.

Sprint-2/2-mandatory-debug/0.js
tee4tao, line 1: You've made a correction to the code but I can't see your explanation.

If you could kindly explain what's wrong with the original code and the what you did to correct it.

Sprint-2/2-mandatory-debug/1.js
```diff
@@ -1,13 +1,20 @@
...
 function sum(a, b) {
   return;
-  a + b;
+a + b;
 }
 
 console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
 
 // =============> write your explanation here
+// The sum of 10 and 32 is undefined. On line five the function has been exited because of the ; this because the function doesn't return anything.
```
tee4tao: You mentioned the function exits specifically because of the ;

What happens if you delete that ; after return, but leave a + b on the next line? Give it a try!

(Hint: Look up Automatic Semicolon Insertion (ASI) in JavaScript to see why the line break itself. Any code below a return statement becomes "unreachable code".)

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -16,4 +16,13 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    console.log("weight", weight);
+    console.log("height",height);
+
+    const heightSquared = (height * height).toFixed(2);
```
tee4tao: Great job breaking the steps down! Quick question on this line: do you know what data type **.toFixed()** returns? Think about how that might affect the division operation on the next line. Is it better to round values in the middle of an equation, or only at the very end?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -16,4 +16,13 @@
...
+    console.log("heightSquared",heightSquared);
+    const BMI = (weight / heightSquared).toFixed(2);
```
tee4tao: Take another quick look at the task instructions regarding the decimal places. Are we returning the correct amount here? Also, keep in mind that **.toFixed()** will return a string. If the function is expected to return a Number, how could you convert it?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -16,4 +16,13 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    console.log("weight", weight);
+    console.log("height",height);
```
tee4tao: These logs are awesome for testing and debugging! However, for a clean final submission, what should we do with them?

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,10 @@
 // You will need to come up with an appropriate name for the function
 // Use the MDN string documentation to help you find a solution
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
+
+function makeUpperCaseSnakeCase(str){
+const result = str.split(" ").join("_").toUpperCase(); //take str, and turn it into an array and then turn it back into a string with underscores then upper case it. 
+console.log("result",result);  
+}
```
tee4tao, line 22: Your string manipulation logic is perfect! However, try running this exact code. You'll notice the final `console.log` at the bottom prints undefined. If a function calculates a value but doesn't explicitly hand it back to where it was called, what happens? How can we fix this?

Sprint-2/3-mandatory-implement/2-cases.js
tee4tao, line 1: This works great! Since the instructions pointed you toward the MDN String documentation, take a peek at the **.replaceAll()** method. How could you use that to achieve the same result without needing to convert the string into an array first?

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,10 @@
 // You will need to come up with an appropriate name for the function
 // Use the MDN string documentation to help you find a solution
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
+
+function makeUpperCaseSnakeCase(str){
+const result = str.split(" ").join("_").toUpperCase(); //take str, and turn it into an array and then turn it back into a string with underscores then upper case it. 
+console.log("result",result);
```
tee4tao: Make sure to clean up your debugging logs before your final submission!

Sprint-2/3-mandatory-implement/3-to-pounds.js
tee4tao, line 1: Can you please read the instruction in the assessment? You need to follow the instructions and make use of the starter code stated.

libanj0161:
@tee4tao i've made the requested changes.

tee4tao (commented):
Well-done!

illicitonion:

PR #1110, opened 2026-02-28
London | 26-ITP-Jan | Angela McLeary | Sprint 2 | Coursework/sprint 2

AngelaMcLeary:
## Learners, PR Template

Self checklist

## Changelist

It is about debugging and understanding how logic works.

### DIFF
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -37,8 +37,7 @@
 // =============> write your new code here
 
 function capitalise(str) {
-  str = `${str[0].toUpperCase()}${str.slice(1)}`;
-  return str;
+  return `${str[0].toUpperCase()}${str.slice(1)}`;
 }
 console.log(capitalise("justice"));
 
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -1,8 +1,25 @@
 // Predict and explain first...
-//  =============> write your prediction here
+//Answer
+//  We have declared the "str" twice.It is already declared 
+//  in this function parameter, hence causing the conflict. 
+//  Although "let" allows re-assignment of variables, the issue is redeclaration, 
+//  not reassignment.
+
 
 // call the function capitalise with a string input
+//Answer
+/// home/justice/Documents/CYF/Module-Structuring-and-Testing-Data/Sprint-2/1-key-errors/tempCodeRunnerFile.js:2
+//  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+//      ^
+
+//SyntaxError: Identifier 'str' has already been declared
+
+
 // interpret the error message and figure out why an error is occurring
+//Answer
+
+
+
 
 function capitalise(str) {
   let str = `${str[0].toUpperCase()}${str.slice(1)}`;
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -1,5 +1,6 @@
 // Predict and explain first...
-//Answer
+//  =============> write your prediction here
+//  1.*Answer
 //  We have declared the "str" twice.It is already declared 
 //  in this function parameter, hence causing the conflict. 
 //  Although "let" allows re-assignment of variables, the issue is redeclaration, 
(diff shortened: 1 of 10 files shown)


Sprint-2/1-key-errors/0.js
```diff
@@ -1,13 +1,45 @@
...
+
+// =============> write your explanation here
+//I have re-assigned the "str" variable and it now runs with no errors.
+// I also checked with console.log to see the output.
+
+// =============> write your new code here
 
 function capitalise(str) {
-  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+  str = `${str[0].toUpperCase()}${str.slice(1)}`;
```
tee4tao: In small examples this work, but do you think modifying function parameters is a good practice in larger codebases?

Could you rewrite this function in a way that avoids mutating the parameter entirely?

Sprint-2/1-key-errors/1.js
```diff
@@ -1,20 +1,36 @@
...
+// Removed the const declaration and returned the percentage calculation
+// using a template literal to append the percent symbol.
 
 // Finally, correct the code to fix the problem
 // =============> write your new code here
+  function convertToPercentage(decimalNumber) {
+    return `${decimalNumber * 100}%`;
+  }
+
+  console.log(decimalNumber);
```
tee4tao: Why do you still have this here?

Sprint-2/2-mandatory-debug/0.js
```diff
@@ -1,14 +1,30 @@
...
 // =============> write your explanation here
+// It return the following output:
+// 320
+// The result of multiplying 10 and 32 is undefined
+
+// why?
+//the computer reads the values in the second console.log "multiply(10, 32)" and
+//the first console.log only prints the value, it does not store it.
+//the second console.log outputs the template literals and undefined
+// as it has no value to reach.
```
tee4tao, line 20: In your explanation, you mentioned that the second **console.log** outputs **undefined** 'as it has no value to reach.' You are exactly right! The technical term for this in JavaScript is an **implicit return**. If you don't explicitly tell a function what to **return**, JavaScript automatically forces it to return **undefined** behind the scenes. Keep up the great work!

Sprint-2/2-mandatory-debug/2.js
```diff
@@ -1,24 +1,54 @@
 // Predict and explain first...
+// The function does not take in any parameters, therefore it 
+// refers to the global const variable = 103 
+// So every console.log will print 3 every time. This function is not 
+// re-useable as it stands.
```
tee4tao, line 5: Good point and observation. This function is not re-useable as it stands, functions should ideally be self-contained and reusable.

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -13,7 +13,15 @@
...
 // Then when we call this function with the weight and height
 // It should return their Body Mass Index to 1 decimal place
+// bmi = weight / height**2).
 
+ // return the BMI of someone based off their weight and height
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    return  (weight/(Math.pow(height, 2))).toFixed(1);
```
tee4tao: Using **.toFixed()** converts your datatype to **String**. How might you convert that final answer back into a true Number before returning it?

You used **Math.pow(height, 2)** to square the height, which is 100% correct. However, modern JavaScript has a shorthand operator for exponents that makes equations look a bit cleaner.

Sprint-2/1-key-errors/0.js
AngelaMcLeary: @tee4tao thank you for your feedback.
I have updated the code to return the value without declaring **str**.
Thank you,
Angela.

Sprint-2/1-key-errors/1.js
AngelaMcLeary: Hi @tee4tao thank you for your feedback.  
That was an error, it should have been  **console.log(convertToPercentage(0.5));**. I have corrected this.
Thank you
Angela,

Sprint-2/2-mandatory-debug/0.js
AngelaMcLeary: @tee4tao Thank you for the feedback. I will note that for future explanations. 
Thank you
Angela.

Sprint-2/2-mandatory-debug/2.js
AngelaMcLeary: @tee4tao 
Thank you.
Angela.

Sprint-2/3-mandatory-implement/1-bmi.js
AngelaMcLeary: @tee4tao , Thanks for the questions and feedback.

**1.**  I have use JavaScript's inbuilt **Number** function to convert the string to a number: return **Number**((weight / (height ** 2)).toFixed(1)); This should fix the problem of returning a string from using .**toFixed**.

**2.** The other way to square the height with modern JavaScript by using "**" instead of Math.pow.
(weight / (**height ** 2**); This should bring thr code uptodate with modern styles. I have also updated the code accordingly.

Thank you
Angela,

illicitonion:

PR #1117, opened 2026-03-01
Birmingham | 26-ITP-Jan | Arun Kumar Akilan | Sprint 2 | Structuring-and-Testing-Data Coursework/Sprint2

arunkumarakilan:
### Self checklist

---

## Changelist

This PR contains my completed Sprint 2 coursework for the Structuring and Testing Data module.

### Summary of changes:

- Completed all mandatory debugging exercises
- Fixed logical errors in existing functions
- Refactored code into reusable helper functions
- Implemented string and number formatting functions
- Handled edge cases such as:
  - Converting 24-hour time to 12-hour format
  - Ensuring consistent two-digit formatting using `padStart`
- Ensured functions return correct values instead of relying on console output
- Tested all exercises to confirm expected behavior

All tasks have been completed and verified to work correctly.

---

## Questions

No questions at this time.

### DIFF
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -20,8 +20,8 @@ function capitalise(str) {
 // This avoids the redeclaration error and the code runs correctly.
 // =============> write your new code here
 function capitalise(str) {
-  str = `${str[0].toUpperCase()}${str.slice(1)}`;
-  return str;
+  const capitalised  = `${str[0].toUpperCase()}${str.slice(1)}`;
+  return capitalised;
 }
 let str = capitalise("Arun");
 console.log(str);
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -1,5 +1,9 @@
-// Predict and explain first...
+// Predict and explain first...  
+//Syntax error
 //  =============> write your prediction here
+// We pass str as a parameter to the function.
+//The parameter becomes a local variable inside the function scope.
+//Therefore, we cannot declare another variable with the same name inside the same scope.
 
 // call the function capitalise with a string input
 // interpret the error message and figure out why an error is occurring
@@ -10,4 +14,16 @@ function capitalise(str) {
 }
 
 // =============> write your explanation here
+// The parameter str is already a local variable inside the function.
+// Previously, redeclaring it using let caused an error because variables cannot be declared twice in the same scope.
+// Instead of redeclaring it, I modified the existing str variable and returned it.
+// This avoids the redeclaration error and the code runs correctly.
 // =============> write your new code here
+function capitalise(str) {
+  str = `${str[0].toUpperCase()}${str.slice(1)}`;
+  return str;
+}
+let str = capitalise("Arun");
+console.log(str);
+
+
--- a/Sprint-2/1-key-errors/1.js
+++ b/Sprint-2/1-key-errors/1.js
@@ -36,5 +36,5 @@ function convertToPercentage(decimalNumber) {
   return percentage;
 }
 
-const decimalNumber = convertToPercentage(0.5) ;
-console.log(decimalNumber);
+const percentageValue = convertToPercentage(0.5) ;
+console.log(percentageValue);
--- a/Sprint-2/1-key-errors/1.js
+++ b/Sprint-2/1-key-errors/1.js
@@ -1,6 +1,12 @@
 // Predict and explain first...
-
+// syntax error identifier decimalNumber be already declare
+// there is no function call and then you can't use the local variable decimalNumber out of the function
 // Why will an error occur when this program runs?
+//decimalNumber is already declared as a parameter, so declaring it again inside the function causes an error.
+
+//There is no function call.
+
+//You cannot use the local variable decimalNumber outside the function scope.
 // =============> write your prediction here
 
 // Try playing computer with the example to work out what is going on
@@ -11,10 +17,24 @@ function convertToPercentage(decimalNumber) {
 
   return percentage;
 }
-
 console.log(decimalNumber);
 
 // =============> write your explanation here
+//The error occurred because a local variable was redeclared inside the function using a variable keyword. Since the parameter already acts as a local variable, redeclaring it caused an error.
 
+//To fix this, I removed the variable keyword and modified the existing parameter instead.
+
+//I also removed the unused variable from console.log because it was outside the function scope.
+
+//Finally, I called the function directly inside console.log using convertToPercentage() so the function executes properly and returns the correct value.
 // Finally, correct the code to fix the problem
 // =============> write your new code here
+function convertToPercentage(decimalNumber) {
+  
+  const percentage = `${decimalNumber * 100}%`;
+
+  return percentage;
+}
+
+const decimalNumber = convertToPercentage(0.5) ;
+console.log(decimalNumber);
--- a/Sprint-2/1-key-errors/2.js
+++ b/Sprint-2/1-key-errors/2.js
@@ -1,20 +1,44 @@
 
 // Predict and explain first BEFORE you run any code...
 
+// The code has a syntax error because a number is used instead of a parameter name in the function definition.
+// Additionally, num is used inside the function without being declared or passed as an argument, which causes a ReferenceError.
+// If the function is not called, it will not execute.
 // this function should square any number but instead we're going to get an error
 
 // =============> write your prediction of the error here
+
+// The code has a syntax error because a number is used instead of a parameter name in the function definition.
+// Additionally, num is used inside the function without being declared or passed as an argument, which causes a ReferenceError.
+// If the function is not called, it will not execute.
 
 function square(3) {
     return num * num;
 }
 
+
+
 // =============> write the error message here
+// syntax error :unexpected number.
 
 // =============> explain this error message here
+// The error occurs because a number is used in the function definition instead of a parameter name. A function definition must contain a parameter (a variable name), not a value.
+
+// Inside the function, num is used but it was never declared or passed as a parameter, which causes a ReferenceError because num is not defined in the scope.
+
+// Also, if the function is not called, it will not execute.
 
 // Finally, correct the code to fix the problem
+// I corrected the function by properly defining num as a parameter instead of using a number in the function definition. This resolved the syntax error.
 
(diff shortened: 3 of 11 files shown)


Sprint-2/1-key-errors/0.js
```diff
@@ -1,5 +1,9 @@
-// Predict and explain first...
+// Predict and explain first...  
+//Syntax error
```
tee4tao, line 2: Is this just a generic syntax issue, or is JavaScript specifically telling you something about variable declarations?

Sprint-2/1-key-errors/0.js
```diff
@@ -10,4 +14,16 @@ function capitalise(str) {
...
 
 // =============> write your explanation here
+// The parameter str is already a local variable inside the function.
+// Previously, redeclaring it using let caused an error because variables cannot be declared twice in the same scope.
+// Instead of redeclaring it, I modified the existing str variable and returned it.
+// This avoids the redeclaration error and the code runs correctly.
 // =============> write your new code here
+function capitalise(str) {
+  str = `${str[0].toUpperCase()}${str.slice(1)}`;
+  return str;
```
tee4tao: You reassigned the parameter instead of redeclaring it.
In small examples this work, but do you think modifying function parameters is a good practice in larger codebases? Why or Why not?

Could you rewrite this function in a way that avoids mutating the parameter entirely?

Sprint-2/1-key-errors/0.js
```diff
@@ -1,5 +1,9 @@
-// Predict and explain first...
+// Predict and explain first...  
+//Syntax error
 //  =============> write your prediction here
+// We pass str as a parameter to the function.
+//The parameter becomes a local variable inside the function scope.
```
tee4tao, line 5: If parameters behave like local variables, why does **let** cause an error here but **var** behaves differently? What does that tell you about scope rules?

Sprint-2/1-key-errors/1.js
```diff
@@ -11,10 +17,24 @@ function convertToPercentage(decimalNumber) {
 
   return percentage;
 }
-
 console.log(decimalNumber);
 
 // =============> write your explanation here
+//The error occurred because a local variable was redeclared inside the function using a variable keyword. Since the parameter already acts as a local variable, redeclaring it caused an error.
+
+//To fix this, I removed the variable keyword and modified the existing parameter instead.
```
tee4tao, line 25: In your explanation here, you said you modified the parameter instead of redeclaring it.
But in your final code (which is good), you didn’t modify the parameter, you removed the redeclaration.
Can you clarify what actually changed?

Sprint-2/1-key-errors/1.js
```diff
@@ -11,10 +17,24 @@ function convertToPercentage(decimalNumber) {
...
 // Finally, correct the code to fix the problem
 // =============> write your new code here
+function convertToPercentage(decimalNumber) {
+  
+  const percentage = `${decimalNumber * 100}%`;
+
+  return percentage;
+}
+
+const decimalNumber = convertToPercentage(0.5) ;
```
tee4tao: You name your variable **decimalNumber** which now holds a percentage string.
Does that variable name still accurately describe the value it contains?

Sprint-2/1-key-errors/1.js
```diff
@@ -1,6 +1,12 @@
 // Predict and explain first...
-
+// syntax error identifier decimalNumber be already declare
+// there is no function call and then you can't use the local variable decimalNumber out of the function
 // Why will an error occur when this program runs?
+//decimalNumber is already declared as a parameter, so declaring it again inside the function causes an error.
+
+//There is no function call.
+
+//You cannot use the local variable decimalNumber outside the function scope.
```
tee4tao, line 9: You mentioned the variable cannot be used outside the function.
Can you explain the difference between a **SyntaxError** and a **ReferenceError** in this example?

Sprint-2/1-key-errors/2.js
```diff
@@ -1,20 +1,44 @@
...
+
 // =============> write the error message here
+// syntax error :unexpected number.
 
 // =============> explain this error message here
+// The error occurs because a number is used in the function definition instead of a parameter name. A function definition must contain a parameter (a variable name), not a value.
+
+// Inside the function, num is used but it was never declared or passed as a parameter, which causes a ReferenceError because num is not defined in the scope.
+
+// Also, if the function is not called, it will not execute.
```
tee4tao, line 29: Is not calling a function an error?

Sprint-2/1-key-errors/2.js
```diff
@@ -1,20 +1,44 @@
...
+// Inside the function, num is now defined, so the ReferenceError is fixed.
 
+// I then called the function with an argument and stored the returned value in a variable named num. Although the same variable name is used, the function parameter and the outer variable exist in different scopes, so there is no conflict.
 // =============> write your new code here
 
 
+function square(num) {
+    return num * num;
+}
+let num =  square(3);
```
tee4tao: This works, but reusing the name **num** outside (as a variable) and inside (as a parameter) is confusing; it’s also easy to misuse in larger scopes.

What variable name outside the function would make the code clearer? And to declare the variable, which operator do you think fits best, **let** or **const** and why?

Sprint-2/2-mandatory-debug/0.js
```diff
@@ -1,14 +1,39 @@
...
 
-// Finally, correct the code to fix the problem
+// To fix the issue, I replaced console.log with return.
+
+// By returning a * b, the function now sends the calculated value back to where it was called.
+
+// As a result, the correct value is displayed instead of undefined.
 //  =============> write your new code here
+function multiply(a, b) {
+  return (a * b);
```
tee4tao: This is good but the parentheses around **a * b** aren't needed. It's cleaner without the parentheses

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,9 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+   let bmi = weight / (height * height);
+    bmi = bmi.toFixed(1);
+    return Number(bmi);
```
tee4tao, line 20: You've successfully rounded the number using a string method (**.toFixed**). If we wanted to avoid converting the number to a **string** entirely, can you think of a purely mathematical way to round a decimal to one place using **Math.round()**? (Hint: Think about multiplying and dividing by 10).

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,9 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+   let bmi = weight / (height * height);
```
tee4tao, line 18: Using **height * height** works perfectly for squaring! Out of curiosity, are you familiar with any JavaScript operators or **Math** methods that are specifically designed for calculating exponents?

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,13 @@
 // You will need to come up with an appropriate name for the function
 // Use the MDN string documentation to help you find a solution
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
+
+function upperSnakeCase(str){
+    str = str.toUpperCase()
+
+  return str.replaceAll(" " , "_");
```
tee4tao: Great job finding the **toUpperCase** and **replaceAll** methods! Notice how **toUpperCase()** returns a new string, and then you apply **replaceAll()** to that result on the next line. Since both of these are string methods, can you think of a way to connect or 'chain' them together into a single line of code?

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,13 @@
 // You will need to come up with an appropriate name for the function
 // Use the MDN string documentation to help you find a solution
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
+
+function upperSnakeCase(str){
+    str = str.toUpperCase()
```
tee4tao: Your code works perfectly, but here you are overwriting the original **str** variable that was passed into the function. In professional development, we try to avoid changing our input parameters. If you use method chaining (commented about that below), how could you return the final result without ever reassigning the str variable?

Sprint-2/3-mandatory-implement/3-to-pounds.js
```diff
@@ -4,3 +4,22 @@
...
+const pounds = paddedPenceNumberString.substring(
+  0,
+  paddedPenceNumberString.length - 2
+);
+
+const pence = paddedPenceNumberString
+  .substring(paddedPenceNumberString.length - 2)
+  .padEnd(2, "0");
+  return `£${pounds}.${pence}`;
+}
```
tee4tao: Spot on! You followed the instructions perfectly here. You successfully wrapped the logic into a reusable function, named the parameter exactly what it needed to be. Excellent work!

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -5,7 +5,8 @@
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
   if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+  time = hours - 12;
```
tee4tao: In your new logic, you assigned the calculated number to **time**. While JavaScript allows this, **time** was originally our input string (e.g., '19:00'). Reassigning parameters to completely different data types can sometimes lead to confusing bugs later on. Can you think of a better way to store your calculated **hours - 12** without overwriting the time parameter? Perhaps a new variable name?

tee4tao (changes requested):

Sprint-2/5-stretch-extend/format-time.js
kfklein15: @arunkumarakilan can you please reply to the reviewer? This has been waiting for 4 days now and you should reply asap.

illicitonion:

PR #1123, opened 2026-03-01
London | 26-ITP-Jan | Boualem Larbi Djebbour | sprint 2 | Coursework

djebsoft:
## Learners, PR Template

Self checklist

## Changelist

complete sprint 2 mandatory tasks

## Questions


Sprint-2/2-mandatory-debug/0.js
```diff
@@ -1,14 +1,28 @@
...
+
+console.log(`The result of multiplying 10 and 32 is ${multiply(10, 32)}`);
+
+// an extra notice: the purpose of creating a function is to reuse it when needed
+// I've noticed that the parameters are already assigned to values 10 and 32, so I changed the code to the following:
+function multiply(a, b) {
+  return console.log(
+    `The result of multiplying ` + a + ` and ` + b + ` is ` + a * b
+  );
+}
```
cjyuan: If there is a need to output the string in a fixed format from different "places" in the program, it would make sense to implement a function do so. 

However, what is the purpose of returning the return value of `console.log()`? What value do you expect `console.log()` to return?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,6 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+  return (BMI = (weight / (height * height)).toFixed(1)); // return the BMI of someone based off their weight and height
```
cjyuan: - What is BMI? Why not just return the value of the expression `(weight / (height * height)).toFixed(1)`?

- What **type** of value do you expect your function to return? A number or a string?

```javascript
  
```

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,9 @@
 // You will need to come up with an appropriate name for the function
 // Use the MDN string documentation to help you find a solution
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
+function upperSnakeCase(wordsSet) {
+  let wsSnakeCase = wordsSet.replaceAll(" ", "_");
+  let wsupperSnakeCase = wsSnakeCase.toUpperCase();
+  return console.log(wsupperSnakeCase);
+}
```
cjyuan, line 21: Accordingly to the description on lines 9-10, the function is expected to **return** the converted value, not output the converted value to the console.

Sprint-2/2-mandatory-debug/0.js
djebsoft: we just actually need the function to pass the output to the next line of code after its destroyed
so we need the function to return what its statement evaluate to

Sprint-2/3-mandatory-implement/1-bmi.js
djebsoft: fixed the statement of the value returned
converted the string to a number

Sprint-2/3-mandatory-implement/2-cases.js
djebsoft: thank you for the review
i fixed the expression of the return statement to return the converted value

djebsoft:
thank you for reviewing my work 
I followed your comments and fixed the bugs

Sprint-2/2-mandatory-debug/0.js
cjyuan: Not sure what you meant by "destroyed".

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,6 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+  return Number(weight / (height * height)).toFixed(1); // return the BMI of someone based off their weight and height
```
cjyuan: How can you be sure that the function is returning a number and not a string?

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,9 @@
 // You will need to come up with an appropriate name for the function
 // Use the MDN string documentation to help you find a solution
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
+function upperSnakeCase(wordsSet) {
+  let wsSnakeCase = wordsSet.replaceAll(" ", "_");
+  let wsupperSnakeCase = wsSnakeCase.toUpperCase();
```
cjyuan, line 19: What do you think about this version:`wsUpperSnakeCase`? (No change needed)

Sprint-2/3-mandatory-implement/1-bmi.js
djebsoft: by adding the number() to the expression that contains strings
but after your comment I checked it using typeof and it turns out it's string
then I included the whole expression in the number() by adding parentheses, then it worked
but prettier didn't allow me to save the change so I made it in my github remote repository

Sprint-2/3-mandatory-implement/2-cases.js
djebsoft: I think it could be shortened and embeded in the previous line

cjyuan:
Changes are good. Well done.

illicitonion:

PR #1130, opened 2026-03-02
Cape Town | 26-ITP-Jan | Pretty Taruvinga | Sprint 3 | Implement and rewrite test

Pretty548:
## Learners, PR Template

Self checklist

## Changelist

This PR implements the required functions and rewrites the existing tests for Sprint 3 in the `1-implement-and-rewrite-tests` directory.

All tests were updated to correctly reflect the expected behaviour, and the implementations were written to make the tests pass.

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
@@ -42,6 +42,3 @@ test(`should return "Invalid angle" when angle is 0`, () => {
 test(`should return "Invalid angle" when angle is 360 or greater`, () => {
   expect(getAngleType(360)).toEqual("Invalid angle");
 });
-test(`should return "Invalid angle" when angle is not a number`, () => {
-  expect(getAngleType("not a number")).toEqual("Invalid angle");
-});
--- a/Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
@@ -2,9 +2,6 @@
 // We will use the same function, but write tests for it using Jest in this file.
 const getAngleType = require("../implement/1-get-angle-type");
 
-// TODO: Write tests in Jest syntax to cover all cases/outcomes,
-// including boundary and invalid cases.
-
 // Case 1: Acute angles
 test(`should return "Acute angle" when (0 < angle < 90)`, () => {
   // Test various acute angles, including boundary cases
@@ -14,7 +11,31 @@ test(`should return "Acute angle" when (0 < angle < 90)`, () => {
 });
 
 // Case 2: Right angle
+test(`should return "Right angle" when angle is exactly 90`, () => {
+  expect(getAngleType(90)).toEqual("Right angle");
+});
+
 // Case 3: Obtuse angles
+test(`should return "Obtuse angle" when (90 < angle < 180)`, () => {
+  expect(getAngleType(91)).toEqual("Obtuse angle");
+  expect(getAngleType(179)).toEqual("Obtuse angle");
+});
+
 // Case 4: Straight angle
+test(`should return "Straight angle" when angle is exactly 180`, () => {
+  expect(getAngleType(180)).toEqual("Straight angle");
+});
+
 // Case 5: Reflex angles
+test(`should return "Reflex angle" when (180 < angle < 360)`, () => {
+  expect(getAngleType(181)).toEqual("Reflex angle");
+  expect(getAngleType(359)).toEqual("Reflex angle");
+});
+
 // Case 6: Invalid angles
+test(`should return "Invalid angle" when angle is negative`, () => {
+  expect(getAngleType(-1)).toEqual("Invalid angle");
+});
+test(`should return "Invalid angle" when angle is 360 or greater`, () => {
+  expect(getAngleType(360)).toEqual("Invalid angle");
+});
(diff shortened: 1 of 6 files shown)


cjyuan (commented):
Function implementation looks good.

This exercise has a second part "**2 Rewrite tests with Jest**".

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
```diff
@@ -14,7 +11,31 @@ test(`should return "Acute angle" when (0 < angle < 90)`, () => {
...
+  expect(getAngleType(359)).toEqual("Reflex angle");
+});
+
 // Case 6: Invalid angles
+test(`should return "Invalid angle" when angle is negative`, () => {
+  expect(getAngleType(-1)).toEqual("Invalid angle");
+});
+test(`should return "Invalid angle" when angle is 360 or greater`, () => {
+  expect(getAngleType(360)).toEqual("Invalid angle");
+});
```
cjyuan, line 44: You did good in identifying one of the boundary cases. There are two boundary cases, and -1 is not one of them. Can you also test the other boundary case?

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -4,7 +4,31 @@ const isProperFraction = require("../implement/2-is-proper-fraction");
...
+  expect(isProperFraction(3, 2)).toEqual(false);
+});
+
+test("should return false when numerator is negative", () => {
+  expect(isProperFraction(-1, 2)).toEqual(false);
+});
+
+test("should return true when both numerator and denominator are negative and numerator is smaller than denominator", () => {
+  expect(isProperFraction(-1, -2)).toEqual(true);
+});
```
cjyuan: The test description on line 28 is not quite accurate. 

For proper fractions, the condition is, "the absolute value of denominator is larger than the absolute value of the numerator". So we can create a category for proper fractions and describe it as:
  `should return true when abs(numerator) < abs(denominator)`

and then test all combinations of positive/negative numerator and denominator (for proper fraction) in this category.

cjyuan (commented):
Please note that in CYF courses, the recommended way to inform the reviewer of your changes is to do both of the following:
  - **Reply to their feedback.**
    - In the responses, clarify how each piece of feedback was addressed to demonstrate that you've carefully reviewed the suggestions.
      - You may find the suggestions in this [PR Guide](https://programming.codeyourfuture.io/guides/reviewing/trainee-pr-guide/#addressing-reviewer-feedback) useful.
    - Your response may trigger a notification (depending on the reviewer's settings), helping ensure they’re aware of the updates you’ve made. 
  - **Replace the "Reviewed" label by a "Needs review" label** (which you have done -- great!)
    - Without this label, the reviewer would not know if your changes is ready to be reviewed.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
Pretty548: Thanks for the feedback! I’ve added a test for the other boundary case (0) and updated the tests.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
Pretty548: Thank you for the feedback. I’ve updated the tests to follow the rule that a proper fraction is when `abs(numerator) < abs(denominator)`. I grouped the proper fraction cases together and included the different positive and negative combinations. Please let me know if there is anything else I should improve.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
```diff
@@ -14,7 +11,37 @@ test(`should return "Acute angle" when (0 < angle < 90)`, () => {
...
+test(`should return "Invalid angle" when angle is 0`, () => {
+  expect(getAngleType(0)).toEqual("Invalid angle");
+});
+test(`should return "Invalid angle" when angle is 360 or greater`, () => {
+  expect(getAngleType(360)).toEqual("Invalid angle");
+});
+test(`should return "Invalid angle" when angle is not a number`, () => {
+  expect(getAngleType("not a number")).toEqual("Invalid angle");
+});
```
cjyuan: What do you expect from this function call:

`getAngleType("100")` or `getAngleType("      +1e2")` ? 

Note: You don't have to include this test because the spec has already stated you can assume the parameters are always valid numbers.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -4,7 +4,29 @@ const isProperFraction = require("../implement/2-is-proper-fraction");
...
   expect(isProperFraction(1, 0)).toEqual(false);
 });
+
+// Case: numerator is zero
+test("should return true when numerator is zero and denominator is positive", () => {
+  expect(isProperFraction(0, 1)).toEqual(true);
+});
+
+// Proper fractions
+test("should return true when abs(numerator) < abs(denominator)", () => {
```
cjyuan, line 18: You could also apply this notation to improper fractions.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
Pretty548: Thanks for the clarification! I did not include tests for non-number inputs since the spec assumes valid numbers.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
Pretty548: Thanks for the suggestion! I updated the improper fraction test description to use the same abs notation and added a negative numerator/denominator case.

cjyuan:
All good now. Well done.

illicitonion:

PR #1133, opened 2026-03-02
Cape Town | 26-ITP-Jan | Pretty Taruvinga | Sprint 3 | Practice TDD

Pretty548:
## Learners, PR Template

Self checklist

## Changelist

In this PR I implemented the required functions using the TDD approach.

Wrote tests first before implementing the functions
Made the tests pass by writing the correct logic
Refactored the code to improve readability and structure
Ensured all tests pass successfully using Jest

## Questions
I would appreciate feedback on:

Whether my TDD workflow is correct
Code structure and readability
Test quality and naming


Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -22,3 +26,16 @@ test("should count multiple occurrences of a character", () => {
...
+  const count = countChar(str, char);
+  expect(count).toEqual(0);
+});
+// Scenario: Case Sensitivity
+test("should be case sensitive", () => {
+  const str = "Hello World";
+  const char = "H";
+  const count = countChar(str, char);
+  expect(count).toEqual(1);
+});
```
cjyuan, line 41: Nice touch to to show match is case sensitive.

Could also consider a case to show that the function is expected to work also for non-alphabets.

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,7 @@
 function getOrdinalNumber(num) {
-  return "1st";
+  const suffixes = ["th", "st", "nd", "rd"];
+  const v = num % 100;
+  return num + (suffixes[(v - 20) % 10] || suffixes[v] || suffixes[0]);
```
cjyuan, line 4: Fancy!

**To check your understanding:**
Suppose `v` is 97, then both  
97 % 10 = 7  
(97-20) % 10 = 77 % 10 =7 

Why not express line 4 as `return num + (suffixes[v % 10] || suffixes[v] || suffixes[0]);`?

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -13,8 +35,18 @@ const getOrdinalNumber = require("./get-ordinal-number");
...
+test("should append 'th' for numbers ending with 11, 12, or 13", () => {
+  expect(getOrdinalNumber(11)).toEqual("11th");
+  expect(getOrdinalNumber(12)).toEqual("12th");
+  expect(getOrdinalNumber(13)).toEqual("13th");
+});
+
+// Case 4: All other numbers
+// When the number does not end with 1, 2, or 3 (or ends with 11, 12, or 13),
+// Then the function should return a string by appending "th" to the number.
+test("should append 'th' for all other numbers", () => {
```

Sprint-3/2-practice-tdd/repeat-str.test.js
cjyuan, line 1: There is only one test in this script.

cjyuan (commented):
Functions look good.

Tests could use some improvement.

Also, can you remove any unused code and unnecessary comments from the files to keep them cleaner? They make the code less readable.

Sprint-3/2-practice-tdd/count.test.js
Pretty548: Thanks for the suggestion! I'll add a test case to ensure the function also handles non-alphabet characters.

Sprint-3/2-practice-tdd/get-ordinal-number.js
Pretty548: Using v % 10 would incorrectly produce values like 11st, 12nd, and 13rd. The (v - 20) % 10 expression ensures that numbers greater than 20 follow the correct last-digit rule while still allowing the fallback suffixes[v] to correctly handle 11–13.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
Pretty548: Thanks for the suggestion! I’ve revised the test description to explicitly state that the case covers numbers ending with 0 or 4–9, which makes the purpose of the test clearer when reading failure messages.

Sprint-3/2-practice-tdd/get-ordinal-number.js
cjyuan: Can you elaborate how the "fallback" works in `suffixes[(v-20) % 10] || suffixes[v] || suffixes[0]` when `v` is 11?

Sprint-3/2-practice-tdd/repeat-str.test.js
```diff
@@ -1,32 +1,44 @@
...
-// When the repeatStr function is called with these inputs,
-// Then it should throw an error, as negative counts are not valid.
+// Case: non-integer count
+test("should throw an error when count is not an integer", () => {
+  const str = "hello";
+  const count = 2.5;
+  expect(() => repeatStr(count, str)).toThrow(
+    "Count must be a non-negative integer"
+  );
+});
```
cjyuan, line 44: Can your function pass these tests?

Project-CLI-Treasure-Hunt
cjyuan, line 1: Can you remove this file, as it is not related to this exercise?

You can share your result on Slack (either in the community channel or your cohort channel).

Sprint-3/2-practice-tdd/get-ordinal-number.js
Pretty548: When v is 11, (v - 20) % 10 does not match a valid suffix index, so it returns undefined. Because of the ||, JavaScript moves to the next option. suffixes[v] is also undefined, so it finally uses suffixes[0], which is "th". That’s why 11 becomes "11th".

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -35,7 +30,7 @@ test("should append 'rd' for numbers ending with 3, except those ending with 13"
 // Case 1: Numbers ending with 1 (but not 11)
 // When the number ends with 1, except those ending with 11,
 // Then the function should return a string by appending "st" to the number.
-test("should append 'th' for numbers ending with 11, 12, or 13", () => {
+test("should append 'th' for special cases 11, 12, or 13", () => {
```
cjyuan: I think the previous description is better.

cjyuan (commented):
Without the "Needs Review" label, I am not sure if this PR ready to be re-reviewed.

Please note that now your branch has three files not belonging to the TDD exercise.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
Pretty548: Yes definitely, i changed it.

Sprint-3/2-practice-tdd/repeat-str.test.js
Pretty548: Yes

cjyuan:
You deleted files in other subfolders under Sprint-3 in your latest commit. Deleted files are considered "changed files", and that's why the validation bot complained.

You can copy the original files from CYF repo to replace the files you deleted, or you can use "git revert" to undo the commit which you deleted all those files. After you restore those files, make a commit.

Or you can create a new PR, include a comment that it is a continuation from this PR, and tag me in the new PR.

cjyuan:
Changes look good.

And well done in fixing the branch.

illicitonion:

PR #1140, opened 2026-03-03
Glasgow | JAN ITP-2026|Tuan Nguyen | Sprint 1 | Programming-fundamentals 

Jacknguyen4438:
Self checklist

## Changelist

I have finish  all 3 section from sprint 1:
  1) Key-exercises.
  2) Mandatory-errors.
  3) Mandatory-interpret
If there is some exercise or course from sprint 1 I need to fix and improve please let me know.

## Questions

Yes I see that in sprint 3 I need to make multiple PR request. Could I do that in the same branch for sprint 3 or I need to make like sub branch from sprint 3 branch in order to submit?

### DIFF
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -1,6 +1,10 @@
 let count = 0;
 
 count = count + 1;
+/* in line three, the variable count is combining the variable count original value of 0 with new value of 1 with the used of plus operator. 
+The equal operator is this case helping to re-assign the new value back to the count variable
+*/
+console.log(count);
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -1,8 +1,7 @@
 let count = 0;
 
 count = count + 1;
-/* in line three, the variable count is combining the variable count original value of 0 with new value of 1 with the used of plus operator. 
-The equal operator is this case helping to re-assign the new value back to the count variable
+/* in line three, the variable count is reassign new value by increment it current value by 1.
 */
 console.log(count);
 
--- a/Sprint-1/1-key-exercises/2-initials.js
+++ b/Sprint-1/1-key-exercises/2-initials.js
@@ -5,7 +5,8 @@ let lastName = "Johnson";
 // Declare a variable called initials that stores the first character of each string.
 // This should produce the string "CKJ", but you must not write the characters C, K, or J in the code of your solution.
 
-let initials = ``;
+let initials = firstName.slice(0,1)+middleName.slice(0,1)+lastName.slice(0,1);
+console.log(initials);
 
 // https://www.google.com/search?q=get+first+character+of+string+mdn
 
--- a/Sprint-1/1-key-exercises/2-initials.js
+++ b/Sprint-1/1-key-exercises/2-initials.js
@@ -5,7 +5,7 @@ let lastName = "Johnson";
 // Declare a variable called initials that stores the first character of each string.
 // This should produce the string "CKJ", but you must not write the characters C, K, or J in the code of your solution.
 
-let initials = firstName.slice(0,1)+middleName.slice(0,1)+lastName.slice(0,1);
+let initials = firstName.charAt(0)+middleName.charAt(0)+lastName.charAt(0);
 console.log(initials);
 
 // https://www.google.com/search?q=get+first+character+of+string+mdn
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -17,11 +17,9 @@ console.log(`The base part of ${filePath} is ${base}`);
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
 
- 
-let dirDirectory = filePath.slice(filePath.indexOf("/"),filePath.lastIndexOf("."));
+let dirDirectory = filePath.slice(filePath.indexOf("Users"),filePath.lastIndexOf("/"));
 let extDirectory = filePath.slice(filePath.lastIndexOf("."));
-let baseDirectory = filePath.slice(filePath.lastIndexOf("/"));
 console.log(`The dir part of the file is ${dirDirectory}.`);
 console.log(`The ext part of the file is ${extDirectory}.`);
-console.log(`The base part of the file is ${baseDirectory}.`);
+
 // https://www.google.com/search?q=slice+mdn
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -23,35 +23,10 @@ const ext =filePath.slice(slashIndex + 49, 53);
 console.log(`The dir part of ${filePath} is ${dir}`);
 console.log(`The ext part of ${filePath} is ${ext}`);
 */
-
-function pathFinder(userInput, filePath){
-  // 1. Check if the user input exists inside the full path
-  if (!fullPath.includes(userInput)) {
-    console.log("Input not found in path");
-    return;
-  }
-
-  // 2. Split the path into parts
-  const parts = filePath.split("/");
-
-  // 3. Base = last element
-  const base = parts[parts.length - 1];
-
-  // 4. Dir = everything except the last element
-  const dir = parts.slice(0, parts.length - 1).join("/");
-
-  // 5. Extension logic
-  const dotIndex = base.lastIndexOf(".");
-  const ext = dotIndex === -1 ? "" : base.slice(dotIndex);
-
-  // 6. Print everything
-  console.log(`Full path: ${filePath}`);
-  console.log(`Directory: ${dir}`);
-  console.log(`Base: ${base}`);
-  console.log(`Extension: ${ext}`);
-
-}
-
-console.log(pathFinder(filePath));
+ 
+let dirDirectory = filePath.slice(filePath.indexOf("/"),filePath.lastIndexOf("/"));
+let extDirectory = filePath.slice(filePath.lastIndexOf("/"));
+console.log(`The dir part of the file is ${dirDirectory}.`);
+console.log(`The ext part of the file is ${extDirectory}.`);
 
 // https://www.google.com/search?q=slice+mdn
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -24,14 +24,32 @@ console.log(`The dir part of ${filePath} is ${dir}`);
 console.log(`The ext part of ${filePath} is ${ext}`);
 */
 
-function pathFinder(path){
-   const lastSlash = path.lastIndexOf("/");
-  const dir = path.slice(0, lastSlash);
-  const base = path.slice(lastSlash + 1);
-  const dot = base.lastIndexOf(".");
-  const ext = dot !== -1 ? base.slice(dot) : "";
-  
-  return { dir, base, ext };
+function pathFinder(userInput, filePath){
+  // 1. Check if the user input exists inside the full path
+  if (!fullPath.includes(userInput)) {
+    console.log("Input not found in path");
+    return;
+  }
+
+  // 2. Split the path into parts
+  const parts = filePath.split("/");
+
+  // 3. Base = last element
+  const base = parts[parts.length - 1];
+
+  // 4. Dir = everything except the last element
+  const dir = parts.slice(0, parts.length - 1).join("/");
+
+  // 5. Extension logic
+  const dotIndex = base.lastIndexOf(".");
+  const ext = dotIndex === -1 ? "" : base.slice(dotIndex);
+
+  // 6. Print everything
+  console.log(`Full path: ${filePath}`);
+  console.log(`Directory: ${dir}`);
+  console.log(`Base: ${base}`);
+  console.log(`Extension: ${ext}`);
+
 }
 
 console.log(pathFinder(filePath));
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -16,17 +16,12 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
-/*
-const slashIndex = filePath.indexOf("/");
-const dir =filePath.slice(slashIndex + 0, 45) ;
-const ext =filePath.slice(slashIndex + 49, 53);
-console.log(`The dir part of ${filePath} is ${dir}`);
-console.log(`The ext part of ${filePath} is ${ext}`);
-*/
+
  
-let dirDirectory = filePath.slice(filePath.indexOf("/"),filePath.lastIndexOf("/"));
-let extDirectory = filePath.slice(filePath.lastIndexOf("/"));
+let dirDirectory = filePath.slice(filePath.indexOf("/"),filePath.lastIndexOf("."));
+let extDirectory = filePath.slice(filePath.lastIndexOf("."));
+let baseDirectory = filePath.slice(filePath.lastIndexOf("/"));
 console.log(`The dir part of the file is ${dirDirectory}.`);
 console.log(`The ext part of the file is ${extDirectory}.`);
-
+console.log(`The base part of the file is ${baseDirectory}.`);
 // https://www.google.com/search?q=slice+mdn
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -16,9 +16,24 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
+/*
 const slashIndex = filePath.indexOf("/");
 const dir =filePath.slice(slashIndex + 0, 45) ;
 const ext =filePath.slice(slashIndex + 49, 53);
 console.log(`The dir part of ${filePath} is ${dir}`);
 console.log(`The ext part of ${filePath} is ${ext}`);
+*/
+
+function pathFinder(path){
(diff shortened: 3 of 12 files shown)


Sprint-1/1-key-exercises/1-count.js
```diff
@@ -1,6 +1,10 @@
 let count = 0;
 
 count = count + 1;
+/* in line three, the variable count is combining the variable count original value of 0 with new value of 1 with the used of plus operator. 
+The equal operator is this case helping to re-assign the new value back to the count variable
+*/
+console.log(count);
```

Sprint-1/1-key-exercises/2-initials.js
```diff
@@ -5,7 +5,8 @@ let lastName = "Johnson";
 // Declare a variable called initials that stores the first character of each string.
 // This should produce the string "CKJ", but you must not write the characters C, K, or J in the code of your solution.
 
-let initials = ``;
+let initials = firstName.slice(0,1)+middleName.slice(0,1)+lastName.slice(0,1);
```
cjyuan: There are easier and more efficient ways to to extract a character at specific position/index in a string than using `.slice()`.

Can you find out what they are and update your code accordingly?

Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -16,8 +16,9 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
-
-const dir = ;
-const ext = ;
-
+const slashIndex = filePath.indexOf("/");
+const dir =filePath.slice(slashIndex + 0, 45) ;
+const ext =filePath.slice(slashIndex + 49, 53);
```
cjyuan: Can you make these two statements to work for any valid path? For example, they could still work correctly if we changed line 12 to
```
const filePath = "/Users/jacknguyen448/cyf/Module-2/Sprint-1/package.json";
```

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -3,7 +3,16 @@ const maximum = 100;
...
 // Try breaking down the expression and using documentation to explain what it means
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+/*Number data type represents any type of number such as integer, float number(decimal number), infinity number, etc...
+In the code at line 4 the num variable are doing math function that generate a integer between variable minimum and maximum value,
+the floor method behind the Math are is main purpose if to help turn value of a float number and round it up to the nearest integer.
+inside the Math.floor parameter the math random method is choosing a random float number between 0 and 1 and have it multiply with the second expression
+to set the range of possible value range from 0 to under 100. The last plus one to make sure the that number stay above 0(0+1) and to 100(99+1);
+*/
```
cjyuan, line 27: Phrases like "a number between X and Y" are not precise enough in a program specification, because they do not clearly state whether the endpoints X and Y are included.

We can also use the concise and precise ***interval notation*** to describe a range of values. 
- `[`, `]` => inclusion
- `(`, `)` => exclusion

---
Can you practice this and use it to describe the range of numbers that could be produced in each of these sub-expressions?

1. `Math.random()`

Sprint-1/2-mandatory-errors/1.js
```diff
@@ -1,4 +1,8 @@
 // trying to create an age variable and then reassign the value by 1
 
-const age = 33;
-age = age + 1;
+let age = 33;
+age = ++age;
```
cjyuan: Part of line 4 is redundant (and is a dangerous practice). Can you refine it?

Sprint-1/2-mandatory-errors/3.js
```diff
@@ -1,5 +1,7 @@
 const cardNumber = 4533787178994213;
-const last4Digits = cardNumber.slice(-4);
+const last4Digits = Number(cardNumber.toString().slice(-4));
+console.log(last4Digits);
+console.log(typeof last4Digits);
```
cjyuan: If `cardNumber` is `4533787178990001`, does your code produce the output you expect?

Sprint-1/2-mandatory-errors/4.js
```diff
@@ -1,2 +1,6 @@
-const 12HourClockTime = "20:53";
-const 24hourClockTime = "08:53";
\ No newline at end of file
+const $12HourClockTime = "20:53";
+const $24HourClockTime = "08:53";
```
cjyuan: Starting a variable name with a `$` is generally not recommended.

Normal practice is to begin a variable name with a lowercase letter.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -20,3 +20,30 @@ console.log(`The percentage change is ${percentageChange}`);
...
+they are call:
+  - Number(carPrice.replaceAll(",","")); in line  4 is a function calls because the .replaceALL is being executed
+  - Number(priceAfterOneYear.replaceAll(",", "")); in line 5 is also a function call because .replaceAll is being executed.
+  - Inside the Number() function the .replaceAll in both line 4 and 5 is also a function call because the the parent function is also
+   being executed.
+  - the console.log is a function calls
+So the total amount of function call in this code is 5
+
+b) The main error for the when the running this line of code if for the missing right parameter after the double quote, so fix this 
+just add the coma right after the double quote.
```
cjyuan: The grammar of this sentence is not quite correct. Can you revise it?

---
We can more precisely describe "A comma is missing between `","` and `""` in the function call" as:  

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -23,3 +23,31 @@ console.log(result);
...
+
+c)At the line 3 we can see the the expression movieLength % 60, 
+what is represent is the reminder value after the division movieLength(in second) value by 60%; 
+this is to remind how much left over second left in the movie.
+
+d)In line 4, the the expression that is assigned to totalMinutes 
+represent the value after the math operation if movieLength value and remainingSeconds then divine by 60;
+
+e)For the variable result it represent the combine total time length of the movie in hour/minutes/second template, for better name for this variable
+I would to rename it to totalMovieLength.
```
cjyuan: The name `totalMovieLength` does not quite indicate the value stored in the variable 
is a formatted string in the form `"2:12:02"`. (Numbers such as 1.5 (hours) and 90 (minutes) are also valid total movie length.)

Can you suggest a more descriptive name?

Sprint-1/3-mandatory-interpret/3-to-pounds.js
cjyuan, line 16: Optional challenge question (no change required):

In other words, do we really need `.padEnd(2, "0")` in this script? Why?

cjyuan (commented):
I noticed some minor inconsistency formatting in the code/comments.

Have you installed **prettier** VSCode extension and enabled formatting on save/paste on VSCode as recommended in 

Jacknguyen4438:
@cjyuan Hello Yuan, thank you for the review you give me I am very grateful for the insight. I can see that there are many section I need to fix, so I would like to notify you that I have read  the PR review and in the progress of fixing it. I understand that you are very busy with other PR as well and I was radio silent do to I am doing sprint 3 of module 2 at the moment. When I finish fixing and ready to be review I will mention you again thank you.

Sprint-1/1-key-exercises/3-paths.js
ykamal: Good spot. 
@Jacknguyen4438 one of the goals of writing code should be that it works with any valid input provided. Here you are limiting your method to only produce the correct result if the input matches a certain criterion. How can you make it dynamic so that entering ANY file path will produce the correct output?

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -2,7 +2,7 @@ let carPrice = "10,000";
 let priceAfterOneYear = "8,543";
 
 carPrice = Number(carPrice.replaceAll(",", ""));
-priceAfterOneYear = Number(priceAfterOneYear.replaceAll("," ""));
+priceAfterOneYear = Number(priceAfterOneYear.replaceAll(",", ""));
```
ykamal, line 5: This is fine, but the link below is worth looking into. it's an abstraction over this and will probably serve you better in the long run:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

Sprint-1/1-key-exercises/1-count.js
Jacknguyen4438: @cjyuan Yes I understand  your concern on this comment . At the time I write this comment base on what I understand, I intentionally write like this because it  help explain in more detail on how the code work. here is my explain with more correct term:

/* in line three, the variable count is reassign new value by increment it current value by 1.
*/

Sprint-1/1-key-exercises/2-initials.js
Jacknguyen4438: Yes I see that you would  like me use other method that is easier. So in this case I will used charAt[0] to help with this task

Sprint-1/1-key-exercises/3-paths.js
Jacknguyen4438: yes I understand that this is not a correct way to write since is force the user have to input the correct value or it might not work. So I will try to write a function to help with this task.

Sprint-1/2-mandatory-errors/1.js
Jacknguyen4438: Yes I understand my mistake on this one. I was try to update and increase the value by 1 so this time I will write with age += 1;

let age = 33;
age += 1 ;

console.log(age);

Sprint-1/2-mandatory-errors/4.js
Jacknguyen4438: Thank you I will  rename the variable in correct way.

Sprint-1/1-key-exercises/1-count.js
cjyuan: Your comment is good. I just wanted to ensure you know the proper term.

Sprint-1/1-key-exercises/3-paths.js
cjyuan: No function needed. Some expressions (like those on lines 13, 14) that can evaluate to the correct index value will do.

Jacknguyen4438: @cjyuan Hello thank you for the advice I have try and now understand why function is not needed here is my solution.

let dirDirectory = filePath.slice(filePath.indexOf("/"),filePath.lastIndexOf("/"));
let extDirectory = filePath.slice(filePath.lastIndexOf("/"));
console.log(`The dir part of the file is ${dirDirectory}.`);
console.log(`The ext part of the file is ${extDirectory}.`);

Sprint-1/2-mandatory-errors/3.js
Jacknguyen4438: Thank you for the hint,. To my understand with the help of Copilot, the reason I get 1 instead of 0001 because of JavaScript numbers use IEEE‑754 double precision. Which mean that the carNumber value if turn to number it will pass the safe limit. So in this case I will turn the cardNumer value into string then slice the last 4 digit leave as is it in string form.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
Jacknguyen4438: Thank you for the advice in this case, I believed that the right programming term is argument.

Jacknguyen4438: Thank you ,I believe the correct  term is arguments.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
Jacknguyen4438: @ykamal Thank you for the link. I have read through and see that it indeed more organise then the current method I write in the code. I will try to used in my course work next time.

Sprint-1/3-mandatory-interpret/2-time-format.js
Jacknguyen4438: Thank you I see that the variable name are too vague and don't reflect the value format. So I would like to change to formattedMovieDuration.

Jacknguyen4438:
Hello @cjyuan I have make change base on the review suggestion from both you and ykamal. I am ready for another PR review and please if I need to make further change please let me know

Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -16,8 +16,17 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
-
-const dir = ;
-const ext = ;
+/*
+const slashIndex = filePath.indexOf("/");
+const dir =filePath.slice(slashIndex + 0, 45) ;
+const ext =filePath.slice(slashIndex + 49, 53);
```
cjyuan: Why keep this code? Best practice is to remove unused code to keep our code clean. And doing so also helps make code review easier.

Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -16,8 +16,17 @@ console.log(`The base part of ${filePath} is ${base}`);
...
+console.log(`The dir part of ${filePath} is ${dir}`);
+console.log(`The ext part of ${filePath} is ${ext}`);
+*/
+ 
+let dirDirectory = filePath.slice(filePath.indexOf("/"),filePath.lastIndexOf("/"));
+let extDirectory = filePath.slice(filePath.lastIndexOf("/"));
+console.log(`The dir part of the file is ${dirDirectory}.`);
+console.log(`The ext part of the file is ${extDirectory}.`);
```
cjyuan: Suppose `filePath` is `"/tmp/file.txt"`. Based on your understanding of the anatomy of a file path on lines 3-8, can you tell which part of the file path is considered the `dir` part and which part of it is considered the `ext` part?

When you run your code, does your code output the values you expected?

Sprint-1/2-mandatory-errors/4.js
```diff
@@ -1,2 +1,6 @@
-const 12HourClockTime = "20:53";
-const 24hourClockTime = "08:53";
\ No newline at end of file
+const twelveHourClockTime = "20:53";
+const twentyFourHourClockTime = "08:53";
+console.log($12HourClockTime);
+console.log($24HourClockTime);
+
```
cjyuan, line 5: You changed only part of the code. As such, the code won't run.

Sprint-1/1-key-exercises/3-paths.js
Jacknguyen4438: Thank you for point this out, I forgot to remove it when I fix the code. When you check again this line of code will be remove

Sprint-1/1-key-exercises/4-random.js
Jacknguyen4438: Thanks you for the feedback! Here is my explanation about  the ranges using interval notation:

Math.random() → produces a float in [0, 1)

Math.random() * (maximum - minimum + 1) → produces a float in [0, maximum - minimum + 1)

Math.floor(Math.random() * (maximum - minimum + 1)) → produces an integer in [0, maximum - minimum]

Math.floor(Math.random() * (maximum - minimum + 1)) + minimum → produces an integer in [minimum, maximum]

This notation makes it clear which endpoints are included or excluded.

Sprint-1/1-key-exercises/3-paths.js
Jacknguyen4438: Thank you for the feed back I have check again and see some part that I write incorrect. My answer will be include when  I finish and commit sync to github.

Sprint-1/2-mandatory-errors/4.js
Jacknguyen4438: Thank you I have make some change you should be able to check when I make commit to github

Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -16,17 +16,12 @@ console.log(`The base part of ${filePath} is ${base}`);
...
-console.log(`The dir part of ${filePath} is ${dir}`);
-console.log(`The ext part of ${filePath} is ${ext}`);
-*/
+
  
-let dirDirectory = filePath.slice(filePath.indexOf("/"),filePath.lastIndexOf("/"));
-let extDirectory = filePath.slice(filePath.lastIndexOf("/"));
+let dirDirectory = filePath.slice(filePath.indexOf("/"),filePath.lastIndexOf("."));
+let extDirectory = filePath.slice(filePath.lastIndexOf("."));
+let baseDirectory = filePath.slice(filePath.lastIndexOf("/"));
```
cjyuan: The value of the "dir" part is still not correct.

Note: 
- "dir" already stands for "directory"
- "ext" stands for extension (or file extension); it is not a directory
- "base" is also not necessary a directory
- The value of "base' has already been extracted on line 14. The value you extracted on line 23 is not quite correct (because your code includes the '/' character in the base).

Jacknguyen4438: Thank you for the feed back out I have fix the dir part so when you check again it shoul prin out correct part of dir

Jacknguyen4438:
@cjyuan thank you for your feed back I have fix the main issue for PR and it ready to be review.

Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -17,7 +17,9 @@ console.log(`The base part of ${filePath} is ${base}`);
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
 
-const dir = ;
-const ext = ;
+let dirDirectory = filePath.slice(filePath.indexOf("Users"),filePath.lastIndexOf("/"));
```
cjyuan: The top-level folder may not necessarily be "Users".

Can you make this statement to work for any valid path? For examples,  
`/tmp/package.json`  
`/mnt/c/Documents/3-paths.js`

Jacknguyen4438: @cjyuan  Thank you for the feed back. After checking your review comment, I see the problem with my solution. to here is my solution base on your review  for dir part::
const dirDirectory = filePath.slice(1,filePath.lastIndexOf("/"));

cjyuan:
All good now. Well done.

illicitonion:

PR #1143, opened 2026-03-03
London | 26-ITP-January | Khaliun Baatarkhuu | Sprint 2 | Coursework. 

khaliun-dev:
Self checklist

## Changelist

- Completed all mandatory debugging exercises
- Fixed logical errors in existing functions 
- Tested all exercises to confirm expected behavior

## Questions

None at the moment.

### DIFF
--- a/Sprint-2/1-key-errors/0.js
+++ b/Sprint-2/1-key-errors/0.js
@@ -1,13 +1,22 @@
 // Predict and explain first...
-//  =============> write your prediction here
+//  =============> write your prediction here: We'll get an error message because "str" paramater has been declared already,
+//using it again inside the function will clash due to redaclaration. 
 
 // call the function capitalise with a string input
 // interpret the error message and figure out why an error is occurring
 
+//function capitalise(str) {
+//  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
+//  return str;
+//}
+
+// =============> write your explanation here: Identifier 'str' has already been declared, 
+//redaclaration in the same scope throws up the error message 
+// =============> write your new code here: 
 function capitalise(str) {
-  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
-  return str;
+  let newStr = `${str[0].toUpperCase()}${str.slice(1)}`;
+  return newStr;
 }
 
-// =============> write your explanation here
-// =============> write your new code here
+//Above code works because we have used another variable "newStr", avoiding redeclaration.
+
--- a/Sprint-2/1-key-errors/1.js
+++ b/Sprint-2/1-key-errors/1.js
@@ -1,20 +1,47 @@
 // Predict and explain first...
 
 // Why will an error occur when this program runs?
-// =============> write your prediction here
+// =============> write your prediction here: potential error, decimalNumber has already been declared, can't redaclare it again with a new variable in the same scope.
+//another error could be, console.log will try to access a local variable from the global scope. 
 
 // Try playing computer with the example to work out what is going on
 
-function convertToPercentage(decimalNumber) {
-  const decimalNumber = 0.5;
-  const percentage = `${decimalNumber * 100}%`;
+// step one: function convertToPercentage(decimalNumber)
+// function created with a parameter named (decimalNumber)
+// step two: const decimalNumber = 0.5;
+// error occurs because decimalNumber has been declared already in the fucntion, cannot declare another one with the same name , in the same scope. 
+// program stops.
+// we have fixed the first error. Code runs again.
+// step three:  const percentage = `${decimalNumber * 100}%`;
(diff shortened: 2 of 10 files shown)


Sprint-2/1-key-errors/0.js
```diff
@@ -1,13 +1,22 @@
...
 function capitalise(str) {
-  let str = `${str[0].toUpperCase()}${str.slice(1)}`;
-  return str;
+  let newStr = `${str[0].toUpperCase()}${str.slice(1)}`;
+  return newStr;
 }
 
-// =============> write your explanation here
-// =============> write your new code here
+//Above code works because we have used another variable "newStr", avoiding redeclaration.
```
ykamal, line 21: This is good. But do you think there could be another way to do this? One that doesn't require you to declare any variables at all?

Sprint-2/1-key-errors/1.js
```diff
@@ -1,20 +1,47 @@
...
+// =============> write your explanation here: my error predictions turned out to be true and i have played computer with js code to further illustrate my reasoning. 
+
+// Finally, correct the code to fix the problem
+// =============> write your new code here:
+
 function convertToPercentage(decimalNumber) {
-  const decimalNumber = 0.5;
   const percentage = `${decimalNumber * 100}%`;
-
   return percentage;
```
ykamal, line 43: This is good. But do you think there could be another way to do this? One that doesn't require you to declare any variables at all?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -14,6 +14,12 @@
 // Then when we call this function with the weight and height
 // It should return their Body Mass Index to 1 decimal place
 
-function calculateBMI(weight, height) {
+ function calculateBMI(weight, height) {
+    const bmi = weight / (height * height);
```
ykamal, line 18: There's a division here. Do you think we need to put in any specific safety guards?

Sprint-2/1-key-errors/0.js
khaliun-dev: function capitalise(str) {
  return str[0].toUpperCase() + str.slice(1);
}

Could this work? Just by removing the unnecessary variable?

Sprint-2/1-key-errors/1.js
khaliun-dev: function convertToPercentage(decimalNumber) {
  return `${decimalNumber * 100}%`;
}

console.log(convertToPercentage(0.5));

Sprint-2/3-mandatory-implement/1-bmi.js
khaliun-dev: We should guard against height being 0 so we could add the following guard: 

if (height <= 0) {
    return "Height must be greater than 0";
  }

khaliun-dev: @ykamal I think i have fixed all three problems you have raised, could you please have another look?

Sprint-2/1-key-errors/0.js
ykamal: Yes. Less memory, simpler. Well done

Sprint-2/1-key-errors/1.js
ykamal: Yes similar to before

ykamal: Make sure to retain the string templating

Sprint-2/3-mandatory-implement/1-bmi.js
ykamal: Less than 0 is still fine. It's 0 that's the problem.

ykamal (changes requested):
After the division change, it should be good

Sprint-2/3-mandatory-implement/1-bmi.js
khaliun-dev: function calculateBMI(weight, height) {
    if (height === 0) {
        throw new Error("Height cannot be 0");
    }

    const bmi = weight / (height * height);
    return Number(bmi.toFixed(1));
}


@ykamal i have changed the guards for 0 specifically

ykamal: Great, thanks. Ah, it's a body weight thing, so less than zero would be a problem 😂😂 but code wise this is fine

ykamal (approved):
LGTM, approved

khaliun-dev:
@ykamal thanks for approving my PR, could you add the completed tag, please?

ykamal:
> @ykamal thanks for approving my PR, could you add the completed tag, please?

Done. Hopefully I've done it right. Nice work

illicitonion:

PR #1147, opened 2026-03-03
Manchester | 26-ITP-Jan | Mehroz Munir | Sprint 3 | Practice TDD Coursework

MehrozMunir:
## Learners, PR Template

Self checklist

## Changelist

Created maximum possible categorical test cases and then corrected the implementation so that all the test cases be passed.


Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,8 @@
 function countChar(stringOfCharacters, findCharacter) {
-  return 5
+  if (findCharacter === "")
```
kyle-tightest: good validation checking. is there perhaps another parameter here that could result in the same output?

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,8 @@
...
+    return 0;
+  else
+    return stringOfCharacters.split(findCharacter).length -1;
```
kyle-tightest: interesting use of `split` to solve this question, it works perfectly.  
how did you think to use `split`?

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,17 @@
 function getOrdinalNumber(num) {
-  return "1st";
+  let numberString = String(num);
+  lastTwoCharactersNumberString = numberString.slice(-2);
+  lastCharacterNumberString = numberString.slice(-1);
+  if(lastTwoCharactersNumberString === "11")
```
kyle-tightest: please check your formatting in this file.  
make sure it follows the style guide and its consistent so that its easily readable and maintainable

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,17 @@
-function repeatStr() {
-  return "hellohellohello";
+function repeatStr(str, count) {
+  try{
```
kyle-tightest: please check your formatting in this file.
make sure it follows the style guide and its consistent so that its easily readable and maintainable

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,17 @@
...
+    if(count === 0)
+      return "";
+    else if(count === 1)
+      return str;
+    else if(count < 0)
+      throw new Error("an error is thrown");
+    else
+    return str.repeat(count);
+  }
+  catch(e){
```
kyle-tightest: why are you wrapping this in a `try catch`?

Sprint-3/2-practice-tdd/count.js
MehrozMunir: Yes I think that I can check the length of the findCharacter here and if that is 0 I would return 0;

`if (findCharacter.length === 0) return 0;`

And I think that is a better validation than checking if string is empty. I have replaced this in my code now.

Thanks

Sprint-3/2-practice-tdd/count.js
MehrozMunir: When I was thinking about the solutions how can i count the number of characters in a given string. The first thing came to my mind was the use of some loop to go through each character of string and compare it with the character i am looking for and add 1 to your count.

But, then I realized that I can solve it more efficiently without using loop with the use of split method as through split I can get an array of strings separated by character i am looking for in the separater parameter. 

So, if I have five occurences of my character in the string, it will return my an array of size 6 that would be strings separated by that character including the empty strings.

Sprint-3/2-practice-tdd/get-ordinal-number.js
MehrozMunir: I am sorry my mistake. I had installed prettier for formatting, but did not enable it in the settings as the default formatter and use it on save. I have done it now. And it looks much better.

Thank you.

Sprint-3/2-practice-tdd/repeat-str.js
MehrozMunir: I was wrapping it because i was catching the thrown error in the catch block and send the error message string back so that I will be able to compare this string in my test file and would know that the error actually is being thrown.

But, I just have learned about the better way how can I test errors thrown in my test file while I was working on a different part of this module. And for that I don't need try catch block here. I have implement this change in the code.

Thanks.

Sprint-3/2-practice-tdd/count.js
kyle-tightest: While that is true I was actually referring to something else.  
Look at the method again and see `is there perhaps another parameter here that could result in the same output?`

Sprint-3/2-practice-tdd/count.js
kyle-tightest: Very good, nice innovative solution

Sprint-3/2-practice-tdd/repeat-str.js
kyle-tightest: Okay awesome.  
Yes there are definitely use cases for "try catch" but its generally not good practice to do "try {} catch (Exception e)" as this treats all exceptions equally and will result in the same behaviour regardless of the type/severity of the exception. it is better to be more explicit with errors and know exactly how you are handling each error. More effort but very worth it when you are trying to debug an issue and all you see is "exception was thrown"

Sprint-3/2-practice-tdd/count.js
MehrozMunir: oh yes. Actually, if `stringOfCharacters.length === 0`, I am expecting the same result. I have added it to my code now.

Sprint-3/2-practice-tdd/repeat-str.js
MehrozMunir: Yes, thank you for the advice. I totally agree with that. I have updated the error message now. 

Thanks.

Sprint-3/2-practice-tdd/count.js
kyle-tightest: @MehrozMunir 100% yes! defensive coding is always a good idea

kyle-tightest (approved):

illicitonion:

PR #1150, opened 2026-03-03
London | 26-ITP-Jan | Carlos Abreu | Sprint 3 | Practice TDD Coursework

carlosyabreu:
## Learners, PR Template

Self checklist

## Changelist
PR for Sprint-3 of Practice TDD Coursework

### DIFF
--- a/Sprint-3/2-practice-tdd/count.js
+++ b/Sprint-3/2-practice-tdd/count.js
@@ -1,5 +1,40 @@
+/**
+ * Original file:
+ *
 function countChar(stringOfCharacters, findCharacter) {
   return 5
 }
 
 module.exports = countChar;
+*
+* End of file
+*/
+
+// Implementation:
+
+/**
+ * Counts the number of times a character occurs in a string
+ * @param {string} str - The string to search 
+ * @param {string} char - The single character to search for
+ * @returns {number} - The count of occurrences
+ */
+function countChar(str, char) {
+  // Input validation
+  if (typeof str !== 'string' || typeof char !== 'string' || char.length !== 1) {
+    return 0;
+  }
+
+  let count = 0;
+
+  // Loop through each character in the string
+  for (let i = 0; i < str.length; i++) {
+    if (str[i] === char) {
+      count++;
+    }
+  }
+
+  return count;
+}
+
+module.exports = countChar;
+
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -1,3 +1,6 @@
+/**
+ * Origignal file:
+ * 
 // implement a function countChar that counts the number of times a character occurs in a string
 const countChar = require("./count");
 // Given a string `str` and a single character `char` to search for,
@@ -22,3 +25,55 @@ test("should count multiple occurrences of a character", () => {
 // And a character `char` that does not exist within `str`.
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of `char` were found.
+*
+* End of file
+*/
+
+// Implementation:
+
+const countChar = require("./count");
+
+describe("countChar function", () => {
+  test("should count multiple occurrences of a character", () => {
+    const str = "aaaaa";
+    const char = "a";
+    const count = countChar(str, char);
+    expect(count).toEqual(5);
+  });
+
+  test("should return 0 when character does not occur in the string", () => {
+    const str = "hello world";
+    const char = "z";
+    const count = countChar(str, char);
+    expect(count).toEqual(0);
+  });
+
+  // Additional test cases to ensure robustness
+  test("should handle empty string", () => {
+    const str = "";
+    const char = "a";
+    const count = countChar(str, char);
+    expect(count).toEqual(0);
+  });
+
+  test("should handle single occurrence", () => {
+    const str = "hello world";
+    const char = "h";
+    const count = countChar(str, char);
+    expect(count).toEqual(1);
+  });
+
+  test("should handle case sensitivity", () => {
+    const str = "Hello World";
+    const char = "h";
+    const count = countChar(str, char);
+    expect(count).toEqual(0); // 'H' is different from 'h'
+  });
+
+  test("should handle invalid inputs gracefully", () => {
+    expect(countChar(123, "a")).toEqual(0); // Non-string first argument
+    expect(countChar("hello", 123)).toEqual(0); // Non-string second argument
+    expect(countChar("hello", "ab")).toEqual(0); // Multiple characters
+  });
+});
+
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.js
@@ -1,5 +1,39 @@
+/**
+ * Original file
+ *
 function getOrdinalNumber(num) {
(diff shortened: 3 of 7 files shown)


Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,40 @@
...
+*/
+
+// Implementation:
+
+/**
+ * Counts the number of times a character occurs in a string
+ * @param {string} str - The string to search 
+ * @param {string} char - The single character to search for
+ * @returns {number} - The count of occurrences
+ */
```
cjyuan, line 20: Nicely commented.

Note: It is ok to delete the original code. Should people need the original code, they can find them from a previous commit. Doing so can help keep the code cleaner.

Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -22,3 +25,55 @@ test("should count multiple occurrences of a character", () => {
...
+
+  test("should return 0 when character does not occur in the string", () => {
+    const str = "hello world";
+    const char = "z";
+    const count = countChar(str, char);
+    expect(count).toEqual(0);
+  });
+
+  // Additional test cases to ensure robustness
+  test("should handle empty string", () => {
```
cjyuan, line 52: The person implementing the function may not immediately understand what "handle" means. It would be clearer to describe the expected behavior directly in the test description, such as specifying what the function should return.

Can you check which test descriptions (in all three test.js files) can be made more informative?

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,39 @@
...
   return "1st";
 }
 
 module.exports = getOrdinalNumber;
+*
+* Enf of file
+*/
+
+function getOrdinalNumber(number) {
+  if (typeof number !== 'number' || !Number.isInteger(number) || number < 0) {
```
cjyuan, line 14: One of the check here is redundant.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -18,3 +21,107 @@ test("should append 'st' for numbers ending with 1, except those ending with 11"
...
+    expect(getOrdinalNumber(111)).toEqual("111th");
+    expect(getOrdinalNumber(112)).toEqual("112th");
+    expect(getOrdinalNumber(113)).toEqual("113th");
+    expect(getOrdinalNumber(1011)).toEqual("1011th");
+    expect(getOrdinalNumber(1012)).toEqual("1012th");
+    expect(getOrdinalNumber(1013)).toEqual("1013th");
+  });
+
+  // Case 5: All other numbers (should get "th")
+  test("should append 'th' for all other numbers", () => {
```
cjyuan, line 74: When a test fails with the message "... all other numbers", it may be unclear what "other numbers" actually refers to.

cjyuan (commented):
Function implementation is solid, especially on validating the parameters.
Test cases are comprehensive! Well done!

I think some of the test descriptions could be made more informative.

Sprint-3/2-practice-tdd/count.js
carlosyabreu: Evening @cjyuan 
I decided to keep the original code it avoids to switch back and forth between implementation and the specification. Totally agree it makes the code not very clean but helps to have what I'm doing on the same place.

cjyuan: You can keep the specification.

The original code is just like a placeholder. It shouldn't have much use while you are implementing yours, and you can always delete it after you have implemented yours.

Sprint-3/2-practice-tdd/count.test.js
carlosyabreu: Evening @cjyuan 
Thanks for the feedback.
I'm not sure if you want me to implement all the code part or just to highlight on the test descriptions how to make it more informative.
Based on my understanding I think that's the second option to make the code more informative so I'll not write all the code implementation here but just the portion of the code that show case it differently using the description below:

1.) countChar.test.js

// Instead of: "should handle empty string"
test("should return 0 when string is empty", () => {
  // ...
});

// Instead of: "should handle single occurrence"
test("should return 1 when character appears once in the string", () => {
  // ...
});

// Instead of: "should handle case sensitivity"
test("should treat lowercase and uppercase characters as different", () => {
  // ...
});

// Instead of: "should handle invalid inputs gracefully"
test("should return 0 when first argument is not a string", () => {
  // ...
});

test("should return 0 when second argument is not a string", () => {
  // ...
});

test("should return 0 when second argument contains multiple characters", () => {
  // ...
});

2.) get-ordinal-number.test.js

This test file I think already has good descriptive names. 
Definitely, the descriptions clearly state what behavior is expected. 
I don't think any changes needed here.

3.) repeat-str.test.js

/ Instead of: "should handle empty string"
test("should return empty string when input string is empty", () => {
  // ...
});

// Instead of: "should handle string with spaces"
test("should preserve spaces when repeating the string", () => {
  // ...
});

// Instead of: "should throw error when first argument is not a string"
test("should throw error if first argument is not a string", () => {
  // ...
});

Sprint-3/2-practice-tdd/get-ordinal-number.js
carlosyabreu: Looking at this code again, the redundant check is the !Number.isInteger(number) condition in the 14th line.

The integer check is necessary for the logic to work correctly (since ordinal numbers only make sense for integers).

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
carlosyabreu: Sure.

// Case 5: All other numbers (should get "th")
  describe("should append 'th' for numbers that don't fit special cases", () => {
    test("numbers ending with 0", () => {
      expect(getOrdinalNumber(0)).toEqual("0th");
      expect(getOrdinalNumber(10)).toEqual("10th");
      expect(getOrdinalNumber(20)).toEqual("20th");
      expect(getOrdinalNumber(100)).toEqual("100th");
      expect(getOrdinalNumber(1000)).toEqual("1000th");
    });

Sprint-3/2-practice-tdd/count.test.js
cjyuan: In TDD, the test description (or test name) should clearly explain what behavior the function should exhibit. 

You proposed descriptions are much clearer.

No changes required as you have demonstrated you understand how to make the test description informative.

Sprint-3/2-practice-tdd/get-ordinal-number.js
cjyuan: `!Number.isInteger(number)` is not the redundant check here.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
cjyuan: This work, and it would create a very comprehensive test. The downside is, you will need a test for each of the digits 0, 4, ..., 9. 

I was thinking in the direction of replacing "all other numbers" by a detailed description of the numbers, like "numbers ending with 0, 4, 5, 6, 7, 8, or 9".

illicitonion:

PR #1161, opened 2026-03-04
Birmingham | ITP Jan-2026 | Arunkumar Akilan | Sprint 3 | Practice TDD

arunkumarakilan:
## Learners, PR Template

Self checklist

## Changelist

Implemented the functions for the Sprint 3 Practice TDD exercises.

Completed:
- countChar
- getOrdinalNumber
- repeatStr

All tests pass successfully.

### DIFF
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.js
@@ -1,5 +1,19 @@
 function getOrdinalNumber(num) {
-  return "1st";
+  if( num%10 === 1 && num%100 !== 11)
+  {
+    return `${num}st`;
+  }
+  else if (num%10 === 2 && num%100 !== 12)
+  {
+    return `${num}nd`;
+  }
+  else if(num%10===3 && num%100 !==13)
+     {
+    return `${num}rd`;
+  }
+  else{
+    return `${num}th`;
+  }
 }
 
 module.exports = getOrdinalNumber;
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.js
@@ -1,17 +1,13 @@
 function getOrdinalNumber(num) {
-  if( num%10 === 1 && num%100 !== 11)
-  {
+  const lastDigit = num % 10; 
+  const lastTwoDigits = num % 100; 
+  if (lastDigit === 1 && lastTwoDigits !== 11) {
     return `${num}st`;
-  }
-  else if (num%10 === 2 && num%100 !== 12)
-  {
+  } else if (lastDigit === 2 && lastTwoDigits !== 12) {
     return `${num}nd`;
-  }
-  else if(num%10===3 && num%100 !==13)
-     {
+  } else if (lastDigit === 3 && lastTwoDigits !== 13) {
     return `${num}rd`;
-  }
-  else{
+  } else {
     return `${num}th`;
   }
 }
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.test.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.test.js
@@ -45,12 +45,15 @@ test("should append 'th' for numbers ending with 13", () => {
   expect(getOrdinalNumber(113)).toEqual("113th");
   expect(getOrdinalNumber(213)).toEqual("213th");
 });
-test("should append 'th' for numbers ending with 4", () => {
+test("should append 'th' for numbers ending with 0, 4–9", () => {
   expect(getOrdinalNumber(4)).toEqual("4th");
-  expect(getOrdinalNumber(14)).toEqual("14th");
-  expect(getOrdinalNumber(134)).toEqual("134th");
+  expect(getOrdinalNumber(10)).toEqual("10th");
+  expect(getOrdinalNumber(25)).toEqual("25th");
+  expect(getOrdinalNumber(36)).toEqual("36th");
+  expect(getOrdinalNumber(48)).toEqual("48th");
+  expect(getOrdinalNumber(27)).toEqual("27th");
+  expect(getOrdinalNumber(9)).toEqual("9th");
 });
-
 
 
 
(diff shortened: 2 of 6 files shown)


Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,19 @@
...
-  return "1st";
+  if( num%10 === 1 && num%100 !== 11)
+  {
+    return `${num}st`;
+  }
+  else if (num%10 === 2 && num%100 !== 12)
+  {
+    return `${num}nd`;
+  }
+  else if(num%10===3 && num%100 !==13)
```
cjyuan: Can you lookup "Advantages of replacing repeated expressions by a variable with a descriptive name"?

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -18,3 +18,43 @@ test("should append 'st' for numbers ending with 1, except those ending with 11"
...
+  expect(getOrdinalNumber(13)).toEqual("13th");
+  expect(getOrdinalNumber(113)).toEqual("113th");
+  expect(getOrdinalNumber(213)).toEqual("213th");
+});
+test("should append 'th' for numbers ending with 4", () => {
+  expect(getOrdinalNumber(4)).toEqual("4th");
+  expect(getOrdinalNumber(14)).toEqual("14th");
+  expect(getOrdinalNumber(134)).toEqual("134th");
+});
+
```
cjyuan: What about numbers ending with the other digits? You could consider generalise this category to include numbers ending with 0, 4, 5, ..., or 9, since the all share the same characteristics.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -18,3 +18,43 @@ test("should append 'st' for numbers ending with 1, except those ending with 11"
...
+test("should append 'rd' for numbers ending with 3, except those ending with 13", () => {
+  expect(getOrdinalNumber(3)).toEqual("3rd");
+  expect(getOrdinalNumber(33)).toEqual("33rd");
+  expect(getOrdinalNumber(133)).toEqual("133rd");
+});
+test("should append 'th' for numbers ending with 13", () => {
+  expect(getOrdinalNumber(13)).toEqual("13th");
+  expect(getOrdinalNumber(113)).toEqual("113th");
+  expect(getOrdinalNumber(213)).toEqual("213th");
+});
```
cjyuan: These are comprehensive. You can also consider grouping numbers ending with 11, 12, 13 into the same cateogry.

Sprint-3/2-practice-tdd/repeat-str.test.js
```diff
@@ -12,8 +12,23 @@ const repeatStr = require("./repeat-str");
 test("should repeat the string count times", () => {
   const str = "hello";
   const count = 3;
-  const repeatedStr = repeatStr(str, count);
-  expect(repeatedStr).toEqual("hellohellohello");
+  expect(repeatStr(str,count)).toEqual("hellohellohello");
+});
+test(" should return original 'str' without repetition", ()=>{
```
cjyuan: Better to state also the "reason" for such behavior. For example, "should return original 'str' without repetition when count is 1".

Sprint-3/2-practice-tdd/repeat-str.test.js
```diff
@@ -12,8 +12,23 @@ const repeatStr = require("./repeat-str");
...
+  const str = "hello";
+  const count = 1;
+  expect(repeatStr(str, count)).toEqual("hello");
+});
+test("should return empty string", ()=>{
+  const str = "hello";
+  const count = 0;
+  expect(repeatStr(str, count)).toEqual("");
+
+});
```
cjyuan: In TDD, the test description (or test name) should clearly explain what behavior the function should exhibit. 

Can you also include the "condition" part into these two descriptions?

Sprint-3/2-practice-tdd/repeat-str.test.js
```diff
@@ -12,8 +12,23 @@ const repeatStr = require("./repeat-str");
...
+test("should throw negative counts are not valid Error", ()=>{
```
cjyuan: What do you think about "should throw an error when count is negative"?

cjyuan (commented):
- Function implementation is correct.

- Code format is a bit inconsistent.

- Tests are quite comprehensive. However, the test descriptions could be made clearer.

Sprint-3/2-practice-tdd/repeat-str.test.js
arunkumarakilan: Yes, that makes sense. I agree that **"should throw an error when count is negative"** is clearer. I will update the test description to use that wording.

Sprint-3/2-practice-tdd/repeat-str.test.js
arunkumarakilan: Thanks for the suggestion! I’ve updated the test descriptions to include the condition, following the format **“should <expected_behavior> when <condition>”**. Please let me know if anything else should be improved.

Sprint-3/2-practice-tdd/get-ordinal-number.js
arunkumarakilan: Thanks for the suggestion! I refactored the code to replace the repeated expressions with descriptive variables (`lastDigit` and `lastTwoDigits`) to improve readability.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
arunkumarakilan: Thanks for the suggestion! I updated the test to generalise the case for numbers that append **'th'**, instead of focusing only on specific digits, since they share the same behaviour.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
arunkumarakilan: Thanks for the suggestion! I grouped the cases for numbers ending with 11, 12, and 13 into a single category and updated the tests accordingly.

cjyuan:
Changes look good. Well done!

illicitonion:

PR #1167, opened 2026-03-04
Birmingham | ITP-Jan | Ahmad Roman Sanaye | Sprint 3 | Practice-tdd

RomanSanaye:
Self checklist

## Changelist

Implemented the following functions:

- count
- getOrdinalNumber
- repeatStr

Each function handles the required cases and passes all Jest test cases, including edge cases (positive, zero, and negative values where applicable).

All tests are passing successfully.

### DIFF
--- a/Sprint-3/2-practice-tdd/count.js
+++ b/Sprint-3/2-practice-tdd/count.js
@@ -8,9 +8,6 @@ function countChar(stringOfCharacters, findCharacter) {
     if (wantedChar === findCharacter) {
       numberOfChar++;
     }
-    if (numberOfChar === 0) {
-      return 0;
-    }
   }
   return numberOfChar;
 }
--- a/Sprint-3/2-practice-tdd/count.js
+++ b/Sprint-3/2-practice-tdd/count.js
@@ -1,5 +1,18 @@
+// our function;
 function countChar(stringOfCharacters, findCharacter) {
-  return 5
+  if (!stringOfCharacters || !findCharacter) {
+    return 0;
+  }
+  let numberOfChar = 0;
+  for (let wantedChar of stringOfCharacters) {
+    if (wantedChar === findCharacter) {
+      numberOfChar++;
+    }
+    if (numberOfChar === 0) {
+     return `0 ${findCharacter} found.`
+    }
+  }
+  return numberOfChar;
 }
 
 module.exports = countChar;
--- a/Sprint-3/2-practice-tdd/count.js
+++ b/Sprint-3/2-practice-tdd/count.js
@@ -9,7 +9,7 @@ function countChar(stringOfCharacters, findCharacter) {
       numberOfChar++;
     }
     if (numberOfChar === 0) {
-     return `0 ${findCharacter} found.`
+      return 0;
     }
   }
   return numberOfChar;
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.test.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.test.js
@@ -18,3 +18,27 @@ test("should append 'st' for numbers ending with 1, except those ending with 11"
   expect(getOrdinalNumber(21)).toEqual("21st");
   expect(getOrdinalNumber(131)).toEqual("131st");
 });
+test("should append 'nd' for numbers ending with 2, except those ending with 12", () => {
+  expect(getOrdinalNumber(2)).toEqual("2nd");
+  expect(getOrdinalNumber(22)).toEqual("22nd");
+  expect(getOrdinalNumber(222)).toEqual("222nd");
+});
+test("should append 'rd' for numbers ending with 3, except those ending with 13", () => {
+  expect(getOrdinalNumber(3)).toEqual("3rd");
+  expect(getOrdinalNumber(33)).toEqual("33rd");
+  expect(getOrdinalNumber(333)).toEqual("333rd");
+});
+test("should append 'th' for numbers ending with 4 and more than 4", () => {
+  expect(getOrdinalNumber(4)).toEqual("4th");
+  expect(getOrdinalNumber(44)).toEqual("44th");
+  expect(getOrdinalNumber(444)).toEqual("444th");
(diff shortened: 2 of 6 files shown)


Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,18 @@
...
+    return 0;
+  }
+  let numberOfChar = 0;
+  for (let wantedChar of stringOfCharacters) {
+    if (wantedChar === findCharacter) {
+      numberOfChar++;
+    }
+    if (numberOfChar === 0) {
+     return `0 ${findCharacter} found.`
+    }
```
cjyuan: The function is expected to return an integer number. Returning a string would make the function difficult to use because the caller would need to check the data type of the return value.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -18,3 +18,27 @@ test("should append 'st' for numbers ending with 1, except those ending with 11"
...
+  expect(getOrdinalNumber(2)).toEqual("2nd");
+  expect(getOrdinalNumber(22)).toEqual("22nd");
+  expect(getOrdinalNumber(222)).toEqual("222nd");
+});
+test("should append 'rd' for numbers ending with 3, except those ending with 13", () => {
+  expect(getOrdinalNumber(3)).toEqual("3rd");
+  expect(getOrdinalNumber(33)).toEqual("33rd");
+  expect(getOrdinalNumber(333)).toEqual("333rd");
+});
+test("should append 'th' for numbers ending with 4 and more than 4", () => {
```
cjyuan: `... ending with 4-9, 0` might be more concise.

Sprint-3/2-practice-tdd/repeat-str.test.js
```diff
@@ -10,23 +10,27 @@ const repeatStr = require("./repeat-str");
...
-  expect(repeatedStr).toEqual("hellohellohello");
+  expect(repeatStr("Love", 3)).toEqual("LoveLoveLove");
 });
 
 // Case: handle count of 1:
 // Given a target string `str` and a `count` equal to 1,
 // When the repeatStr function is called with these inputs,
 // Then it should return the original `str` without repetition.
-
+test("should print the original string", () => {
```
cjyuan: "print" may suggest the function should output the string.

Sprint-3/2-practice-tdd/repeat-str.test.js
```diff
@@ -10,23 +10,27 @@ const repeatStr = require("./repeat-str");
...
+  expect(repeatStr("Love", 1)).toEqual("Love");
+});
 // Case: Handle count of 0:
 // Given a target string `str` and a `count` equal to 0,
 // When the repeatStr function is called with these inputs,
 // Then it should return an empty string.
-
+test("should return an empty string", () => {
```
cjyuan: You could probably make this description more informative by including the `when ...` part.

cjyuan (commented):
I noticed some minor format inconsistency.

Have you installed the **prettier** VSCode extension and enabled "Format on save/paste" on VSCode, as recommended in 
?

If you have enabled "Format on save" but it is not working, it is likely that you haven't assign a formatter for JS file. This could happen if you have zero or multiple extensions that can format .js file.

1. Use "Format document" to format the JS file. Sometimes, VSCode will ask you to choose a formatter, and you can manually select "Prettier".
2. Edit `settings.json` and set Prettier as the default formatter for JS.

cjyuan:
There is no new commits on this branch, and the comments are not yet addressed.

If you have responded to the comments, you may want to check if the responses are in "Pending" state. Other users cannot see the pending comments.

Sprint-3/2-practice-tdd/count.js
RomanSanaye: Hello @cjyuan ,
I have updated the function so it now always returns an integer. Returning 0 instead of a string ensures the function is easier to use and matches the expected return type.

Sprint-3/2-practice-tdd/repeat-str.test.js
RomanSanaye: I have updated the test description to say “return” instead of “print” to accurately reflect that the function returns a string rather than printing it.

Sprint-3/2-practice-tdd/repeat-str.test.js
RomanSanaye: I have revised the test descriptions to include the input condition (when …) so they clearly state under what circumstances the function should return the expected result.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
RomanSanaye: I have updated the ordinal number test descriptions to group numbers ending in 4-9 and 0, making the tests more concise while still covering all cases.

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,18 @@
...
+  for (let wantedChar of stringOfCharacters) {
+    if (wantedChar === findCharacter) {
+      numberOfChar++;
+    }
+    if (numberOfChar === 0) {
+      return 0;
+    }
+  }
+  return numberOfChar;
 }
```
cjyuan, line 13: This function has a bug; it does not always return the correct count.

If you cannot figure out the bug, try testing a few more samples in `count.test.js`.

RomanSanaye:
Hi @cjyuan , thanks for your feedback.

I have addressed all the comments:

* Updated the function to always return a number instead of a string
* Fixed the bug in `countChar` by removing the early return inside the loop so it now correctly counts all characters
* Improved test descriptions to use “return” instead of “print”
* Made test descriptions more informative by including the “when…” part
* Simplified ordinal number tests by grouping cases (e.g. numbers ending in 4–9 and 0)

I have also added the **“Needs Review”** label.

Please let me know if anything else is required.

cjyuan:
Changes look good, and thanks for the clear explanation in the responses. Well done.

illicitonion:

PR #1172, opened 2026-03-04
London | 26-ITP-Jan | Carlos Abreu | Sprint 3 | Refactoring Dead Code 

carlosyabreu:
## Learners, PR Template

Self checklist

## Changelist
PR for Sprint-3 of Refactoring dead code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 node_modules
 .DS_Store
 .vscode
-*package.json
-*package-lock.json
 **/.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 node_modules
 .DS_Store
 .vscode
+*package.json
+*package-lock.json
 **/.DS_Store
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,4 +1,6 @@
 node_modules
 .DS_Store
 .vscode
-**/.DS_Store
\ No newline at end of file
+*package.json
```
SlideGauge: package.json and package-lock.json should not be in gitignore
Also, will *package.json ignore only one type of filename?

SlideGauge (commented):
Dead code removed correctly, my main concern is the way how code  is organized: we have old code in comments and new code. Could you just remove the  dead code, as the task asks?

.gitignore
carlosyabreu: Evening @SlideGauge 
Thank for the feedback.
I remove the old code on the implementation and made it more cleaner.
Thanks for info not to include package.json and package-lock.json file in .gitignore. 
I didn't know that.
About the last question, *package.json when put in .gitignore fiel will ignore any files that end with package.json. In other words all files (*) with suffixes package.json

SlideGauge:
Nice job, complete

illicitonion:

PR #1175, opened 2026-03-04
Glasgow | 26-ITP-JAN | Prakash Dcosta | Sprint 3 | Practice TDD

dcostaprakash:
Self checklist

Completed all the tasks as per the coursework

### DIFF
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -22,3 +22,9 @@ test("should count multiple occurrences of a character", () => {
 // And a character `char` that does not exist within `str`.
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of `char` were found.
+test("should return 0 when the character is not found", () => {
+  const str = "hello";
+  const char = "z";
+  const count = countChar(str, char);
+  expect(count).toEqual(0);
+});
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -28,3 +28,15 @@ test("should return 0 when the character is not found", () => {
   const count = countChar(str, char);
   expect(count).toEqual(0);
 });
+test("should return 0 when the character is case sensitive", () => {
+  const str = "Hello";
+  const char = "h";
+  const count = countChar(str, char);
+  expect(count).toEqual(0);
+});
+test("should count multiple occurrences of a symbols", () => {
+  const str = "Brilliant!!!";
+  const char = "!";
+  const count = countChar(str, char);
+  expect(count).toEqual(3);
+});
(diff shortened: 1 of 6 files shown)


Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -18,3 +18,34 @@ test("should append 'st' for numbers ending with 1, except those ending with 11"
   expect(getOrdinalNumber(21)).toEqual("21st");
   expect(getOrdinalNumber(131)).toEqual("131st");
 });
+
+// Case 2: Numbers ending with 2 (but not 12)
+// When the number ends with 2, except for those ending with 12,
+// Then the function should return a string by appending "nd" to the number.
+ test("returns 'nd' for 2, 22, 102", () => {
```
cjyuan: In TDD, the test description (or test name) should clearly explain what behavior the function should exhibit. 

If the description only lists three numbers, then it could be misinterpreted as, "the function should only append 'nd' to these three numbers.

To make the description clearer, can you describe the characteristics of the numbers in this and the next categories?

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -18,3 +18,34 @@ test("should append 'st' for numbers ending with 1, except those ending with 11"
...
+  test("returns 'rd' for 3, 23, 103", () => {
+    expect(getOrdinalNumber(3)).toBe("3rd");
+    expect(getOrdinalNumber(23)).toBe("23rd");
+    expect(getOrdinalNumber(103)).toBe("103rd");
+  });
+
+// Case 4: All other numbers including umbers ending with 11, 12 or 13
+// These are for all other numbers including special cases and should always end with "th",
+// Then the function should return a string by appending "th" to the number.
+test("returns 'th' for all other numbers including those ending with 11, 12, 13", () => {
```

Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -22,3 +22,9 @@ test("should count multiple occurrences of a character", () => {
 // And a character `char` that does not exist within `str`.
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of `char` were found.
+test("should return 0 when the character is not found", () => {
+  const str = "hello";
+  const char = "z";
+  const count = countChar(str, char);
+  expect(count).toEqual(0);
+});
```
cjyuan, line 30: Could consider testing a few more samples in this script - higher chance to detect bugs in code.

cjyuan (commented):
- Function implementation is correct

- Some test descriptions could be made clearer

Sprint-3/2-practice-tdd/count.test.js
dcostaprakash: Got it. I have added both the test cases. Thank you

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
dcostaprakash: Sure, i have made the changes as suggested. Thank you

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
dcostaprakash: Got it. Have made the change. Thank you.

dcostaprakash:
Have modified the descriptions in the TDD to make the test cases more clear. Thank you

cjyuan:
LGTM.

illicitonion:

PR #1182, opened 2026-03-05
London | 26-ITP-January | Khaliun Baatarkhuu | Sprint 3 | Practice tdd. 

khaliun-dev:
Self checklist

## Changelist

Completed all the exercises from Practice TDD. 

## Questions

None at the moment.

### DIFF
--- a/Sprint-3/2-practice-tdd/repeat-str.js
+++ b/Sprint-3/2-practice-tdd/repeat-str.js
@@ -1,5 +1,8 @@
-function repeatStr() {
-  return "hellohellohello";
+//function repeatStr() {
+//  return "hellohellohello";
+function repeatStr(str, count) {
+  if (count >= 0) return str.repeat(count);
+  else throw new Error("Invalid format: count cannot be");
 }
 
 module.exports = repeatStr;
--- a/Sprint-3/2-practice-tdd/repeat-str.test.js
+++ b/Sprint-3/2-practice-tdd/repeat-str.test.js
@@ -21,12 +21,25 @@ test("should repeat the string count times", () => {
 // When the repeatStr function is called with these inputs,
 // Then it should return the original `str` without repetition.
 
+test("should return the original string when count is 1", () => {
+  expect(repeatStr("hello", 1)).toEqual("hello");
+});
+
 // Case: Handle count of 0:
 // Given a target string `str` and a `count` equal to 0,
 // When the repeatStr function is called with these inputs,
 // Then it should return an empty string.
 
+test("should return an empty string when count is 0", () => {
+  expect(repeatStr("hello", 0)).toEqual("");
+});
+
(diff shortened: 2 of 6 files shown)


Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,8 @@
-function repeatStr() {
-  return "hellohellohello";
+//function repeatStr() {
+//  return "hellohellohello";
+function repeatStr(str, count) {
+  if (count >= 0) return str.repeat(count);
```
favourO, line 4: The main logic is working here. Could you make this a little easier to read by expanding the if / else block, and have another look at the error message? At the moment it feels a bit unfinished, so a clearer message would help someone understand what went wrong.

Sprint-3/2-practice-tdd/repeat-str.test.js
```diff
@@ -21,12 +21,25 @@ test("should repeat the string count times", () => {
...
+
 // Case: Handle negative count:
 // Given a target string `str` and a negative integer `count`,
 // When the repeatStr function is called with these inputs,
 // Then it should throw an error, as negative counts are not valid.
+
+
+test("should throw an error when count is negative", () => {
+  expect(() => repeatStr("hello", -1)).toThrow();
+});
```
favourO, line 45: Nice to see this edge case covered. At the moment the test only checks that an error is thrown. Do you think it would make the test more helpful if it also checked that the error message explains what went wrong?

favourO (commented):
Nice work overall, count and getOrdinalNumber are clear and the tests cover the main cases well. I’ve left 2 comments on repeatStr and it's test to help clean up the final implementation before approval.

Sprint-3/2-practice-tdd/repeat-str.js
khaliun-dev: function repeatStr(str, count) {
  if (count >= 0) {
    return str.repeat(count);
  } else {
    throw new Error("Invalid argument: 'count' must be a non-negative number.");
  }
}

module.exports = repeatStr;

Sprint-3/2-practice-tdd/repeat-str.test.js
khaliun-dev: test("should throw a helpful error when count is negative", () => {
  expect(() => repeatStr("hello", -1))
    .toThrow("Count cannot be negative");
});

khaliun-dev:
@favourO Thank you for reviewing my PR. 

Could you, please, have another look? I think i have addressed both issues you have raised, properly.

Cheers, Khaliun

favourO (commented):
Nice job, 
We are ready to close now.

illicitonion:

PR #1186, opened 2026-03-05
Sheffield | ITP-Jan-26 | Hayriye Saricicek | Sprint 2 | Module Structuing and Testing Data

mshayriyesaricicek:
Self checklist

## Changelist

Coding for Sprint 2

### DIFF
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -16,7 +16,10 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
+    // this will be a number to 1 decimal place
     const bmi = weight / (height * height);  //calculation to calculate BMI
-    return bmi.toFixed(1);  
+    return Number(bmi.toFixed(1));  
 }
 console.log(calculateBMI(70, 1.73));
+
+// tested and it works, the function calculates the BMI correctly and returns it to 1 decimal place.
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -16,4 +16,7 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
+    const bmi = weight / (height * height);  //calculation to calculate BMI
+    return bmi.toFixed(1);  
 }
+console.log(calculateBMI(70, 1.73));
--- a/Sprint-2/3-mandatory-implement/2-cases.js
+++ b/Sprint-2/3-mandatory-implement/2-cases.js
@@ -24,7 +24,8 @@ function toUpperSnakeCase(str) {
         // checks if the character is an uppercase letter or a digit
  
     }).join('');
-  // joins the array of characters back into a single string and replaces spaces with underscores
+  // joins the array of characters back into a single string and replaces spaces or special 
+  // characters such as "?" with underscores
 }
 
 console.log(toUpperSnakeCase("there-once was/a young lady from+London"));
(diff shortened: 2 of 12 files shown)


Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -16,4 +16,7 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    const bmi = weight / (height * height);  //calculation to calculate BMI
+    return bmi.toFixed(1);
```

```javascript
  
```

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,19 @@
...
+function toUpperSnakeCase(str) {
+ 
+    return str.toUpperCase().split('').map(function(c) {  
+    // splits string into individual characters and maps each character to a new value
+   
+        return /[A-Z0-9]/.test(c) ? c : '_';
+        // checks if the character is an uppercase letter or a digit
+ 
+    }).join('');
+  // joins the array of characters back into a single string and replaces spaces with underscores
```
cjyuan: The comment on line 27 and the code on lines 20-26 do not quite match because
space characters are not the only characters that will get replaced by `_` in your implementation. 

We can also use `.replace()` or `.replaceAll()` to achieve to same outcome -- less code to write.

Sprint-2/3-mandatory-implement/3-to-pounds.js
```diff
@@ -4,3 +4,21 @@
...
+    //removes last 2 digits to get pounds
+const pence = paddedPenceNumberString.substring(paddedPenceNumberString.length - 2).padEnd(2, "0");
+    //gets last 2 digits to get pence, if less than 2 digits adds zeros to end of string but shouldnt be less than 3
+    //due to padStart above
+return `£${pounds}.${pence}`;
+    // returns string with pounds and pence in correct format
+}
+
+console.log(getPenceString("399p"));
+    // tested with 123p 1200p 24589p 89p 9p and 0p
```
cjyuan, line 24: Indentation is off.

Have you installed **prettier** VSCode extension and enabled formatting on save/paste on VSCode
as recommended in 
?

Sprint-2/3-mandatory-implement/1-bmi.js
mshayriyesaricicek: Thank you for pointing this out.  I have researched this now and understand that the answer would have been a number if I hadn't requested it to be to 1 decimal place.  The tofixed(1) converts the number to a string.  In this case it doesn't effect the outcome of the answer printed but I now understand that if I did need to do calculations on the answer then it would need to be converted back to a number in order to do this so it is good practice to convert back to a number.  

I have amended the program to convert the string answer back to a number and print a number.

I know as good practice there are usually inbuilt checks in case people input incorrect data.  If you want me to do this I can.

Sprint-2/3-mandatory-implement/2-cases.js
mshayriyesaricicek: I have changed line 27 to show that it replacds special characters as well as spaces with underscore.

I have researched .replace() and .replaceAll() and see that this works on the whole string at once so wouldn't need to split the string then join it again.  Thank you for the advice.

Sprint-2/3-mandatory-implement/3-to-pounds.js
mshayriyesaricicek: I did install prettier and enabled formatting on save/paste.  I may have done it after I had done this or it may not be working.

Sprint-2/3-mandatory-implement/1-bmi.js
cjyuan: As long as you are aware of the type of the data you are returning, its is enough. 

The spec was not clear on this anyway, and returning the result as a number as its shorting coming too.

Sprint-2/3-mandatory-implement/3-to-pounds.js
cjyuan: Understood. 

I cannot keep track of whom I had reminded of this. So I just nagged the person with this reminder whenever I spotted improperly formatted code.

cjyuan:
All good! Well done!

illicitonion:

PR #1192, opened 2026-03-06
Birmingham | 26-ITP-26 | Tasleem Adedokun | Sprint 3 | Coursework practice tdd

tasleemadedokun:
## Learners, PR Template

Self checklist

## Changelist

Brief;

I have done all the required coursework and also made committed changes.

### DIFF
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -22,3 +22,11 @@ test("should count multiple occurrences of a character", () => {
 // And a character `char` that does not exist within `str`.
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of `char` were found.
+
+
+test("should return 0 when the character does not exist in the string", () => {
+  const str = "aaaaa";
+  const char = "z";
+  const count = countChar(str, char);
+  expect(count).toEqual(0);
+});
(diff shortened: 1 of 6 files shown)


Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,26 @@
+
 function getOrdinalNumber(num) {
-  return "1st";
+  const absNum = Math.abs(num); // handle negative numbers
```
cjyuan, line 3: Not sure why you made the assumption that `num` can be a negative value.

In you plan to reject invalid values, a stricter rule would be, only accept non-negative integers.

Sprint-3/2-practice-tdd/count.test.js

The original specification did not clearly state whether the character match should be case-sensitive. Most people would probably assume that it is, but to demonstrate our understanding or clarify the assumption we made, we could add test cases to convey this. For examples,

cjyuan (commented):
Function implementation is correct.

Test cases are well described and cover all possibility.

Sprint-3/2-practice-tdd/count.test.js
tasleemadedokun: > Could consider testing a few more samples in this script - higher chance to detect bugs in code.
> 
> The original specification did not clearly state whether the character match should be case-sensitive. Most people would probably assume that it is, but to demonstrate our understanding or clarify the assumption we made, we could add test cases to convey this. For examples,
> 
> * A case to show that the match is case sensitive
> * A case to show that the function is expected to work also for non-alphabets

@cjyuan Thank you very much for always taking the time to highlight detailed information to work on, despite your busy schedules. I am glad it has been completed. However, at a free time I will go back to look into it. Once again, thank you very much.

illicitonion:

PR #1193, opened 2026-03-06
Birmingham | ITP-Jan-26 | Ayodeji Ayorinde | Sprint 2 | Structuring and Testing Data

Ayogit1:
## Learners, PR Template

Self checklist

## Changelist

This the complete of Sprint 2 tasks

### DIFF
--- a/Sprint-2/2-mandatory-debug/2.js
+++ b/Sprint-2/2-mandatory-debug/2.js
@@ -1,7 +1,8 @@
 // Predict and explain first...
 
 // Predict the output of the following code:
-// =============> Write your prediction here
+// =============> Write your prediction here. It may run but give incorrect results because of const num & no parameter specified
+
 
 const num = 103;
 
@@ -15,10 +16,19 @@ console.log(`The last digit of 806 is ${getLastDigit(806)}`);
 
 // Now run the code and compare the output to your prediction
 // =============> write the output here
+// The last digit of 42 is 3
+// The last digit of 105 is 3
+// The last digit of 806 is 3
 // Explain why the output is the way it is
-// =============> write your explanation here
+// =============> write your explanation here. num is a constant variable, it's value cannot change.
 // Finally, correct the code to fix the problem
 // =============> write your new code here
-
+function getLastDigit() {
+  return num.toString().slice(-1);
+}
 // This program should tell the user the last digit of each number.
 // Explain why getLastDigit is not working properly - correct the problem
+// getLastDigit is not working properly because the value for num is fixed at '103'
+function getLastDigit(num) {
+  return num.toString().slice(-1);
+}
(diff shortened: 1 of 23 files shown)


Sprint-2/2-mandatory-debug/2.js
```diff
@@ -15,10 +16,19 @@ console.log(`The last digit of 806 is ${getLastDigit(806)}`);
 
 // Now run the code and compare the output to your prediction
 // =============> write the output here
+// The last digit of 42 is 3
+// The last digit of 105 is 3
+// The last digit of 806 is 3
 // Explain why the output is the way it is
-// =============> write your explanation here
+// =============> write your explanation here. num is a constant variable, it's value cannot change.
```
cjyuan: You fixed the bug, but it is not because of the reason you described on line 23. (You could replace the `const` on line 7 by `let` and the code would still produce the same result).

Can you use AI to help you figure out the actual reason, and revise your explanation?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -16,4 +16,5 @@
 
 function calculateBMI(weight, height) {
     // return the BMI of someone based off their weight and height
+    return Math.round((weight) / (height ** 2)).toFixed(1)
```
cjyuan: - Have you tested your function to see if it returns the correct value? For example, what value do you expect `calculateBMI(22.34, 1)` to return? Does your function return the value you expect? 

- In addition, what **type** of value do you expect your function to return? A number or a string?

```javascript
  
```

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,7 @@
 // You will need to come up with an appropriate name for the function
 // Use the MDN string documentation to help you find a solution
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
+function upCase(strCar) {
```
cjyuan: The current function name does not quite tell people it returns the converted string in upper snake case format. 

Can you suggest a more descriptive function name?

Sprint-2/2-mandatory-debug/2.js
Ayogit1: Section redone completely.

Sprint-2/3-mandatory-implement/1-bmi.js
Ayogit1: Section redone completely.

Sprint-2/3-mandatory-implement/2-cases.js
Ayogit1: @cjyuan , Section redone completely.

cjyuan:
Look good.

illicitonion:

PR #1194, opened 2026-03-06
London | 26-ITP-Jan | Ebrahim Moqbel | Sprint 3 | Implement and Rewrite Tests

Ebrahim-Moqbel:
## Learners, PR Template

Self checklist

- [x ] I have titled my PR with Region | Cohort | FirstName LastName | Sprint | Assignment Title
- [ x] My changes meet the requirements of the task
- [x ] My changes follow the [style guide](https://curriculum.codeyourfuture.io/guides/reviewing/style-guide/)

## Changelist
completed the implement and rewrite test as follows:
- implemeted getAngleType and rewrote the test using jest 
- Implemented isProperFraction using jest tests
-implemented getCardValue with validation and worte jest test using .toThrow() for invalid input

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -11,6 +11,14 @@
 // execute the code to ensure all tests pass.
 
 function isProperFraction(numerator, denominator) {
+  if (denominator >numerator && denominator !==0) {
+    return ture ;
+    
+  }
+  else{
+    return false;
+  }
+  return Math.abs(numerator) < Math.abs(denominator);
   // TODO: Implement this function
 }
 
@@ -31,3 +39,16 @@ function assertEquals(actualOutput, targetOutput) {
 
 // Example: 1/2 is a proper fraction
 assertEquals(isProperFraction(1, 2), true);
+
+//2/1 is not a proper fraction
+assertEquals(isProperFraction(2,1), false);
+//0/1 is a proper fraction
+assertEquals(isProperFraction(0,1), true);
+//negative fraction -1/2 is a proper fraction
+assertEquals(isProperFraction(-1,2), true);
+//negative fraction -1/-2 is not a proper fraction
+assertEquals(isProperFraction(-1,-2), false);
+//fraction with zero denominator is not a proper fraction
+assertEquals(isProperFraction(1,0), false);
+//fraction with zero numerator is a proper fraction
+assertEquals(isProperFraction(0,5), true);
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -11,15 +11,8 @@
 // execute the code to ensure all tests pass.
 
 function isProperFraction(numerator, denominator) {
-  if (denominator >numerator && denominator !==0) {
-    return ture ;
-    
-  }
-  else{
-    return false;
-  }
+  if (denominator === 0) return false; // matches your test expectation
   return Math.abs(numerator) < Math.abs(denominator);
-  // TODO: Implement this function
 }
 
 // The line below allows us to load the isProperFraction function into tests in other files.
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
@@ -40,7 +40,7 @@ function getCardValue(card) {
   }
   if (!["♠", "♥", "♦", "♣"].includes(cardFace)) {
   }
-  throw new Error(`Invalid card rank: ${rank}`);
+  throw new Error(`Invalid card`);
 }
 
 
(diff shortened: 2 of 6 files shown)


Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -11,6 +11,14 @@
 // execute the code to ensure all tests pass.
 
 function isProperFraction(numerator, denominator) {
+  if (denominator >numerator && denominator !==0) {
+    return ture ;
```
favourO: A couple of things to look at in this block:
- There is a small typo in **ture**. (maybe you wanted to write true)
- The last return Math.abs(...) line will never run because the function already returns above.

Your tests allow negative numbers. Can you make sure your implementation follows the mathematical definition of a proper fraction so those cases work correctly?

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
```diff
@@ -14,7 +14,36 @@ test(`should return "Acute angle" when (0 < angle < 90)`, () => {
...
+  // Test various reflex angles, including boundary cases
+  expect(getAngleType(181)).toEqual("Reflex angle");
+  expect(getAngleType(270)).toEqual("Reflex angle");
+  expect(getAngleType(359)).toEqual("Reflex angle");
+});
+
 // Case 6: Invalid angles
+test(`should throw an error when (angle < 0 or angle >= 360)`, () => {
+  expect(() => getAngleType(-1)).toThrow("Invalid angle");
+  expect(() => getAngleType(360)).toThrow("Invalid angle");
```
favourO: These tests expect the function to throw an error, but the implementation currently returns "Invalid angle" as a string. It would be good to make sure the tests and implementation follow the same behaviour.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -23,8 +23,27 @@
...
+  if (rank === "A") {
+    return 11;
+  }
+  if (+rank >= 2 && +rank <= 9) {
+    return +rank;
+  }
+  if (["K", "10", "Q", "J"].includes(rank)) {
+    return 10;
+  }
+  if (!["♠", "♥", "♦", "♣"].includes(cardFace)) {
```
favourO, line 41: This if block appears to be empty at the moment, so it isn’t doing anything. Can you check if you accidentally removed some logic?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -23,8 +23,27 @@
...
+  }
+  throw new Error(`Invalid card rank: ${rank}`);
```
favourO: Can you confirm that the error messages in the implementation match what the tests expect?

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -11,8 +11,27 @@ test(`Should return 11 when given an ace card`, () => {
 
 // Suggestion: Group the remaining test data into these categories:
 //   Number Cards (2-10)
+test(`Should return the correct value for number cards (2-10)`), () => {
+  expect(getCardValue("2♠")).toEqual(2);
```
favourO, line 15: The test call here looks slightly mis-structured, the callback function appears to be outside the test(...) call.
It might be worth checking the syntax so the test runner can execute this block correctly.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -11,8 +11,27 @@ test(`Should return 11 when given an ace card`, () => {
...
+  expect(getCardValue("10♠")).toEqual(10);
+}
 //   Face Cards (J, Q, K)
-//   Invalid Cards
+test(`Should return 10 when given a face card (J, Q, K)`), () => {
+  expect(getCardValue("J♠")).toEqual(10);
+  expect(getCardValue("Q♠")).toEqual(10);
+  expect(getCardValue("K♠")).toEqual(10);
+}
+//   Invalid Card suits
```
favourO: Nice test cases.
Can you confirm that the implementation throw the same error messages that the tests expect here? we want to keep the behaviour predictable

also can you check if you properly closed the test block for this _test(`Should throw an error when given an invalid card suit`)_

favourO (commented):
Nice work,
I left a few comments on specific blocks where the implementation and tests don’t quite match yet, and where there are a couple of small syntax or logic issues. 

Once you address them, the PR should be in a good place.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
Ebrahim-Moqbel: Thank you favourO. I have matched the function message thron error to the tests expectation.

favourO (commented):
Great job

illicitonion:

PR #1196, opened 2026-03-06
London | 26-ITP-Jan | Divine Mankrado | Sprint 3 | Implement and Rewrite Tests

divinmank:
## Learners, PR Template

Self checklist

## Changelist

Finished coursework on sprint 3 implement-rewrite-and-tests.

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -11,7 +11,7 @@
 // execute the code to ensure all tests pass.
 
 function isProperFraction(numerator, denominator) {
-  // TODO: Implement this function
+  return numerator > 0 && denominator > 0 && numerator < denominator;
 }
 
 // The line below allows us to load the isProperFraction function into tests in other files.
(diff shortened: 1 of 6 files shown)


Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -31,3 +31,19 @@ function assertEquals(actualOutput, targetOutput) {
 
 // Example: 1/2 is a proper fraction
 assertEquals(isProperFraction(1, 2), true);
+
```
favourO, line 19: Great job, your current implementation treats every negative fraction as false, and your Jest tests are written to match that. But what is the actual rule for a “proper fraction”? 

It might be worth checking whether your implementation is based on the maths definition, or whether the tests and implementation have just been made consistent with each other.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -41,12 +73,53 @@ function assertEquals(actualOutput, targetOutput) {
...
 try {
   getCardValue("invalid");
-
-  // This line will not be reached if an error is thrown as expected
   console.error("Error was not thrown for invalid card");
-} catch (e) {}
+} catch (e) { }
+
+// Test invalid rank
+try {
```
favourO, line 102: It’s good that you’re thinking about invalid inputs as well as valid ones. One small improvement to think about: right now these checks only confirm that some error was thrown. Would it make the test more helpful if it also checked that the error explains the problem clearly?

favourO (commented):
Nice job so far,
Please check the comments

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
divinmank: Thanks for pointing this out! I just checked the meaning of a proper fraction and realized that negative fractions can still be proper if the absolute value of the numerator is less than the absolute value of the denominator.

I’ve updated my implementation to use Math.abs() and adjusted the tests to include negative proper fractions as well.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
divinmank: Thanks for the suggestion! It makes sense. I’ve just updated the tests so they also check the error message to ensure the error clearly explains the problem with the input.

favourO (commented):
All done now

illicitonion:

PR #1212, opened 2026-03-07
London | 26-ITP-Jan | Johnny Vargas | Sprint 3 | Practice TDD

JohnnyBoyV:
## Learners, PR Template

Self checklist

## Changelist

Did as asked for the Practice TDD.

### DIFF
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -22,3 +22,9 @@ test("should count multiple occurrences of a character", () => {
 // And a character `char` that does not exist within `str`.
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of `char` were found.
+test("should count no occurrences of a character", () => {
+  const str = "FFFFF"
+  const char= "z"
+  const count = countChar(str, char);
+  expect(count).toEqual(0);
+});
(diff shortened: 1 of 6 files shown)


cjyuan (commented):
Mistakenly sent.

Sprint-3/2-practice-tdd/count.test.js

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
cjyuan, line 1: The test categories in this script do not yet cover all possible numbers.

Sprint-3/2-practice-tdd/repeat-str.test.js
```diff
@@ -20,13 +20,30 @@ test("should repeat the string count times", () => {
...
+  const count = 1;
+  const repeatedStr = repeatStr(str, count);
+  expect(repeatedStr).toEqual("hello");
+});
 
 // Case: Handle count of 0:
 // Given a target string `str` and a `count` equal to 0,
 // When the repeatStr function is called with these inputs,
 // Then it should return an empty string.
+test("should print nothing", () => {
```
cjyuan, line 34: - "print" would suggest the function is expected to output (instead of returning) the resulting string.

- In TDD, the test description (or test name) should clearly explain what behavior the function should exhibit. 

Can you also include the "condition" part into these descriptions?

cjyuan (commented):
- Function implementation is correct.

- Some test descriptions could be made clearer.

illicitonion:

PR #1240, opened 2026-03-10
Sheffield | 26-ITP-Jan | Martha Ogunbiyi | Sprint 3 | coursework/sprint-3-implement-and-rewrite

marthak1:
## Learners, PR Template

Self checklist

## Changelist

- Implemented functions and wrote test cases using jest

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -11,8 +11,7 @@
 // execute the code to ensure all tests pass.
 
 function isProperFraction(numerator, denominator) {
-  const properFraction = numerator / denominator;
-  return properFraction > 0 && properFraction < 1;
+ return Math.abs(numerator) < Math.abs(denominator)
 }
 
 // The line below allows us to load the isProperFraction function into tests in other files.
@@ -41,7 +40,7 @@ assertEquals(isProperFraction(1, 1), false);
 assertEquals(isProperFraction(8, 3), false)
 
 // Negative Values
-assertEquals(isProperFraction(-3, 8), false);
+assertEquals(isProperFraction(-3, 8), true);
 assertEquals(isProperFraction(-8, -3), false);
 assertEquals(isProperFraction(-3, -8), true);
 
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -11,7 +11,8 @@
 // execute the code to ensure all tests pass.
 
 function isProperFraction(numerator, denominator) {
-  // TODO: Implement this function
+  const properFraction = numerator / denominator;
+  return properFraction > 0 && properFraction < 1;
 }
 
 // The line below allows us to load the isProperFraction function into tests in other files.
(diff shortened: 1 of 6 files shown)


Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
Edu-Vin, line 1: I think you should check the definition of a proper fraction again and see if your test case for negative numbers is correct expecially -3/8. I will advise you to use Match.abs()

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -11,8 +11,24 @@ test(`Should return 11 when given an ace card`, () => {
...
+});
 //   Face Cards (J, Q, K)
+test(`Should return 10`, () => {
+  expect(getCardValue("K♠")).toEqual(10);
+});
 //   Invalid Cards
+test(`Should throw an error`, () => {
+  expect(() => {
+    getCardValue("♠9");
+  }).toThrow();
```
Edu-Vin, line 25: This is good. To take it further, add a test to see if the returned error message is what you are expecting.

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
marthak1: Thanks for the feedback! My implementation checks whether the value of the fraction is between 0 and 1, which is the standard mathematical definition of a proper fraction. Based on that definition:

-3/8 → -0.375 → not proper

-8/-3 → 2.66… → not proper

-3/-8 → 0.375 → proper

Using Math.abs() would treat negative fractions like -3/8 as proper, even though their value is negative. That’s why I didn’t use Math.abs(), and why my tests reflect the “between 0 and 1” definition. Happy to adjust if a different definition is expected.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
marthak1: Thank you for the feedback . Would add further test to see what error messages they return.

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
Edu-Vin: The first rule of a proper fraction is that the numerator is less than the denominator.  Your solution will fail on most negative numbers. Reason why a proper fraction is when the absolute value of the numerator is less than the absolute value of the denominator. A proper fraction is always less than 1

Based on your code, let's consider the following examples
you are returning `properFraction > 0 && properFraction < 1`

-1/2 will return false because the division is -0.5, and it is not greater than 0. But this is wrong as the result is less than 1 and the first rule of numerator < denominator stands.

Same thing with -3/8.

You can check this for further info https://www.splashlearn.com/math-vocabulary/proper-fraction

Math.abs() helps you work with negative numbers.

You can do more research on it.

Edu-Vin:
Your recent commits added a solution for 2-practice-tdd. The Readme file for Sprint-3 says 1 pull request per directory, so 2-practice-tdd is supposed to have its own pull request.

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
marthak1: Thanks for explaining! I understand now that you're using the definition based on absolute values. I’ll update my code and tests to use Math.abs() so it matches that approach.

Edu-Vin:
@marthak1 noticed a lot of activities on the pull request. tag me on the pull request when it is ready to be reviewed again

marthak1:
@Edu-Vin added the changes to the proper fraction function

illicitonion:

PR #1241, opened 2026-03-11
London | 26-ITP-Jan | Miriam Jorna | Sprint 3 | Practice TDD

miriamjorna:
## Learners, PR Template

Self checklist

## Changelist

Tests done and tested, files written and tested.

## Questions

For the Ordinal Numbers question I have made an exception for the number 0 because I believe "0th" does not exist. Claude.ai disagrees and says "0th" should be included. What say you?

### DIFF
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.js
@@ -1,5 +1,23 @@
 function getOrdinalNumber(num) {
-  return "1st";
+   if (num === 0) {
+    return "0";
+  }
+  const lastTwoDigits = num % 100;
+  const lastDigit = num % 10;
+
+  if (lastTwoDigits === 11 || lastTwoDigits === 12 || lastTwoDigits === 13) {
+    return `${num}th`;
+  }
+  if (lastDigit === 1) {
+    return `${num}st`;
+  }
+  if (lastDigit === 2) {
+    return `${num}nd`;
+  }
+  if (lastDigit === 3) {
+    return `${num}rd`;
+  }
+  return `${num}th`;
 }
 
 module.exports = getOrdinalNumber;
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.js
@@ -6,16 +6,16 @@ function getOrdinalNumber(num) {
   const lastTwoDigits = num % 100;
   const lastDigit = num % 10;
 
-  switch (true) {
-    case lastTwoDigits === 11:
-    case lastTwoDigits === 12:
-    case lastTwoDigits === 13:
-      return `${num}th`;
-    case lastDigit === 1:
+  if (lastTwoDigits === 11 || lastTwoDigits === 12 || lastTwoDigits === 13) {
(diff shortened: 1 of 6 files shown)


Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,23 @@
...
+    return "0";
+  }
+  const lastTwoDigits = num % 100;
+  const lastDigit = num % 10;
+
+  if (lastTwoDigits === 11 || lastTwoDigits === 12 || lastTwoDigits === 13) {
+    return `${num}th`;
+  }
+  if (lastDigit === 1) {
+    return `${num}st`;
```
Edu-Vin: Your use of if statement works, but a switch statement will be better.

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,12 @@
-function repeatStr() {
-  return "hellohellohello";
+function repeatStr(str, count) {
+  if (count < 0) {
+    throw new Error("Negative counts are not valid");
+  }
+  let result = "";
+  for (let i = 0; i < count; i++) {
```
Edu-Vin, line 9: This works. Don't you think if the count is 0, it should return the empty string early?

miriamjorna: It was in that order because I followed the format of the task as given. Now changed into better order with the test for 0 first. 
Thank you!

Edu-Vin: Oh, that is not what I meant. I mean, you should check if count === 0, then return an empty string before the for loop

Sprint-3/2-practice-tdd/get-ordinal-number.js
miriamjorna: Replaced the if statements with switch statements as requested. Looks nice and clean :)

Sprint-3/2-practice-tdd/repeat-str.js
miriamjorna: Oh! Apologies for overlooking to include that *and* then not seeing what you meant when you pointed it out! 
Sorted now.

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,23 +1,25 @@
...
-    return `${num}nd`;
-  }
-  if (lastDigit === 3) {
-    return `${num}rd`;
+  switch (true) {
+    case lastTwoDigits === 11:
+    case lastTwoDigits === 12:
+    case lastTwoDigits === 13:
+      return `${num}th`;
+    case lastDigit === 1:
```
Edu-Vin: Well done!. This works, but I think it can be cleaner. Move the case check for lastTwoDigits to the original if statement you had. Then the switch statement will be for lastDigit. 
`switch(lastDigit) {
    case 1:
return `${num}st`;
}`

miriamjorna: Took me a moment to straighten that out but I believe it's done now. Thank you!

illicitonion:

PR #1250, opened 2026-03-12
Sheffield | 26-ITP-jan | Richard Frimpong | Sprint 3 | Stretch Coursework

Richiealx:
## Learners, PR Template

Self checklist

## Changelist

Completed the Sprint 3 stretch tasks.

### Password Validator
- Implemented password validation rules in `password-validator.js`
- Added Jest tests in `password-validator.test.js`

### Credit Card Validator
- Implemented credit card validation logic in `card-validator.js`
- Added Jest tests in `card-validator.test.js`

### Testing
All tests were run using:

All tests passed successfully.

### DIFF
--- a/Sprint-3/4-stretch/card-validator.js
+++ b/Sprint-3/4-stretch/card-validator.js
@@ -1,0 +1,33 @@
+// Validate whether a credit card number meets the rules from card-validator.md
+function validateCreditCardNumber(cardNumber) {
+  // Rule 1: the value must be exactly 16 characters long and contain only digits.
+  if (!/^\d{16}$/.test(cardNumber)) {
+    return false;
+  }
+
+  // Rule 2: not all digits can be the same.
+  const uniqueDigits = new Set(cardNumber);
+  if (uniqueDigits.size < 2) {
+    return false;
+  }
+
+  // Rule 3: the final digit must be even.
+  const lastDigit = Number(cardNumber[cardNumber.length - 1]);
+  if (lastDigit % 2 !== 0) {
+    return false;
+  }
+
+  // Rule 4: the sum of all digits must be greater than 16.
+  let sum = 0;
+  for (const digit of cardNumber) {
+    sum += Number(digit);
+  }
+
+  if (sum <= 16) {
+    return false;
+  }
+
+  return true;
+}
+
+module.exports = validateCreditCardNumber;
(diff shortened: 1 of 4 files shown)


Sprint-3/4-stretch/password-validator.js
```diff
@@ -1,6 +1,38 @@
+// A small list of previous passwords that must not be reused.
+const previousPasswords = ["Password1!", "Welcome2#", "Strong3$"];
+
 function passwordValidator(password) {
-    return password.length < 5 ? false : true
-}
+  // Password must be at least 5 characters long.
+  if (password.length < 5) {
```
ykamal, line 11: Could `password` just not be provided at all?

Richiealx: Hi @ykamal,

Good point, thank you.

I updated the function to handle the case where a password is not provided.  
Previously the function accessed `password.length`, which would throw an error if the value was `undefined`.

I added an input validation check at the start of the function to ensure the password exists and is a string. If it is missing or not a string, the function now safely returns `false`.

I also added test cases to cover these scenarios.

Sprint-3/4-stretch/card-validator.js
```diff
@@ -0,0 +1,33 @@
+// Validate whether a credit card number meets the rules from card-validator.md
+function validateCreditCardNumber(cardNumber) {
+  // Rule 1: the value must be exactly 16 characters long and contain only digits.
```
kyle-tightest, line 3: writing down these rules as comments in code is very helpful for making sure you complete all of them and also helpful for debugging in plain english. good job

Sprint-3/4-stretch/card-validator.js
```diff
@@ -0,0 +1,33 @@
+// Validate whether a credit card number meets the rules from card-validator.md
+function validateCreditCardNumber(cardNumber) {
```
kyle-tightest, line 2: this is more of an advanced thing, but as a general rule if you're implementing a function that returns a boolean (true/false) its best to start the function name with "is" or "has" to make it clear to a reader that it returns a boolean. So this one could be called "isCreditCardNumberValid" for example

Sprint-3/4-stretch/card-validator.js
```diff
@@ -0,0 +1,33 @@
...
+
+  // Rule 4: the sum of all digits must be greater than 16.
+  let sum = 0;
+  for (const digit of cardNumber) {
+    sum += Number(digit);
+  }
+
+  if (sum <= 16) {
+    return false;
+  }
```
kyle-tightest, line 28: while this works, it is slightly inefficient as it requires you to loop through the entire list of digits before saying true/false even if the first 4 3 digits are more than 16 for example. Can you think of a way to make this more efficient and return earlier?

illicitonion:

PR #1253, opened 2026-03-12
Sheffield | 26-ITP-Jan | Martha Ogunbiyi | Sprint 3 | dead code

marthak1:
## Learners, PR Template

Self checklist

## Changelist

Removed dead code from exercise

### DIFF
--- a/Sprint-3/3-dead-code/exercise-2.js
+++ b/Sprint-3/3-dead-code/exercise-2.js
@@ -2,12 +2,8 @@
 // The countAndCapitalisePets function should continue to work for any reasonable input it's given, and you shouldn't modify the pets variable.
 
 const pets = ["parrot", "hamster", "horse", "dog", "hamster", "cat", "hamster"];
-const capitalisedPets = pets.map((pet) => pet.toUpperCase());
 const petsStartingWithH = pets.filter((pet) => pet[0] === "h");
 
-function logPets(petsArr) {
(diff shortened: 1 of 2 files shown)


Sprint-3/3-dead-code/exercise-2.js
```diff
@@ -2,12 +2,8 @@
 // The countAndCapitalisePets function should continue to work for any reasonable input it's given, and you shouldn't modify the pets variable.
 
 const pets = ["parrot", "hamster", "horse", "dog", "hamster", "cat", "hamster"];
-const capitalisedPets = pets.map((pet) => pet.toUpperCase());
 const petsStartingWithH = pets.filter((pet) => pet[0] === "h");
```
ykamal: Would this work if the character were uppercase?

marthak1: Thanks for the feedback. Given this current code the answer is no, because the filter is only looking for  lowercase character "h". I have now updated the filter for case sensitivity

ykamal: Nicely done! 👍

ykamal (approved):
Nicely done 👍 I think this looks good. Excellent work on figuring out that characters in a string can be lowercase OR uppercase, we can't predict it. And kudos on getting that edge case sorted.
https://en.wikipedia.org/wiki/Edge_case

illicitonion:

PR #1258, opened 2026-03-13
Scotland | ITP JAN-2026 | Tuan Nguyen | Sprint 3 | 2- Practice-tdd

Jacknguyen4438:
## Learners, PR Template

Self checklist

## Changelist

I have done practice and fully understand how to write a test for both normal JS and Jest test.

## Questions

No Question.

### DIFF
--- a/Sprint-3/2-practice-tdd/count.js
+++ b/Sprint-3/2-practice-tdd/count.js
@@ -5,7 +5,7 @@ function countChar(fullString,findCharacters) {
     throw new Error("Invalid input");
   };
   
-  if (findCharacters.length < 1) {
+  if (findCharacters.length !== 1) {
     throw new Error("findCharacters must be a single character");
   }
 
--- a/Sprint-3/2-practice-tdd/count.js
+++ b/Sprint-3/2-practice-tdd/count.js
@@ -1,5 +1,33 @@
-function countChar(stringOfCharacters, findCharacter) {
-  return 5
+function countChar(fullString,findCharacters) {
+  let count = 0;
+  // both of these if statement is for input validation if the input not a string.
+  if( typeof findCharacters !== "string" || typeof fullString !== "string"){
+    throw new Error("Invalid input");
+  };
+  
+  if (findCharacters.length < 1) {
+    throw new Error("findCharacters must be a single character");
+  }
+
+  for (let i = 0; i < fullString.length; i++){
+    if(fullString[i] === findCharacters){
+      count++;
+    }
+  }
+  return count;
 }
 
 module.exports = countChar;
+
+
+function assertTest(testInput,testCheck){
+ console.assert(
+  testInput === testCheck,
+  `Expect ${testInput} equal to ${testCheck}`
+ );
+};
+
+assertTest(countChar("whale fat hat cat","a"),4)
+assertTest(countChar("I need to lean more and know more","e"),5)
+assertTest(countChar("the city centre currently have a carnival","c"),4)
+assertTest(countChar("the cruise ship in in transit to south America","s"),4)
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -23,5 +23,23 @@ test("should count multiple occurrences of a character", () => {
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of `char` were found.
 test( "should return count of zero occurrences of a character", () =>{
-   expect(countChar("go home and study","c")).toEqual(0)
+   expect(countChar("go home and study","c")).toEqual(0);
+});
+
+// Scenario: Case-sensitive matching
+// Given a string containing both uppercase and lowercase versions of a letter,
+// When the function is called with a lowercase character,
+// Then it should only count the lowercase occurrences.
+
+test("should treat character matching as case-sensitive", () => {
+ expect(countChar("AaAa","A")).toEqual(2);
+});
+
+// Scenario: Non-alphabet characters
+// Given a string containing digits and symbols,
+// When the function is called with a non-alphabet character,
+// Then it should correctly count occurrences of that character.
+
+test("should count occurrences of non-alphabet characters", () => {
+  expect(countChar("1-2-3-1-1","1")).toEqual(3);
 });
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -22,3 +22,6 @@ test("should count multiple occurrences of a character", () => {
 // And a character `char` that does not exist within `str`.
 // When the function is called with these inputs,
 // Then it should return 0, indicating that no occurrences of `char` were found.
+test( "should return count of zero occurrences of a character", () =>{
+   expect(countChar("go home and study","c")).toEqual(0)
+});
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.js
@@ -1,5 +1,50 @@
 function getOrdinalNumber(num) {
-  return "1st";
+// This is for num argument validation that go inside the function.
+if(typeof num !== "number" || Number.isNaN(num)){
+  throw new Error ("Invalid input: the value must be a number");
+}
+
+// This allow float number to be round into it nearest integer.
+num = Math.round(num);
+
+ if(num % 100 === 11 ||num % 100 === 12 || num % 100 === 13 ){
+  return `${num}th`;
+ }
+switch( true ){
+  case num % 10 === 1 : return `${num}st`;
+  case num % 10 === 2 : return `${num}nd`;
+  case num % 10 === 3 : return `${num}rd`;
+  default : return `${num}th`;
+}
 }
 
 module.exports = getOrdinalNumber;
+
+
+function testAssert (inputNumber,outputNUmber){
+console.assert(
+ inputNumber === outputNUmber,
+ `Test failed: expected ${outputNUmber}, but got ${inputNumber}`
+);
+};
+//Basic test of first digit integer
+testAssert(getOrdinalNumber(1), "1st");
+testAssert(getOrdinalNumber(2), "2nd");
+testAssert(getOrdinalNumber(3), "3rd");
+testAssert(getOrdinalNumber(4), "4th");
+//Test for 11, 12 and 13 
+testAssert(getOrdinalNumber(11), "11th");
+testAssert(getOrdinalNumber(12), "12th");
+testAssert(getOrdinalNumber(13), "13th");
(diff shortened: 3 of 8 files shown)


Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,33 @@
...
+function countChar(fullString,findCharacters) {
+  let count = 0;
+  // both of these if statement is for input validation if the input not a string.
+  if( typeof findCharacters !== "string" || typeof fullString !== "string"){
+    throw new Error("Invalid input");
+  };
+  
+  if (findCharacters.length < 1) {
+    throw new Error("findCharacters must be a single character");
+  }
```
cjyuan: Is `findCharacters` allowed to be more than one characters?

Sprint-3/2-practice-tdd/count.test.js

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,50 @@
...
+
+ if(num % 100 === 11 ||num % 100 === 12 || num % 100 === 13 ){
+  return `${num}th`;
+ }
+switch( true ){
+  case num % 10 === 1 : return `${num}st`;
+  case num % 10 === 2 : return `${num}nd`;
+  case num % 10 === 3 : return `${num}rd`;
+  default : return `${num}th`;
+}
```
cjyuan: Can you look up the pros and cons between these two styles of expressing the code?

```
  if(num % 100 === 11 ||num % 100 === 12 || num % 100 === 13 ){
```

```
  const lastTwoDigits = num % 100;
  if (lastTwoDigits === 11 || lastTwoDigits === 12 || lastTwoDigits === 13 ) {
```

?

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,50 @@
 function getOrdinalNumber(num) {
-  return "1st";
+// This is for num argument validation that go inside the function.
+if(typeof num !== "number" || Number.isNaN(num)){
+  throw new Error ("Invalid input: the value must be a number");
+}
+
+// This allow float number to be round into it nearest integer.
+num = Math.round(num);
```
cjyuan: Why do you decide to round the numbers with decimal places? Why not reject them?

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -18,3 +18,64 @@ test("should append 'st' for numbers ending with 1, except those ending with 11"
...
+  expect(getOrdinalNumber(12)).toEqual("12th");
+  expect(getOrdinalNumber(13)).toEqual("13th");
+  expect(getOrdinalNumber(111)).toEqual("111th");
+  expect(getOrdinalNumber(212)).toEqual("212th");
+});
+
+// Case 5: All other numbers
+// When the number does not end with 1, 2, or 3,
+// Then the function should append "th".
+test("should append 'th' for all other numbers", () => {
```

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,24 @@
...
-}
+function repeatStr(fullString,repeatCount) {
+ if( repeatCount < 0){
+  throw new Error("repeatCount cannot be negative")
+ };
+
+ switch (true){
+  case repeatCount === 0 : return "";
+  default: return fullString.repeat(repeatCount) ;
+ };
```
cjyuan: What do you think about this writing style?
```
if (repeatCount === 0) return "":

return fullString.repeat(repeatCount);
```

Jacknguyen4438:
@cjyuan Hello thank you for the review, I see there are many part that need to be fix and improve. I will check all 5 part that you would like me to explain and fix each of them. When I finish I will mention you again to ask for a review.

Sprint-3/2-practice-tdd/count.js
Jacknguyen4438: Thank you for feed back, I see that I have make some validation error. I can see that instead of validation for just a specific character. So instead I will change this test so that is can check correctly

Sprint-3/2-practice-tdd/count.test.js
Jacknguyen4438: @cjyuan Thank you for the suggestion here is my jest test case for this 2 scenario:

// Scenario: Case-sensitive matching
// Given a string containing both uppercase and lowercase versions of a letter,
// When the function is called with a lowercase character,
// Then it should only count the lowercase occurrences.

test("should treat character matching as case-sensitive", () => {
 expect(countChar("AaAa","A")).toEqual(2);
});

// Scenario: Non-alphabet characters
// Given a string containing digits and symbols,
// When the function is called with a non-alphabet character,
// Then it should correctly count occurrences of that character.

test("should count occurrences of non-alphabet characters", () => {
  expect(countChar("1-2-3-1-1","1")).toEqual(3);
});

Sprint-3/2-practice-tdd/get-ordinal-number.js
Jacknguyen4438: @cjyuan  Thank you for the feed back, 

I see that that for the first line the main pro is that :
- The pros is that the code shorter, no extra variable  and is good for when 1 argument pass to each condition.
-   The cons is that the code it harder to understand for some new beginner and might caught with unexpected error. Also the argument num is now pass 3 time for each condition and it can cause issue if the code is broke outside the function.

For the second line I see that:
- The pros of this line is that is more solid and safer then the first line. It help other coder or beginner to understand the code easier. Also the value now store inside a container that can be used safely  and if need be can be change later.
The cons for this line of code as far as I can see is that is only add 1 more line of code.

So after I see both Pros and cons of these 2 line,  understand for the best I will used the second line of code since is more save and easier to change and fix. I will make this change inside my code. Thank you.

Sprint-3/2-practice-tdd/get-ordinal-number.js
Jacknguyen4438: For this one I, the main reason is that since I would like to for the function to work even is the user pass a float number value. When pass inside the function the argument num will automatically will be round to their nearest number then do the reminder. But I see now is might caught the issue since it will return not what the user put in so I will max change that reject the decimal value.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
Jacknguyen4438: Thank you for the feed back I will fix it right away

Sprint-3/2-practice-tdd/repeat-str.js
Jacknguyen4438: Thank you for the reply , I think that if I used the same method as you using with this line code. It will run and return the value just fine since is a if else statement but more shorthand compact. I understand that by using switch case is make make the code look better but in long term can cause some issue in the long run. So for the best I will try this if else style for easier reading and maintain the code.

Jacknguyen4438:
@cjyuan Hello, thank you for spending your time your time to do my PR, I am very grateful for you help. I have make change base on you suggestion and ready to be review

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,51 @@
...
+ if(mainReminder === 11 ||mainReminder  === 12 || mainReminder  === 13 ){
+  return `${num}th`;
+ }
+switch( true ){
+  case secondReminder === 1 : return `${num}st`;
+  case secondReminder=== 2 : return `${num}nd`;
+  case secondReminder=== 3 : return `${num}rd`;
+  default : return `${num}th`;
+}
 }
```
cjyuan: Some of the code is not consistently formatted.

Have you installed the **prettier** VSCode extension and enabled "Format on save/paste" on VSCode,
as recommended in https://github.com/CodeYourFuture/Module-Structuring-and-Testing-Data/blob/main/readme.md ?

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,51 @@
 function getOrdinalNumber(num) {
-  return "1st";
+// This is for num argument validation that go inside the function.
+if(typeof num !== "number" || Number.isNaN(num) || !Number.isInteger(num)){
```
cjyuan: Some of  these tests are redundant. Can you simplify the check?

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
cjyuan: The test description still reads: `"should append 'th' for all other numbers"`

Jacknguyen4438:
@cjyuan Hello I have make change needed for review again. If I have make another error I need to be done please let me know.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -62,12 +62,12 @@ test("should append 'th' for all other numbers", () => {
 // When the input is a float,
 // Then the function should round it and return the correct ordinal.
 test("should return error for invalid input", () => {
-  expect(() =>getOrdinalNumber(1.2)).toThrow("Invalid input");
```
cjyuan, line 65: Note: "Return an error" is not the same as "throw an error".

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -62,12 +62,12 @@ test("should append 'th' for all other numbers", () => {
...
-  expect(() =>getOrdinalNumber(2.5)).toThrow("Invalid input");   
-  expect(() =>getOrdinalNumber(10.51)).toThrow("Invalid input"); 
-  expect(() =>getOrdinalNumber(12.49)).toThrow("Invalid input"); 
-  expect(() =>getOrdinalNumber(12.5)).toThrow("Invalid input"); 
+  expect(() => getOrdinalNumber(1.2)).toThrow("Invalid input");
+  expect(() => getOrdinalNumber(1.8)).toThrow("Invalid input");
+  expect(() => getOrdinalNumber(2.5)).toThrow("Invalid input");
+  expect(() => getOrdinalNumber(10.51)).toThrow("Invalid input");
+  expect(() => getOrdinalNumber(12.49)).toThrow("Invalid input");
+  expect(() => getOrdinalNumber(12.5)).toThrow("Invalid input");
```
cjyuan, line 70: Why test so many decimal numbers with decimal point? If the objective is to show the function should treat any number with decimal points as invalid input, one or two samples should be enough.

Under the "invalid input" category, it would probably be more useful to show what other kinds of input are considered invalid. For examples, 
- "300" -- a string that resembles an integer
- Infinity
- NaN

cjyuan (commented):
Changes look good.

illicitonion:

PR #1264, opened 2026-03-13
London | 26-ITP-Jan | Zadri Abdule | Sprint 3 | coursework/sprint-3-implement-and-rewrite

Zadri415:
## Learners, PR Template

Self checklist

Completed the required tasks - implemented the required functions and rewrote the tests using Jest.

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -12,7 +12,14 @@
 
 function isProperFraction(numerator, denominator) {
   // TODO: Implement this function
+  if (denominator === 0){
+    return false;
+  }
+  
+  return numerator < denominator;
+
 }
+exports.isProperFraction = isProperFraction;
 
 // The line below allows us to load the isProperFraction function into tests in other files.
 // This will be useful in the "rewrite tests with jest" step.
@@ -25,9 +32,17 @@ function assertEquals(actualOutput, targetOutput) {
     `Expected ${actualOutput} to equal ${targetOutput}`
   );
 }
+exports.assertEquals = assertEquals;
 
 // TODO: Write tests to cover all cases.
 // What combinations of numerators and denominators should you test?
 
 // Example: 1/2 is a proper fraction
 assertEquals(isProperFraction(1, 2), true);
+assertEquals(isProperFraction(2, 1), false);
+assertEquals(isProperFraction(0, 1), true);
+assertEquals(isProperFraction(1, 0), false);
+assertEquals(isProperFraction(-1, 2), true);
+assertEquals(isProperFraction(1, -2), false);
+assertEquals(isProperFraction(-1, -2), false);
+
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -16,7 +16,7 @@ function isProperFraction(numerator, denominator) {
     return false;
   }
   
-  return numerator < denominator;
+  return Math.abs(numerator) < Math.abs(denominator);
 
 }
 exports.isProperFraction = isProperFraction;
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
@@ -38,8 +38,8 @@ exports.assertEquals = assertEquals;
 // What combinations of numerators and denominators should you test?
 
 // Example: 1/2 is a proper fraction
-assertEquals(isProperFraction(1, 2), true);
-assertEquals(isProperFraction(2, 1), false);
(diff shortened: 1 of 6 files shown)


Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -25,9 +32,17 @@ function assertEquals(actualOutput, targetOutput) {
...
 
 // Example: 1/2 is a proper fraction
-assertEquals(isProperFraction(1, 2), true);
+assertEquals(isProperFraction(1, 2), true); // 2 is greater than 1, so this should return true.
+assertEquals(isProperFraction(2, 1), false); // 1 is not greater than 2, so this should return false.
+assertEquals(isProperFraction(0, 1), true);
+assertEquals(isProperFraction(1, 0), false);
+assertEquals(isProperFraction(-1, 2), true);
+assertEquals(isProperFraction(1, -2), false);
+assertEquals(isProperFraction(-1, -2), false);
```
cjyuan, line 47: Can you lookup if -1/-2, 1/-2, -1/2 are considered proper fractions, and then update 
your implementation and tests accordingly?

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -23,7 +23,34 @@
...
+    "K": 10  // King is worth 10
+  };
+  if (faceCardValues[rank]) {  // Check if the rank is a face card or ace
+    return faceCardValues[rank]; // If it is, return the corresponding value
+  }
+
+  const numericValue = parseInt(rank); // Try to parse the rank as a number
+  if (numericValue >= 2 && numericValue <= 10) { // check if the numeric value is between 2 and 10
+    return numericValue; // If it is, return the numeric value
+  }
```
cjyuan, line 47: In JavaScript, strings that represent valid numeric literals in the language can be safely converted to equivalent numbers. For examples, "0x02", "2.1", or "0002".

```
getCardValue("0x02♠");
getCardValue("2.1♠");
getCardValue("0002♠");
```

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
```diff
@@ -8,13 +8,49 @@ const getAngleType = require("../implement/1-get-angle-type");
...
+test(`should return "Invalid angle" when angle (angle <= 0 or angle >= 360)`, () => {
+  expect(getAngleType(-1)).toEqual("Invalid angle");
+  expect(getAngleType(0)).toEqual("Invalid angle");
+  expect(getAngleType(-0.1)).toEqual("Invalid angle");
+  expect(getAngleType(-10)).toEqual("Invalid angle");
+  expect(getAngleType(360)).toEqual("Invalid angle");
+  expect(getAngleType("string")).toEqual("Invalid angle");
+  expect(getAngleType(null)).toEqual("Invalid angle");
+  expect(getAngleType(undefined)).toEqual("Invalid angle");
+  expect(getAngleType(NaN)).toEqual("Invalid angle");
```
cjyuan: The spec has already made an assumption the function can expect the argument is a valid number. So the tests  on lines 51-54 are not needed. 

One possible side-effect of including the test on line 51 is that it may suggest the function is expected to return "Invalid angle" when the the argument is a string.  However, with your current implementation, a function call such as `getAngleType("90")` would return "Right angle" instead of "Invalid angle".

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -9,10 +9,33 @@ test(`Should return 11 when given an ace card`, () => {
...
+  expect(getCardValue("Q♦")).toEqual(10);
+  expect(getCardValue("K♣")).toEqual(10);
+});
+
 //   Invalid Cards
+test(`Should throw an error for invalid cards`, () => {
+  expect(() => getCardValue("1♠")).toThrow("Invalid card");
+  expect(() => getCardValue("11♣")).toThrow("Invalid card");
+  expect(() => getCardValue("A♦")).toThrow("Invalid card");
+});
```
cjyuan, line 38: Could also consider testing the following cases:
- Valid rank + invalid suit character
- Valid rank with no suit
- Empty string


Note: Why is "A♦" considered an invalid card?

Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
Zadri415: Thanks! I'll take another look and update it.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
Zadri415: Thanks for pointing that out. I'll make sure to update it.

illicitonion:

PR #1269, opened 2026-03-15
London | 26-ITP-Jan | Miriam Jorna | Sprint 3 | Coursework implement and rewrite tests

miriamjorna:
## Learners, PR Template

Self checklist

## Changelist

Coursework for Sprint 3: Implement and rewrite tests

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
@@ -15,6 +15,9 @@
 // execute the code to ensure all tests pass.
 
 function getAngleType(angle) {
+  if (angle === 0) {    
+    return "Zero angle";
+  }
   if (angle > 0 && angle < 90) {
     return "Acute angle";
   }
@@ -33,9 +36,7 @@ function getAngleType(angle) {
   if (angle === 360) {    
     return "This is a circle, not an angle";
   }
-  if (angle === 0) {    
-    return "";
-  }
+  
   return "Invalid angle"; 
   
 }
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
@@ -33,6 +33,9 @@ function getAngleType(angle) {
   if (angle === 360) {    
     return "This is a circle, not an angle";
   }
+  if (angle === 0) {    
+    return "";
+  }
   return "Invalid angle"; 
   
 }
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
@@ -15,7 +15,26 @@
 // execute the code to ensure all tests pass.
 
 function getAngleType(angle) {
-  // TODO: Implement this function
+  if (angle > 0 && angle < 90) {
+    return "Acute angle";
+  }
+  if (angle === 90) {
+    return "Right angle";
+  }
+  if (angle > 90 && angle < 180) {
+    return "Obtuse angle";
+  }
+  if (angle === 180) {
+    return "Straight angle";
+  }
+  if (angle > 180 && angle < 360) {
+    return "Reflex angle";
+  }
+  if (angle === 360) {    
+    return "This is a circle, not an angle";
+  }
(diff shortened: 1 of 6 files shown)


Sprint-3/1-implement-and-rewrite-tests/implement/2-is-proper-fraction.js
```diff
@@ -11,8 +11,15 @@
 // execute the code to ensure all tests pass.
 
 function isProperFraction(numerator, denominator) {
-  // TODO: Implement this function
+  // denominator cannot be 0 in a proper fraction
+  if (denominator === 0) {
```
ykamal, line 15: Excellent check for `0`. Whenever it comes to a division, always check for `0`. 🎃

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -23,6 +23,27 @@
 
 function getCardValue(card) {
   // TODO: Implement this function
+  if (typeof card !== "string") {
```
ykamal, line 26: So, what you're trying to ensure is that this argument is provided, is a string and can be used down the line. This is a solid approach to do that 👍 
However, you also need to ensure it's long enough.

Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
```diff
@@ -15,7 +15,26 @@
 // execute the code to ensure all tests pass.
 
 function getAngleType(angle) {
```
ykamal, line 17: This is good work 👍 You're accounting for most types of angles. But there is a problem here. 0 is also *sometimes* considered a valid angle. Regardless, if your code assumes that it's valid, it should take that into account, and if it thinks 0 degrees is not a valid angle, it should take that into account as well.
How would you do that?

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -7,4 +7,29 @@ const isProperFraction = require("../implement/2-is-proper-fraction");
...
+  expect(isProperFraction(2, -1)).toEqual(false);
+  expect(isProperFraction(-2, -1)).toEqual(false);
+  expect(isProperFraction(5, 0)).toEqual(false);
+  expect(isProperFraction(-5, 0)).toEqual(false);
+});
+
+test(`should return true when the numerator is zero and the denominator is non-zero`, () => {
+  expect(isProperFraction(0, 5)).toEqual(true);
+  expect(isProperFraction(0, -5)).toEqual(true);
+}
```
ykamal: There is a formatting issue here. I think it's worth investing some time in VS Code tooling that can help you find these out, such as Prettier auto-formating on save and code linting.

ykamal (changes requested):
Overall, very well done 👍 
There are a few things here and there. I've left some comments in the corresponding lines.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
miriamjorna: Thank you for your kind words here and above. Have added a string length criterium.

Sprint-3/1-implement-and-rewrite-tests/implement/1-get-angle-type.js
miriamjorna: Schrödinger's angle... Initially I thought to return an empty string would be best, but on second thought it's better to just name what we have: "Zero angle"

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
miriamjorna: Ahh, apologies, I do use Prettier. This is due to a user fault/keyboard issue on my side (sometimes on *save* my keyboard wants to make an extra diacritic; deleted that but didn't spot the extra return it generated)

miriamjorna: Many thanks for your time - all should be sorted now

ykamal (approved):
Nice work! I'm going to approve this.

illicitonion:

PR #1271, opened 2026-03-16
London | 26-ITP-Jan | Zadri Abdule | Sprint 3 | coursework practice tdd

Zadri415:
## Learners, PR Template

Self checklist

Completed the Sprint 3 practice TDD tasks
- Wrote tests before implementation 
- Feature implemented to pass tests

### DIFF
--- a/Sprint-3/2-practice-tdd/count.js
+++ b/Sprint-3/2-practice-tdd/count.js
@@ -1,5 +1,13 @@
 function countChar(stringOfCharacters, findCharacter) {
-  return 5
+  let count = 0;
+  for (let i = 0; i < stringOfCharacters.length; i++) {
+    if (stringOfCharacters[i] === findCharacter) {
+      count++;
+    }
+  }
+  
+  return count;
 }
+
 
 module.exports = countChar;
--- a/Sprint-3/2-practice-tdd/count.js
+++ b/Sprint-3/2-practice-tdd/count.js
@@ -1,11 +1,20 @@
 function countChar(stringOfCharacters, findCharacter) {
-  let count = 0;
-  for (let i = 0; i < stringOfCharacters.length; i++) {
-    if (stringOfCharacters[i] === findCharacter) {
-      count++;
-    }
-  }
+  // treat undefined/null input as empty string -> no matches
+  if (stringOfCharacters == null) return 0;
   
+  // coerce to string safely
+  const s = String(stringOfCharacters);
+  const target = String(findCharacters);
+  
+  if (target === '') {
+    // define behaviour for empty target
+    return 0;
+    }
+
+  let count = 0;
+  for (const ch of s) {
+    if (ch === target) count++;
+  }
   return count;
 }
 
--- a/Sprint-3/2-practice-tdd/count.test.js
+++ b/Sprint-3/2-practice-tdd/count.test.js
@@ -1,18 +1,10 @@
 // implement a function countChar that counts the number of times a character occurs in a string
-const countChar = require("./count");
-function countChar(stringOfCharacters, findCharacters){
-  let count = 0;
-  for (let i = 0; i < stringOfCharacters.length; i++){
-    if (stringOfCharacters[i] === findCharacters){
-      count++;
-    }
-  }
-  return count;
-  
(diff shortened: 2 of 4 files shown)


Sprint-3/2-practice-tdd/count.test.js
```diff
@@ -1,9 +1,11 @@
 // implement a function countChar that counts the number of times a character occurs in a string
-const countChar = require("./count");
+const countChar = require("./count");{
+}
```
ykamal: Do you think there is a syntax issue here? How would you ensure your IDE can catch such issues? 
Worth checking: file formatting with Priettier

ykamal (changes requested):
I think overall this is good. Good job so far 👍  

There are a few things you should look into:

1. For counting the occurrences of a character in a string, you could use something like `.split()` and do a count on the returning array. The benefit is that it's cleaner and probably more performant. 
2. This function could also fail if the string provided doesn't match certain criteria. As a developer, you should account for those. Details in the comment associated.
3. There is a syntax issue in the `count.test.js` file.  Details in the comment associated.
4. `getOrdinalNumber` could fail if the number is not as expected. Details in the comment associated.

Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,13 @@
 function countChar(stringOfCharacters, findCharacter) {
```
ykamal, line 1: What if the string `stringOfCharacters` was not provided at all? What if it's undefined, or null?
You should take into account where your code could produce an error and account for that.
How do you think you'd account for those scenarios?

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,18 @@
 function getOrdinalNumber(num) {
```
ykamal, line 9: Since it's a number, it could be negative or a float (decimal). 
Does this function assume that the number is always going to be a positive whole number?
How would you account for those scenarios?

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -17,4 +17,42 @@ test("should append 'st' for numbers ending with 1, except those ending with 11"
...
+test("should append 'th' for numbers ending with 11, 12, or 13", () => {
+  expect(getOrdinalNumber(11)).toEqual("11th");
+  expect(getOrdinalNumber(12)).toEqual("12th");
+  expect(getOrdinalNumber(13)).toEqual("13th");
+});
+
+// Case 5: ALL other numbers
+// When the number does not end with 1, 2, 3, 11, 12, or 13,
+// Then the function should return a string by appending "th" to the number.
+test("should append 'th' for all other numbers", () => {
```
mirandaweston, line 53: Might be worth adjusting this test to be more specific, as 'all other numbers' failing won't tell us which numbers they are? Other tests look great!

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,18 @@
...
-  return "1st";
+  const lastDigits = num % 100;
+  if (lastDigits >= 11 && lastDigits <= 13) {
+    return num + "th";
+  }
+  switch (lastDigits % 10) {
```
mirandaweston: Nice use of a switch statement here!

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
Zadri415: Thanks. I'll make sure to adjust it.

Sprint-3/2-practice-tdd/count.test.js
Zadri415: I'll make sure to have another look at it. Thanks for the feedback.

Sprint-3/2-practice-tdd/get-ordinal-number.js
Zadri415: Thanks.

illicitonion:

PR #1283, opened 2026-03-30
Sheffield | 26-ITP-Jan |  Seti Mussa |  Sprint 3 | Coursework  pratice tdd

Seti-Jemal:
Self checklist

## Changelist

Write a function that counts how many times a given character appears in a string, and passes all Jest tests.

## Questions

N/A

### DIFF
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.test.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.test.js
@@ -33,16 +33,20 @@ test("should append 'rd' for number anding with 3 expect those ending with 13", 
   expect(getOrdinalNumber(43)).toEqual("43rd");
 });
 
-// Case 4: Number ending with 4  (but not 14)
-test("should append 'th' for number ending with 4 expcet those ending with 14", () => {
-  expect(getOrdinalNumber(4)).toEqual("4th");
-  expect(getOrdinalNumber(44)).toEqual("44th");
-  expect(getOrdinalNumber(54)).toEqual("54th");
+// Case 4: Number ending with 11,12, 13
+test("should append 'th' for number ending with 11, 12, 13", () => {
+  expect(getOrdinalNumber(11)).toEqual("11th");
+  expect(getOrdinalNumber(12)).toEqual("12th");
+  expect(getOrdinalNumber(13)).toEqual("13th");
 });
 
-// Case 5: Number ending with 5 (but not 15)
-test("should append 'th' for number ending with 5 except those ending with 15", () => {
+// Case 5: Numbers ending with 0, 4, 5, 6, 7, 8, or 9 (but not those ending with 11, 12, or 13)
+test("should append 'th' for number ending with 0, 4, 5, 6, 7, 8, or 9 ,(but not those ending with 11, 12, or 13)", () => {
+  expect(getOrdinalNumber(4)).toEqual("4th");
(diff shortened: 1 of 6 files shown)


Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -16,5 +16,33 @@ const getOrdinalNumber = require("./get-ordinal-number");
...
+  expect(getOrdinalNumber(44)).toEqual("44th");
+  expect(getOrdinalNumber(54)).toEqual("54th");
+});
+
+// Case 5: Number ending with 5 (but not 15)
+test("should append 'th' for number ending with 5 except those ending with 15", () => {
+  expect(getOrdinalNumber(5)).toEqual("5th");
+  expect(getOrdinalNumber(55)).toEqual("55th");
+  expect(getOrdinalNumber(65)).toEqual("65th");
 });
```
cjyuan: What are so special about 14 and 15? 

These two test categories do not yet cover all possible numbers not included in the previous test cases. You can consider just creating one test category for all the numbers that should be appended with "th". 
However, you should avoid describing these numbers as "all other numbers" because
when a test fails with such a message, it may be unclear what "other numbers" actually refers to.

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,5 +1,5 @@
-function repeatStr() {
-  return "hellohellohello";
+function repeatStr(n, str) {
+  return str.repeat(n);
 }
```
cjyuan: Can this function pass the tests in `repeat-str.test.js`?

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
Seti-Jemal: Hi CJ, 
Thank you for the feedback. I have updated the test, removed 14 and 15, and added the special case as required.

Sprint-3/2-practice-tdd/repeat-str.js
Seti-Jemal: I have fixed it now.

Sprint-3/2-practice-tdd/get-ordinal-number.test.js
```diff
@@ -16,5 +16,40 @@ const getOrdinalNumber = require("./get-ordinal-number");
...
+  expect(getOrdinalNumber(11)).toEqual("11th");
+  expect(getOrdinalNumber(12)).toEqual("12th");
+  expect(getOrdinalNumber(13)).toEqual("13th");
+  expect(getOrdinalNumber(111)).toEqual("111th");
+  expect(getOrdinalNumber(112)).toEqual("112th");
+  expect(getOrdinalNumber(113)).toEqual("113th");
+});
+
+// Case 5: Numbers ending with 0, 4, 5, 6, 7, 8, or 9 (but not those ending with 11, 12, or 13)
+test("should append 'th' for number ending with 0, 4, 5, 6, 7, 8, or 9 ,(but not those ending with 11, 12, or 13)", () => {
```
cjyuan, line 47: Part the the condition specified in this description does not carry new information, and therefore can be dropped.

cjyuan (commented):
Changes look good.

illicitonion:

PR #1298, opened 2026-05-04
 Glasgow | ITP-May-26 | Alasdair MacDonald | Sprint 3 | Coursework 3 implement and rewrite

MacDonald91:
## Learners, PR Template

Self checklist

## Changelist
resubmission.

## Questions

n/a

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
@@ -23,6 +23,30 @@
 
 function getCardValue(card) {
   // TODO: Implement this function
+
+  const suits = ["♠", "♥", "♦", "♣"];
+
+  if (typeof card !== "string" || card.length < 2) {
+    throw new Error("Invalid card");
+  }
+
+  const suit = card.slice(-1);
+  const rank = card.slice(0, -1);
+
+  if (!suits.includes(suit)) {
+    throw new Error("Invalid card");
+  }
+
+  if (rank === "A") return 11;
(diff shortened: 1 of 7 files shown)


Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -1,54 +1,31 @@
...
-  );
-}
+  // Face cards
+  if (["J", "Q", "K"].includes(rank)) return 10;
 
-// TODO: Write tests to cover all outcomes, including throwing errors for invalid cards.
-// Examples:
-assertEquals(getCardValue("9♠"), 9);
+  // Number cards
+  const validNumbers = ["2","3","4","5","6","7","8","9","10"];
```
illicitonion: This works but involves writing down a lot of numbers, and it would be hard to spot if any numbers in the middle aren't valid (e.g. if 7 wasn't valid) - can you think how to write this check in a more expressive way?

illicitonion (commented):
Looks good - just one thing to think about

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
MacDonald91: Thanks for the feedback!

I’ve updated the number validation to remove the hardcoded list and replaced it with a numeric range check (2–10). This makes the logic more expressive and avoids the risk of missing values.

Let me know if there’s anything else I can improve.

illicitonion:

PR #1311, opened 2026-05-07
Inline implementation of padStart

illicitonion:
Python Tutor doesn't support padStart.


RaihanSharif (approved):
I have tested the code with the following inputs:

2
12
123
-1
-12
-1232

Output is the same as padStart().

PR #1314, opened 2026-05-11
London | 26-ITP-May | Alex Jamshidi | Sprint 1 | Coursework

Alex-Jamshidi:
## Learners, PR Template

Self checklist

## Changelist

Completed solutions to sprint 1 coursework


Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -17,7 +17,14 @@ console.log(`The base part of ${filePath} is ${base}`);
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
 
-const dir = ;
-const ext = ;
+const lastDotIndex = filePath.lastIndexOf(".");
 
-// https://www.google.com/search?q=slice+mdn
\ No newline at end of file
+const dir = filePath.slice(1, lastSlashIndex);
```
illicitonion, line 22: I realise the diagram above is a bit ambiguous, but the dir actually includes the leading `/` as well

Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -17,7 +17,14 @@ console.log(`The base part of ${filePath} is ${base}`);
...
+const ext = filePath.slice(lastDotIndex);
```
illicitonion, line 23: This works for this particular path.

What would happen if I gave you a filePath that didn't contain a `.` at all? What would you expect it to return? What will your solution return?

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -7,3 +7,44 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
...
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+
+console.log(num);
+
+// num returns a value of 60, running it again returns a value of 15, looking at the code I see a .random() function
+// I can make a hypothesis that this returns a random number of some kind, hence the inconsistent result
+
+// Let's find out...
+// Code in inner-most brackets is evaluated first
```
illicitonion, line 17: Not quite, actually - it doesn't matter here, but the `Math.random()` will actually be evaluated before the inner parens.

The order of execution is _generally_ top-to-bottom, left-to-right, with a few exceptions:
* When we call a function, we evaluate the arguments left-to-right before we can call the function (so we need to evaluate `Math.random() * (maximum  minimum + 1) before we can call `Math.floor`).
* Things in parentheses get evaluated when we hit the opening parenthesis. So `1 + (2 * 3)` will evaluate 1, then evaluate `(2 * 3)` before it does the `1 + (3)`.

The point of the parentheses here is to say "do `maximum - minimum` before `*`", not "do `maximum - minimum` before `Math.random()`" :)

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -7,3 +7,44 @@ const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
...
+console.log(number);
+console.log(Math.floor(number));
+
+// because Math.random() creates a new random number every time it is run, I had to set the
+// result to a variable so I can use the same number in multiple functions
+// When number is 87.52, Math.floor is 87, when number is 70.02, math.floor is 70
+// The function Math.floor appears to be rounding down
+// Checking the documentation, this is true, it returns the largest integer <= the given number
+
+// Therefore Math.floor(Math.random() * (maximum - minimum + 1)) must give random numbers between 0 and 99
```
illicitonion, line 46: One precision thing here - when talking about ranges, we'd generally describe whether the bounds are inclusive of exclusive. Here "between 0 and 99" it isn't clear whether that includes 0 or starts from 1. I'd generally write this "between 0 and 99, inclusive" (or in other contexts, perhaps "between 0 (inclusive) and 100 (exclusive)")

illicitonion (commented):
LGTM - I'll mark this as complete, but please do think about the comments I've left :)

illicitonion:

PR #1316, opened 2026-05-11
London | 26-ITP-May | Alex Jamshidi | Sprint 2 | Coursework

Alex-Jamshidi:
## Learners, PR Template

Self checklist

## Changelist

Completed solutions to sprint 2 coursework


Sprint-2/1-key-errors/1.js
```diff
@@ -13,8 +17,20 @@ function convertToPercentage(decimalNumber) {
 }
 
 console.log(decimalNumber);
+*/
 
 // =============> write your explanation here
+// Uncaught SyntaxError: Identifier 'decimalNumber' has already been declared
+// As predicted the function is trying to declare a variable that has already been determined
 
 // Finally, correct the code to fix the problem
```
illicitonion, line 26: Does this file now do what you'd expect when you run it?

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,5 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
```
illicitonion, line 17: Try calling this function - does it calculate the value you expect?

(e.g. with the example given of 70 and 1.73 which has an expected answer. Or try 50 and 2 - what do you get?)

illicitonion (commented):
Generally LGTM, but a couple of things to look at.

Sprint-2/1-key-errors/1.js
Alex-Jamshidi: Not a very useful function if input is immediately ignored
removed variable line entirely so percentage is calculated from function argument

Sprint-2/3-mandatory-implement/1-bmi.js
Alex-Jamshidi: ah!

I assumed that ^ was going to raise to the power as in scientific notation.
When I tested for 70 1.73, it was so close that I assumed it was a rounding error in the question. :')

I understand ^ is not that at all, have updated code to ** 2 for powers

Sprint-2/1-key-errors/1.js
illicitonion: I don't see this update - are you sure you committed+pushed?

Sprint-2/3-mandatory-implement/1-bmi.js
illicitonion: Makes sense!

For the future, if you think the expected outputs we give you for exercises are wrong, please raise it with us so we can fix them (or work out they're actually right)! :)

Sprint-2/1-key-errors/1.js
Alex-Jamshidi: updated now

illicitonion:

PR #1319, opened 2026-05-12
London | 26-ITP-May | Alex Jamshidi | Sprint 3 | Practice tdd

Alex-Jamshidi:
## Learners, PR Template

Self checklist

## Changelist

Added solutions to Sprint 3 Practice TDD

### DIFF
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.js
@@ -5,16 +5,16 @@ function getOrdinalNumber(num) {
 
   const lastTwoDigits = num.toString().slice(-2);
   const lastDigit = lastTwoDigits.slice(-1);
-  if (lastTwoDigits == 11 || lastTwoDigits == 12 || lastTwoDigits == 13) {
+  if (lastTwoDigits === 11 || lastTwoDigits === 12 || lastTwoDigits === 13) {
     return `${num}th`;
   }
-  if (lastDigit == 1) {
+  if (lastDigit === 1) {
     return `${num}st`;
   }
-  if (lastDigit == 2) {
+  if (lastDigit === 2) {
     return `${num}nd`;
   }
-  if (lastDigit == 3) {
+  if (lastDigit === 3) {
     return `${num}rd`;
   } else return `${num}th`;
 }
--- a/Sprint-3/2-practice-tdd/get-ordinal-number.js
+++ b/Sprint-3/2-practice-tdd/get-ordinal-number.js
@@ -3,8 +3,8 @@ function getOrdinalNumber(num) {
     throw new Error("Invalid number");
   }
 
-  const lastTwoDigits = num.toString().slice(-2);
-  const lastDigit = lastTwoDigits.slice(-1);
+  const lastTwoDigits = Number(num.toString().slice(-2));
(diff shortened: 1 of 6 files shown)


Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,13 @@
 function getOrdinalNumber(num) {
-  return "1st";
+  if (!Number.isInteger(num) || num < 1) {throw new Error("Invalid number");}
+
+  let number = num.toString().slice(-2);
```
illicitonion: "num" and "number" are very hard to distinguish variables - it's not clear from their names what the differences between them are. Can you think how to improve this?

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,13 @@
...
+  if (number == 11 || number == 12 || number == 13) {return `${num}th`;}
+  if (number.slice(-1) == 1) {return `${num}st`;}
```
illicitonion: I'd extract a variable for that `number.slice(-1)` operation that you're repeating three times.

illicitonion (commented):
LGTM - a couple of naming things to look at

Sprint-3/2-practice-tdd/get-ordinal-number.js
Alex-Jamshidi: renamed number to lastTwoDigits to make clearer what is happening

Sprint-3/2-practice-tdd/get-ordinal-number.js
Alex-Jamshidi: extracted and named lastDigit, code much more readable now

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,22 @@
 function getOrdinalNumber(num) {
-  return "1st";
+  if (!Number.isInteger(num) || num < 1) {
+    throw new Error("Invalid number");
+  }
+
+  const lastTwoDigits = num.toString().slice(-2);
+  const lastDigit = lastTwoDigits.slice(-1);
+  if (lastTwoDigits == 11 || lastTwoDigits == 12 || lastTwoDigits == 13) {
```
illicitonion: We almost never want to use `==` in JavaScript, and almost always want to prefer `===` - can you do some research as to why, and update your code to use `===` instead?

Alex-Jamshidi: So I did know the difference here, == doesn't care about type, just value.
I've not really thought about being strict with this when I know what the code is going to do, but I realise it's much better to restrict as much as possible to prevent any incorrect data leaking through. I will aim to always use === going forward, expect in cases where type make not matter.

updated code. thanks

illicitonion: Does your code now pass its tests? :)

Alex-Jamshidi: I thought I check them!
No they didn't pass, because I was now trying to compare a string to number
converted strings to numbers to pass === check

tests all pass

illicitonion:

PR #1321, opened 2026-05-14
Fix variable names

illicitonion:
Currently they're the wrong way around, which isn't the point of the exercise.


Poonam-raj (approved):
TIL you can't start variables with numbers, how did I not know this

PR #1329, opened 2026-06-02
London | 26-ITP-May | Zadri Abdule | Sprint 3 | Implement & rewrite tests

Zadri415:
## Learners, PR Template

Self checklist

## Changelist
Added  Jest tests for Sprint-3 implement and rewrite (exercises 1–3)

### DIFF
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
@@ -37,8 +37,23 @@ function getCardValue(card) {
   const rank = card.slice(0, -1);
   if (rank === "A") return 11;
   if (rank === "J" || rank === "Q" || rank === "K") return 10;
+}
 
- 
+if (
+    rank === "2" ||
+    rank === "3" ||
+    rank === "4" ||
+    rank === "5" ||
+    rank === "6" ||
+    rank === "7" ||
+    rank === "8" ||
+    rank === "9" ||
+    rank === "10"
+  ) {
+    return Number(rank);
+  }
+
+  throw new Error("Invalid card");
 }
 
 // The line below allows us to load the getCardValue function into tests in other files.
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
@@ -22,6 +22,7 @@
 // execute the code to ensure all tests pass.
 
 const validSuits = ["♠", "♣", "♥", "♦"];
+const validRanks = ["2", "3", "4", "5", "6", "7", "8", "9", "10", "J", "Q", "K", "A"];
 
 function getCardValue(card) {
   if (typeof card !== "string" || card.length < 2) {
@@ -37,9 +38,7 @@ function getCardValue(card) {
   if (rank === "A") return 11;
   if (rank === "J" || rank === "Q" || rank === "K") return 10;
 
-  const num = Number(rank);
-  if (Number.isInteger(num) && num >= 2 && num <= 10) return num;
-  throw new Error("Invalid card");
+ 
 }
 
 // The line below allows us to load the getCardValue function into tests in other files.
--- a/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
+++ b/Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
@@ -22,7 +22,41 @@
 // execute the code to ensure all tests pass.
 
 function getCardValue(card) {
-  // TODO: Implement this function
+  // Student-friendly step-by-step parsing and validation.
+  // 1) Expect a string like "A♠", "10♥", "K♦" where the last character is the suit emoji
+  // 2) The rank is everything before the final character (so "10" or "A")
+  // 3) Check the suit is one of the four allowed suits, and the rank is valid
+
+  if (typeof card !== "string") {
+    throw new Error('Card must be a string, e.g. "A♠"');
+  }
+
+  const text = card.trim();
+  if (text.length < 2) {
+    throw new Error(
+      'Invalid card: too short. Expect a rank and a suit, e.g. "10♠"'
+    );
+  }
+
+  // The suit is the last character, the rank is the rest
+  const suit = text[text.length - 1];
+  const rank = text.slice(0, -1);
+
+  const validSuits = new Set(["♠", "♥", "♦", "♣"]);
+  if (!validSuits.has(suit)) {
+    throw new Error(`Invalid suit "${suit}". Use one of: ♠ ♥ ♦ ♣`);
+  }
+
+  // Handle special ranks first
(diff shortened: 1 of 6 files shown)


Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -21,8 +21,25 @@
...
+  if (!validSuits.includes(suit)) {
+    throw new Error("Invalid card");
+  }
+
+  const rank = card.slice(0, -1);
+  if (rank === "A") return 11;
+  if (rank === "J" || rank === "Q" || rank === "K") return 10;
+
+  const num = Number(rank);
+  if (Number.isInteger(num) && num >= 2 && num <= 10) return num;
```
cjyuan: In JavaScript, strings that represent valid numeric literals in the language can be safely converted to equivalent numbers. For examples, "0x02", "2.000", or "0002".

```
getCardValue("0x02♠");
getCardValue("2.000♠");
getCardValue("0002♠");
```

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
```diff
@@ -14,7 +14,33 @@ test(`should return "Acute angle" when (0 < angle < 90)`, () => {
...
+});
+
 // Case 6: Invalid angles
+test(`should return "Invalid angle" when invalid`, () => {
+  expect(getAngleType(-1)).toEqual("Invalid angle");
+  expect(getAngleType(0)).toEqual("Invalid angle");
+  expect(getAngleType("90")).toEqual("Invalid angle");
+  expect(getAngleType(1000)).toEqual("Invalid angle");
+  expect(getAngleType(9999.999)).toEqual("Invalid angle");
+});
```
cjyuan: - When a test fails with the message "... when invalid", it may be unclear what "invalid" actually refers to.

- Could also test the other boundary case, 360.

- Regarding the test `expect(getAngleType("90")).toEqual("Invalid angle")`:
    - It is not immediately clear why `"90"` is considered an invalid angle
    - Why `"90"` is considered invalid, but `"91"` is not?
    - It may imply the function is expected to check the type of the parameter

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/2-is-proper-fraction.test.js
```diff
@@ -7,4 +7,36 @@ const isProperFraction = require("../implement/2-is-proper-fraction");
...
+});
+
+test(`should return true when numerator is zero and denominator is non-zero`, () => {
+  expect(isProperFraction(0, 1)).toEqual(true);
+  expect(isProperFraction(0, -1)).toEqual(true);
+  expect(isProperFraction(0, 100)).toEqual(true);
+  expect(isProperFraction(0, -100)).toEqual(true);
+});
+
+test(`should return true when absolute value of numerator is less than absolute value of denominator`, () => {
```
cjyuan, line 21: Note: Using pseudocode and notations like `abs(...)` or `| ... |` is a common and widely accepted practice for describing conditions in a concise manner.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
```diff
@@ -11,10 +11,34 @@ test(`Should return 11 when given an ace card`, () => {
 
 // Suggestion: Group the remaining test data into these categories:
 //   Number Cards (2-10)
+test(`Should return the numeric value for number cards`, () => {
```
cjyuan, line 19: Could also include a test to show that the function is expected to check the suit character.

cjyuan (commented):
Very solid function implementation and tests.

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
Zadri415: Thanks for the helpful feedback . I've updated the test description to make them clearer, added the missing boundary case for 360, revisit the "90" test so the invalid-input expectation makes more sense, and add a test to show that getCardValue checks the suit character too. I also appreciate the note about using mathematical notation in test descriptions.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
```diff
@@ -21,8 +21,39 @@
 // After you have implemented the function, write tests to cover all the cases, and
 // execute the code to ensure all tests pass.
 
+const validSuits = ["♠", "♣", "♥", "♦"];
+const validRanks = ["2", "3", "4", "5", "6", "7", "8", "9", "10", "J", "Q", "K", "A"];
```
cjyuan, line 25: What's the purpose of preparing `validRanks`? Could using it simplify the solution?

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/1-get-angle-type.test.js
```diff
@@ -14,7 +14,37 @@ test(`should return "Acute angle" when (0 < angle < 90)`, () => {
...
+
 // Case 5: Reflex angles
+test(`should return "Reflex angle" when (180 < angle < 360)`, () => {
+  expect(getAngleType(181)).toEqual("Reflex angle");
+  expect(getAngleType(270)).toEqual("Reflex angle");
+  expect(getAngleType(359)).toEqual("Reflex angle");
+});
+
 // Case 6: Invalid angles
+test(`should return "Invalid angle" when angle is not greater than 0`, () => {
```
cjyuan: Why not simply express "not greater than 0" as `<= 0`? It's fine to use pseudocode or well-known notation in test descriptions when it makes them more concise. (Same for the "not less than 360")

In this case, `<=` is a widely recognized operator meaning "less than or equal to".

Sprint-3/1-implement-and-rewrite-tests/rewrite-tests-with-jest/3-get-card-value.test.js
cjyuan, line 1: > ... add a test to show that getCardValue checks the suit character too. 

I could not find this test.

Sprint-3/1-implement-and-rewrite-tests/implement/3-get-card-value.js
Zadri415: To check if rank is valid, instead of comparing it to each value one by one.

cjyuan: Sounds good. Why not implement what you described? Otherwise, it is best practice to remove all unused variables to keep the code clean.

cjyuan:
Changes look good.

Zadri415:
Thank you.

PR #1353, opened 2026-06-17
Manchester | 26-ITP-May | Joanne O’Malley | Sprint 3 | Practice TDD

joanne342:
## Learners, PR Template

Self checklist

## Changelist

Sprint 3 | Practice TDD

### DIFF
--- a/Sprint-3/2-practice-tdd/repeat-str.js
+++ b/Sprint-3/2-practice-tdd/repeat-str.js
@@ -4,9 +4,6 @@ function repeatStr(str, count) {
 if (count < 0) {
     throw new Error("Negative count is not allowed");
 
-  } else if (count === 0) {
-    return "";
-
   } else {
     return Array.from({ length: count }, () => str).join("");
   }
--- a/Sprint-3/2-practice-tdd/repeat-str.js
+++ b/Sprint-3/2-practice-tdd/repeat-str.js
@@ -2,7 +2,7 @@ function repeatStr(str, count) {
   // Your implementation of this function must *not* call String.prototype.repeat (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat).
   // The goal is to re-implement that function, not to use it.
 if (count < 0) {
-    throw new Error("Something went wrong");
+    throw new Error("Negative count is not allowed");
 
   } else if (count === 0) {
     return "";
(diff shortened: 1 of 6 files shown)


Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,7 +1,7 @@
 function repeatStr() {
   // Your implementation of this function must *not* call String.prototype.repeat (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat).
   // The goal is to re-implement that function, not to use it.
-  return "hellohellohello";
+  return str.repeat(count);
```
LonMcGregor: Did you read the comment in this task? The goal is to implement without using the repeat method. Can you try again?

LonMcGregor (commented):
Good work on the first two tasks, you need to re-attempt the repeat task however

Sprint-3/2-practice-tdd/repeat-str.js
joanne342: Yes, Sorry, I overlooked it.  I've changed it to

return Array.from({ length: count }, () => str).join("");

Sprint-3/2-practice-tdd/repeat-str.js
```diff
@@ -1,7 +1,7 @@
 function repeatStr() {
   // Your implementation of this function must *not* call String.prototype.repeat (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat).
   // The goal is to re-implement that function, not to use it.
-  return str.repeat(count);
+  return Array.from({ length: count }, () => str).join("");
```
LonMcGregor: This is an interesting approach. I can see what you are going for. Have you tried testing this function as it is written? There might be something still missing.

LonMcGregor (commented):
Almost done, still some changes needed though

Sprint-3/2-practice-tdd/repeat-str.js
joanne342: I didn't test it because I'm on my work laptop and I'm not allowed to install anything but I made a codespace and tested it now and added a conditional in case of a negative count.

LonMcGregor (commented):
Great work! always a good idea to fully test before submitting a PR. this task is complete now

illicitonion:

PR #1363, opened 2026-06-18
London | 26-ITP-May | Martin Mwaka | Sprint 2 | Coursework Exercises

Temceo:
Self checklist

## Changelist
JavaScript mandatory coursework exercises for sprint 2

### DIFF
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -15,5 +15,12 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
+  // return the BMI of someone based off their weight and height
+  const heightSquared = Number((height * height).toFixed(2));
+  const bmi = Number(weight / heightSquared).toFixed(1);
+  return bmi;
 }
+
+console.log(calculateBMI(70, 1.73));
+console.log(calculateBMI(90, 1.99));
+console.log(calculateBMI(80, 1.68));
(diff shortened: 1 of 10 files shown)


Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,12 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+  // return the BMI of someone based off their weight and height
+  const heightSquared = Number((height * height).toFixed(2));
```
LonMcGregor: Do you need to do the fixed and number conversion on this line 19?

LonMcGregor (commented):
Good work, I just have one question about one of the tasks

Sprint-2/3-mandatory-implement/1-bmi.js
Temceo: Converting to number has no effect as toFixed ensures text is returned.  I have update this line as follows:  return (weight / heightSquared).toFixed(1);

LonMcGregor: I was more asking about why you do tofixed(2) on the first line, then do another tofixed(1) on the final line. toFixed is used for changing the presentation of the final output, so why are you using it as an earlier part of the calculation?

Temceo: I understand.  I have refactored this to a single line now

LonMcGregor (commented):
Great! It's a good idea to avoid rounding off early in calculations as that can introduce inaccuracies in the result. This task is complete now

illicitonion:

PR #1367, opened 2026-06-20
Glasgow | 26-ITP-May | Niangh Ciang | Sprint 2 | Coursework

Niangh-Ciang:
Self checklist

## Changelist

Completed the courwork exercises in Sprint 2.

### DIFF
--- a/Sprint-2/2-mandatory-debug/2.js
+++ b/Sprint-2/2-mandatory-debug/2.js
@@ -2,8 +2,9 @@
 
 // Predict the output of the following code:
 // =============> Write your prediction here
+//I think all three console.logs will output 3 because the function always uses the global variable.
 
-const num = 103;
+/*const num = 103;
 
 function getLastDigit() {
   return num.toString().slice(-1);
@@ -11,14 +12,29 @@ function getLastDigit() {
 
 console.log(`The last digit of 42 is ${getLastDigit(42)}`);
 console.log(`The last digit of 105 is ${getLastDigit(105)}`);
-console.log(`The last digit of 806 is ${getLastDigit(806)}`);
+console.log(`The last digit of 806 is ${getLastDigit(806)}`);*/
 
 // Now run the code and compare the output to your prediction
 // =============> write the output here
+/*The last digit of 42 is 3
+The last digit of 105 is 3
+The last digit of 806 is 3*/
+
 // Explain why the output is the way it is
 // =============> write your explanation here
+/*The output is the way it is because the variable num (103) is declared 
+in the global scope. The function getLastDigit always uses this global variable
+ instead of the argument passed in, so it always returns the last digit of 103, which is 3.*/
+
 // Finally, correct the code to fix the problem
 // =============> write your new code here
+function getLastDigit(num) {
+  return num.toString().slice(-1);
+}
+console.log(`The last digit of 103 is ${getLastDigit(103)}`);
+console.log(`The last digit of 42 is ${getLastDigit(42)}`);
+console.log(`The last digit of 105 is ${getLastDigit(105)}`);
+console.log(`The last digit of 806 is ${getLastDigit(806)}`);
 
 // This program should tell the user the last digit of each number.
 // Explain why getLastDigit is not working properly - correct the problem
--- a/Sprint-2/3-mandatory-implement/1-bmi.js
+++ b/Sprint-2/3-mandatory-implement/1-bmi.js
@@ -15,5 +15,7 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
+  let bmi = weight / (height * height);
+  return bmi.toFixed(1);
 }
+console.log(calculateBMI(70, 1.73));
--- a/Sprint-2/3-mandatory-implement/2-cases.js
+++ b/Sprint-2/3-mandatory-implement/2-cases.js
@@ -16,6 +16,7 @@
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
 
 function snakeCase(str) {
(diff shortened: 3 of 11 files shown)


Sprint-2/2-mandatory-debug/2.js
```diff
@@ -2,23 +2,39 @@
 
 // Predict the output of the following code:
 // =============> Write your prediction here
+//I think all three console.logs will output 3 because the function always uses the global variable.
 
-const num = 103;
+/*const num = 103;
```
cjyuan, line 8: Note: Deleting the global `num` is optional. Within the function block, `num` is resolved to the parameter `num`.
If you are interested in the topic, you can looking up these two concepts, ***identifier scope*** and
***identifier resolution***, in the context of JavaScript programming. ChatGPT can give a good explanation.

Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,7 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+  let bmi = weight / (height * height);
+  return bmi.toFixed(1);
```
cjyuan, line 19: What **type** of value do you expect your function to return? A number or a string?

```javascript
  
```

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,8 @@
 // You will need to come up with an appropriate name for the function
 // Use the MDN string documentation to help you find a solution
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
+
+function snakeCase(str) {
+  return str.toUpperCase().replace(" ", "_");
+}
+console.log(snakeCase("hello there"));
```
cjyuan, line 21: Could you also check if your function works as expected with this argument?
```
console.log(snakeCase("a b c d"));
```

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -7,6 +7,13 @@ function formatAs12HourClock(time) {
...
   }
+  if (hours === 12) {
+    return `${hours}:00 pm`;
+  }
+  if (hours === 0) {
+    const minutes = time.slice(3, 5);
+    return `12:${minutes} am`;
+  }
   return `${time} am`;
 }
```
cjyuan, line 18: First, without running the code, predict the output from the following function calls.
Then, check if your function is returning the value you expected?

```
console.log( formatAs12HourClock("12:35") );
console.log( formatAs12HourClock("01:01") );
console.log( formatAs12HourClock("13:01") );
```

Sprint-2/2-mandatory-debug/2.js
Niangh-Ciang: I researched more about identifier scope and identifier resolution.
When I add the parameter num in function getLastDigit(num), it creates a new local identifier inside the function.
Because of scope rules, this local num only exists inside the function.
During identifier resolution, JavaScript always chooses the local identifier first before checking the global one.
So after adding the parameter, it doesn’t matter if the global const num = 103 is still there.

Sprint-2/3-mandatory-implement/1-bmi.js
Niangh-Ciang: My function returns a string because toFixed(1) always returns a string.
If I wanted it to return a number instead, I could convert it using Number(bmi.toFixed(1)).

Sprint-2/3-mandatory-implement/2-cases.js
Niangh-Ciang: When I tested the function with "a b c d", it didn’t work as expected because it only replaced the first space. The output was "A_B C D".
The issue was that I used replace(" ", "_"), which only replaces one space.
I fixed it by using replace(/ /g, "_") so it replaces all spaces.

Sprint-2/5-stretch-extend/format-time.js
Niangh-Ciang: When I tested "12:35" and "13:01", the minutes were wrong because I originally put const minutes = time.slice(3, 5) inside the hours === 0 block, so only the midnight case kept the minutes.
For the other cases (hours > 12 and hours === 12), the function always returned :00.
I fixed it by moving const minutes to the top of the function and using ${minutes} in every return statement. Now all the tests pass.

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,14 +4,14 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
+  const minutes = time.slice(3, 5);
```
cjyuan, line 7: `.slice()` supports negative index. So we could also extract the last two characters as `time.slice(-2)`.

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -48,3 +48,21 @@ console.assert(
...
+console.assert(
+  currentOutput7 === targetOutput7,
+  `current output: ${currentOutput7}, target output: ${targetOutput7}`
+);
+const currentOutput8 = formatAs12HourClock("13:01");
+const targetOutput8 = "1:01 pm";
+console.assert(
+  currentOutput8 === targetOutput8,
+  `current output: ${currentOutput8}, target output: ${targetOutput8}`
+);
```
cjyuan, line 68: If "01:01" is converted to "01:01 am", it is probably reasonable for the caller to expect "13:01" to be converted to "01:01 pm".

```
01:00 am
1:00 pm
12:00 am
01:00 pm
```

cjyuan (commented):
Changes look good. Your responses were clear and to the point. Excellent job!

illicitonion:

PR #1370, opened 2026-06-21
London | 26-ITP-May | Zadri Abdule | Sprint 3 | Practice tdd

Zadri415:
## Learners, PR Template

Self checklist

### **Changelist**

Implemented & added Jest tests for Sprint-3 practice TDD.

### DIFF
--- a/Sprint-3/2-practice-tdd/count.js
+++ b/Sprint-3/2-practice-tdd/count.js
@@ -1,5 +1,50 @@
+/**
+ * countChar(stringOfCharacters, findCharacter)
+ *
+ * Count the number of times a single character appears in a string.
+ *
+ * Behaviour / contract:
+ * - The function is case-sensitive: 'a' ≠ 'A'.
+ * - `stringOfCharacters` may be `null` or `undefined`; in that case the function returns 0.
+ * - `findCharacter` is treated as a single character (a single Unicode code point).
+ *   If you pass a longer string it will never match.
+ *
+ * Parameters:
+ * @param {string|null|undefined} stringOfCharacters - The string to search through.
+ * @param {string} findCharacter - The single character to count (should be length 1).
+ *
+ * Returns:
+ * @returns {number} The number of occurrences of `findCharacter` in `stringOfCharacters`.
+ *
+ * Throws:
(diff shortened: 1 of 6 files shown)


Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,50 @@
+/**
+ * countChar(stringOfCharacters, findCharacter)
+ *
+ * Count the number of times a single character appears in a string.
+ *
+ * Behaviour / contract:
+ * - The function is case-sensitive: 'a' ≠ 'A'.
+ * - `stringOfCharacters` may be `null` or `undefined`; in that case the function returns 0.
+ * - `findCharacter` is treated as a single character (a single Unicode code point).
+ *   If you pass a longer string it will never match.
```
cjyuan, line 10: If these behaviors are expected from the function, then in TDD we should prepare tests to ensure the function meet all these requirements, including 
- A test to show the function is case-sensitive: 'a' ≠ 'A'.
- A test to show that the function will return 0 if the 2nd parameter is a string containing more than one character

Sprint-3/2-practice-tdd/get-ordinal-number.js
```diff
@@ -1,5 +1,25 @@
 function getOrdinalNumber(num) {
-  return "1st";
+  // Validate input: must be a number (not a string, null or undefined) and not NaN
+  if (typeof num !== "number" || Number.isNaN(num)) {
+    throw new TypeError("Input must be a number");
+  }
```
cjyuan, line 5: Note: A stricter check would be to ensure `num` is an integer.

I don't think ordinal numbers apply to non-integers.

cjyuan (commented):
Code looks good. I just have some suggestions.

cjyuan:
On separate note, the PR description is missing a "Checklist" heading (a level 2 heading)

Sprint-3/2-practice-tdd/get-ordinal-number.js
Zadri415: Thanks for feedback.

cjyuan:
Changes look good. Well done.

PR #1379, opened 2026-06-22
London | 26-ITP-May | Yonatan Teklemariam | Sprint 2 | Coursework

Yonatanteklemariam:
## Learners, PR Template

Self checklist

## Changelist

debugged the code and corrected the function along with the parameters

### DIFF
--- a/Sprint-2/1-key-errors/1.js
+++ b/Sprint-2/1-key-errors/1.js
@@ -5,13 +5,12 @@
 
 // Try playing computer with the example to work out what is going on
 
-
 function convertToPercentage(decimalNumber) {
-  return `${decimalNumber * 100}%`;
+  const decimalNumber = 0.5;
+  const percentage = `${decimalNumber * 100}%`;
+
+  return percentage;
 }
-
-console.log(convertToPercentage(0.1));
-
 
 console.log(decimalNumber);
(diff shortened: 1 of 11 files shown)


Sprint-2/1-key-errors/1.js
```diff
@@ -1,20 +1,34 @@
...
+/* ===========> function convertToPercentage(decimalNumber) {
+  let decimalNum = 0.5;
+  const percentage = `${decimalNum * 100}%`;
+
+
+  return percentage;
+}
+const result = convertToPercentage();
+console.log(result);
+*/
```
cjyuan, line 25: You get rid of the syntax error but the function still does not work quite correctly.
I expect `convertToPercentage(0.1)` to return `"10%"`, but it returns `"50%"` instead.

cjyuan (commented):
Most of the code looks good.

Sprint-2/1-key-errors/1.js
Yonatanteklemariam: yes i worked it out as well and it did the same because decimalNumber argument has already been defined inside the function as 0.5 and for that matter, it returned 50%. but here i've corrected the function and can return any decimal number passed through it as argument

cjyuan:
Changes look good.

illicitonion:

PR #1384, opened 2026-06-22
London | 26-ITP-May | Russom Gebremeskel | Sprint 1 | Coursework

russom-g:
## Learners, PR Template

Self checklist

## Changelist

By referring to documentation, syntax and other type of errors fixed. Some of the program code explained line by line.

### DIFF
--- a/Sprint-1/2-mandatory-errors/3.js
+++ b/Sprint-1/2-mandatory-errors/3.js
@@ -1,4 +1,4 @@
-const cardNumber = 4533787178994213;
+const cardNumber = "4533787178994213"; // parentises
 const last4Digits = cardNumber.slice(-4);
 
 // The last4Digits variable should store the last 4 digits of cardNumber
@@ -7,3 +7,11 @@ const last4Digits = cardNumber.slice(-4);
 // Then run the code and see what error it gives.
 // Consider: Why does it give this error? Is this what I predicted? If not, what's different?
 // Then try updating the expression last4Digits is assigned to, in order to get the correct value
+
+// Answer
+// "TypeError: cardNumber.slice is not a function" is the error that it was given when running the code. 
+// This error occurs because the `slice` method is a string method, and `cardNumber` is a number, not a string. 
+// Therefore, we cannot use `slice` directly on a number.
+// To fix the issue, we need to convert `cardNumber` to a string by adding parentises.
+
+  
--- a/Sprint-1/3-mandatory-interpret/1-percentage-change.js
+++ b/Sprint-1/3-mandatory-interpret/1-percentage-change.js
@@ -2,7 +2,7 @@ let carPrice = "10,000";
 let priceAfterOneYear = "8,543";
 
 carPrice = Number(carPrice.replaceAll(",", ""));
-priceAfterOneYear = Number(priceAfterOneYear.replaceAll("," ""));
+priceAfterOneYear = Number(priceAfterOneYear.replaceAll("," , ""));
 
 const priceDifference = carPrice - priceAfterOneYear;
 const percentageChange = (priceDifference / carPrice) * 100;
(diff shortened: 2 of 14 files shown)


Sprint-1/2-mandatory-errors/3.js
```diff
@@ -1,4 +1,4 @@
-const cardNumber = 4533787178994213;
+const cardNumber = "4533787178994213"; // parentises
 const last4Digits = cardNumber.slice(-4);
```
cjyuan, line 2: Suppose you were not allowed to modify the statement `const cardNumber = 4533787178994213;`
How would you modify the code (through type conversion) to still being able to use `.slice(-4)` to extract the last 4 digits from the given number.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -20,3 +20,19 @@ console.log(`The percentage change is ${percentageChange}`);
 // d) Identify all the lines that are variable declarations
 
 // e) Describe what the expression Number(carPrice.replaceAll(",","")) is doing - what is the purpose of this expression?
+
+// Answer
+// a) Line 4: Number(carPrice.replaceAll(",", "")); and carPrice.replaceAll(",", "")
+// a) Line 5: Number(priceAfterOneYear.replaceAll("," "")); and priceAfterOneYear.replaceAll("," "")
+
+// b) The error is coming from line 5. The error is occurring in the replaceAll method a comma is missing between ("," "") it should be (",", "") to fix the problem.
```
cjyuan, line 28: We could also more precisely describe "A comma is missing between `","` and `""` in the function call" as:  

Sprint-1/4-stretch-explore/chrome.md
```diff
@@ -9,10 +9,16 @@ Let's try an example.
...
+
 Now try invoking the function `prompt` with a string input of `"What is your name?"` - store the return value of your call to `prompt` in an variable called `myName`.
 
 What effect does calling the `prompt` function have?
+
+- a new window popup saying what is your name with a text input field for user to type.
+
 What is the return value of `prompt`?
+
+- A text string that input by the user is returned.
```
cjyuan, line 24: What if the user clicked "Cancel" instead of "OK"?

Sprint-1/4-stretch-explore/objects.md
```diff
@@ -5,12 +5,20 @@ In this activity, we'll explore some additional concepts that you'll encounter i
...
 
 Try also entering `typeof console`
+- Object
 
 Answer the following questions:
 
 What does `console` store?
+
+- console means display or print a message 
+
```
cjyuan, line 21: I think the question is asking, "What does the object assigned to `console` store?"

Note: "console" does not mean "display or print a message".

Sprint-1/2-mandatory-errors/3.js
russom-g: Thanks for reviewing my code. I think I could have used toString() method as the following: 

Sprint-1/3-mandatory-interpret/1-percentage-change.js
russom-g: Double quotes?

cjyuan: A word describing the values passed to a function. It begins with 'A'

Sprint-1/4-stretch-explore/chrome.md
russom-g: I run the code again and if "Cancel " clicked "null" is returned.

cjyuan: Is the "null" you mentioned a string?

Sprint-1/4-stretch-explore/objects.md
russom-g: console stores object

Sprint-1/4-stretch-explore/chrome.md
russom-g: No. Because the user didn't type or input anything

Sprint-1/3-mandatory-interpret/1-percentage-change.js
russom-g: That is an arguments.

Sprint-1/4-stretch-explore/chrome.md
cjyuan: 1. When you says "null" is returned, I wasn't sure if you were referring to `null` (a value representing 'nothing') or a string "null". So which null was you referring to?

2. The user could type/input something and then clicked "Cancel". So it is not quite correct to say "the user didn't type or input anything".

russom-g: 1. I meant null representing nothing. 
2. You are right. I didn't explain it properly. The prompt didn't receive any value.

cjyuan:
Thanks for clarification. All good.

illicitonion:

PR #1409, opened 2026-06-26
West-Midlands |26-May-ITP |Maryam Janjua |Sprint 1 |Complete Sprint 1 Course Work

maryam-devio:
## Learners, PR Template

Self checklist

## Changelist

- In this task we need to solve JavaScript tasks. 
- I deeply understand the concept read documentation, solve tasks and tested changes.
- Theses exercises helped me to understand java-script syntax, built-in methods and data types briefly.


Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,21 @@ console.log(result);
...
+// This is a reminder operator. It divides one number by another and gives a reminder.
 
 // d) Interpret line 4, what does the expression assigned to totalMinutes mean?
-
+/* const totalMinutes = (movieLength - remainingSeconds) / 60 
+Firstly it will evaluate bracket i.e (movieLength - remainingSeconds) and then divide the value by 60.
+*/
 // e) What do you think the variable result represents? Can you think of a better name for this variable?
-
+// The result represents the how long the movie is, in hours, minutes and seconds. It can named as MovieDuration.
```
LonMcGregor: Does"MovieDuration" make enough of a distinction to the other variable name called "movieLength"?

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -12,14 +12,21 @@ console.log(result);
...
 // f) Try experimenting with different values of movieLength. Will this code work for all values of movieLength? Explain your answer
+// Yes, I changed the value of movieLength and it worked. The movieLength is the variable used in calculating other 
+// values and to evaluate the total length of movie.
```
LonMcGregor: It may work without errors, but are there any possible inputs where the output doesn't look quite right?

LonMcGregor (commented):
Well done doing the stretch tasks.

There's just one file I have comments about

Sprint-1/3-mandatory-interpret/2-time-format.js
maryam-devio: Thanks for the feedback. I agree that MovieDuration was too similar to movieLength, so I've renamed it to duration to make the distinction clearer.

Sprint-1/3-mandatory-interpret/2-time-format.js
maryam-devio: Yes, I found that although the code worked without errors, the output was not always formatted correctly. For example, when the hours, minutes, or seconds were less than 10 (e.g. 7 or 5), they were displayed as single digits instead of two digits. I fixed this by adding a helper function that adds a leading zero to single-digit values before formatting the duration.
Thank you for your feedback.

LonMcGregor:
Some follow up comments:
* Good answer of your testing the various inputs for the film
* Your new variable name "duration" is still very similar to the old one "movieDuration". think if you saw this code for the first time, would it be obvious which one is formatted a particular way and which one is just a number?

illicitonion:

PR #1417, opened 2026-06-27
Manchester | ITP-May-2026 | Szidonia Bodo | Sprint 1 | Coursework Exercises 

bodoszidi:
## Learners, PR Template

Self checklist

## Changelist

to explain the code line by line

### DIFF
--- a/Sprint-1/1-key-exercises/1-count.js
+++ b/Sprint-1/1-key-exercises/1-count.js
@@ -4,3 +4,6 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+
+// Line 3, we reassign a new value to the count variable,
+// the new value is count plus 1.
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -16,8 +16,8 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
-
-const dir = ;
-const ext = ;
+const dir = filePath.slice(1, 45);
+const ext = filePath.slice(-8);
+console.log(`The base part of ${dir} is ${ext}`);
 
 // https://www.google.com/search?q=slice+mdn
--- a/Sprint-1/1-key-exercises/3-paths.js
+++ b/Sprint-1/1-key-exercises/3-paths.js
@@ -16,8 +16,8 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
-const dir = filePath.slice(1, 45);
-const ext = filePath.slice(-8);
+const dir = filePath.slice(1, 16);
+const ext = filePath.slice(-4);
 console.log(`The base part of ${dir} is ${ext}`);
 
 // https://www.google.com/search?q=slice+mdn
--- a/Sprint-1/1-key-exercises/4-random.js
+++ b/Sprint-1/1-key-exercises/4-random.js
@@ -4,6 +4,13 @@ const maximum = 100;
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 
 // In this exercise, you will need to work out what num represents?
+console.log(Math.random())
+console.log(num);
 // Try breaking down the expression and using documentation to explain what it means
+
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+// num = Math.floor which will remove all the numbers after the decimal point 
+// (Math.random that will generate a random decimal number 
+// which will be times with the value of maximum - minimum + 1 + minimum ) 
+// which is 100 * the Math.random generated
--- a/Sprint-1/2-mandatory-errors/3.js
+++ b/Sprint-1/2-mandatory-errors/3.js
@@ -1,9 +1,14 @@
-const cardNumber = 4533787178994213;
+const cardNumber = "4533787178994213";
 const last4Digits = cardNumber.slice(-4);
+console.log(last4Digits);
 
 // The last4Digits variable should store the last 4 digits of cardNumber
 // However, the code isn't working
 // Before running the code, make and explain a prediction about why the code won't work
 // Then run the code and see what error it gives.
+//const last4Digits = cardNumber.slice(-4);
+                               
 // Consider: Why does it give this error? Is this what I predicted? If not, what's different?
+// The error comes up cause the value of the cardNumber is not a string it's a number
+//Once you change the value to a string the .slice()
 // Then try updating the expression last4Digits is assigned to, in order to get the correct value
--- a/Sprint-1/3-mandatory-interpret/1-percentage-change.js
+++ b/Sprint-1/3-mandatory-interpret/1-percentage-change.js
@@ -2,7 +2,7 @@ let carPrice = "10,000";
 let priceAfterOneYear = "8,543";
 
 carPrice = Number(carPrice.replaceAll(",", ""));
-priceAfterOneYear = Number(priceAfterOneYear.replaceAll("," ""));
+priceAfterOneYear = Number(priceAfterOneYear.replaceAll(",", ""));
 
 const priceDifference = carPrice - priceAfterOneYear;
 const percentageChange = (priceDifference / carPrice) * 100;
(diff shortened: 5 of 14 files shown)


Sprint-1/1-key-exercises/1-count.js
```diff
@@ -4,3 +4,6 @@ count = count + 1;
 
 // Line 1 is a variable declaration, creating the count variable with an initial value of 0
 // Describe what line 3 is doing, in particular focus on what = is doing
+
+// Line 3, we reassign a new value to the count variable,
+// the new value is count plus 1.
```
cjyuan, line 9: Operation like `count = count + 1` is very common in programming, and there is a programming term describing such operation.

Could you find out what one-word programming term describes the operation on line 3?

Sprint-1/1-key-exercises/3-paths.js
```diff
@@ -16,8 +16,8 @@ console.log(`The base part of ${filePath} is ${base}`);
 
 // Create a variable to store the dir part of the filePath variable
 // Create a variable to store the ext part of the variable
-
-const dir = ;
-const ext = ;
+const dir = filePath.slice(1, 16);
+const ext = filePath.slice(-4);
+console.log(`The base part of ${dir} is ${ext}`);
```
cjyuan, line 21: Could you make these statements to work for any valid path? For example, they could still work correctly if we changed line 12 to
```
const filePath = "/Users/jacknguyen448/cyf/Module-2/Sprint-1/package.json";
```

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -4,6 +4,13 @@ const maximum = 100;
 const num = Math.floor(Math.random() * (maximum - minimum + 1)) + minimum;
 
 // In this exercise, you will need to work out what num represents?
+console.log(Math.random())
+console.log(num);
 // Try breaking down the expression and using documentation to explain what it means
+
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+// num = Math.floor which will remove all the numbers after the decimal point
```
cjyuan, line 13: Could you look up the differences among `Math.round()`, `Math.trunc()`, `Math.floor()`, `Math.ceil()`?
All of these functions remove the decimal part of a number, but in different manner.

Sprint-1/1-key-exercises/4-random.js
```diff
@@ -4,6 +4,13 @@ const maximum = 100;
...
+console.log(Math.random())
+console.log(num);
 // Try breaking down the expression and using documentation to explain what it means
+
 // It will help to think about the order in which expressions are evaluated
 // Try logging the value of num and running the program several times to build an idea of what the program is doing
+// num = Math.floor which will remove all the numbers after the decimal point 
+// (Math.random that will generate a random decimal number 
+// which will be times with the value of maximum - minimum + 1 + minimum ) 
+// which is 100 * the Math.random generated
```
cjyuan, line 16: `Math.random()` returns a random number within a range.

Could you describe what each of these expressions does, and the range of the numbers it may produce?

1. `Math.random()`

---
Note: To describe a range of numbers, we could use the math ***interval notation***: 
- `[`, `]` => inclusion
- `(`, `)` => exclusion

Sprint-1/2-mandatory-errors/3.js
```diff
@@ -1,9 +1,14 @@
-const cardNumber = 4533787178994213;
+const cardNumber = "4533787178994213";
 const last4Digits = cardNumber.slice(-4);
```
cjyuan, line 2: Suppose you were not allowed to modify the statement `const cardNumber = 4533787178994213;`
How would you modify the code (through type conversion) to still being able to use `.slice(-4)` to extract the last 4 digits from the given number.

Sprint-1/3-mandatory-interpret/1-percentage-change.js
```diff
@@ -12,11 +12,16 @@ console.log(`The percentage change is ${percentageChange}`);
 // Read the code and then answer the questions below
 
 // a) How many function calls are there in this file? Write down all the lines where a function call is made
+// there is 5 calls: 2* Number() 2* .replaceAll() AND 1* .log()
 
 // b) Run the code and identify the line where the error is coming from - why is this error occurring? How can you fix this problem?
+// line 5 missing a comma: .replaceAll("," "")
```
cjyuan, line 18: We could more precisely describe "A comma is missing between `","` and `""` in the function call" as:  

What is this programming term that refers to the values passed to a function? It begins with an 'a'.

Sprint-1/3-mandatory-interpret/2-time-format.js
```diff
@@ -11,15 +13,23 @@ console.log(result);
...
 // c) Using documentation, explain what the expression movieLength % 60 represents
 // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators
+// the logic is to calculate the movieLength remainder - operator returns the remainder left over when one operand is divided by a second operand. 
+// It always takes the sign of the dividend.
 
 // d) Interpret line 4, what does the expression assigned to totalMinutes mean?
+// to calculate the movie length into minutes
 
 // e) What do you think the variable result represents? Can you think of a better name for this variable?
+// the result represent the movie length - hours minutes and second in a string - movieTime
```
cjyuan, line 32: The name `movieTime` does not quite indicate the value stored in the variable 
is a formatted string in the form `"2:12:02"`.

Could you suggest a more descriptive name?

Sprint-1/4-stretch-explore/chrome.md
```diff
@@ -10,9 +10,10 @@ Let's try an example.
...
-What effect does calling the `alert` function have?
+What effect does calling the `alert` function have? - alert is a pop up display message with an OK button - you can close once you click on the OK button 
 
 Now try invoking the function `prompt` with a string input of `"What is your name?"` - store the return value of your call to `prompt` in an variable called `myName`.
 
-What effect does calling the `prompt` function have?
-What is the return value of `prompt`?
+What effect does calling the `prompt` function have? - the prompt will also display a message usually question  - however this time you will have an empty field to enter the value or answer - a cancel and a OK button
+
+What is the return value of `prompt`? - whatever the user inputs into the empty field
```
cjyuan, line 19: What if the a enters something and then click "Cancel" instead of "OK?

Sprint-1/1-key-exercises/1-count.js
bodoszidi: The term would be incrementing ++ or if it's the opposite, decrementing --

Sprint-1/1-key-exercises/3-paths.js
bodoszidi: Yes, if we use the already created slash index and create a new dot index, the file path length can change at any point. The code below will find the "/" and the "." and we no longer need to count the length of our filepath: 
const lastDotIndex = filePath.lastIndexOf(".")
const dir = filePath.slice(0, lastSlashIndex);
const ext = filePath.slice(lastDotIndex + 1);

Sprint-1/1-key-exercises/4-random.js
bodoszidi: Math.round() - rounds the number to the nearest -> .1 - .4 ⬇️  & .5 - .9  ⬆️
Math.trunc() - removes all the numbers after the decimal point - without rounding the number
Math.floor() -  rounds down and returns the largest integer less than or equal to a given number. rounding the number down ⬇️
Math.ceil() - rounds the given number to the smallest integer greater than or equal to a given number. rounding the number ⬆️ 

The above answer I gave was actually correct for Math.trunc() - Math.floor() will actually round the number to the largest integer less than or equal to the given number - if we have a negative integer, we would get the lower number back, as Math.floor(-3.03) would return -4

cjyuan: Your approach is correct, and you are right that either Math.trunc() or Math.floor() works. I just want you to be aware of different math functions.

Sprint-1/1-key-exercises/4-random.js
bodoszidi: 1. Math.random() - generates a decimal number between [0, 1] - but excluding 1
2. Math.random() * (maximum - minimum + 1) - the generated decimal number * 100 - [0, 100]
3. Math.floor(Math.random() * (maximum - minimum + 1)) - the generated decimal number * 100 - Math.floor() round the number to the nearest lower number  - [0, 99]
4. Math.floor(Math.random() * (maximum - minimum + 1)) + minimum 
- (the generated decimal * 100) +1 - Math.floor() round the number to the nearest lower number [1, 100]

Sprint-1/2-mandatory-errors/3.js
bodoszidi: If I can't modify the number would need to be converted to a string:
const cardNumber = 4533787178994213 
const cardNumberString = cardNumber.toString()
const last4Digit = cardNumberString.slice(-4)

or 

const cardNumber = 4533787178994213 
const last4Digit = cardNumber.toString().slice(-4)

Sprint-1/3-mandatory-interpret/1-percentage-change.js
bodoszidi: The comma is missing between the two arguments "," and ""

Sprint-1/3-mandatory-interpret/2-time-format.js
bodoszidi: movieLengthFormatted - would represent the value in a clearer way, also would indicate to the given variable movieLength - but in a formatted output

Sprint-1/4-stretch-explore/chrome.md
bodoszidi: If the user cancels the prompt, the returned value will be null.

cjyuan:
Well done addressing all comments, and they are all correct.

illicitonion:

PR #1429, opened 2026-06-30
London | 26-ITP_May | Vito Moratti | Sprint 2 | Coursework/sprint 2

vmoratti:
## Learners, PR Template

Self checklist

## Changelist

I have completed all the exercises in the course work.


Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -13,7 +13,10 @@
...
 // Then when we call this function with the weight and height
 // It should return their Body Mass Index to 1 decimal place
-
+//
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+    let bmi = weight / (height **2 )// weight divided by height squared.
+    return  `Your Body Mass Index is ${bmi.toFixed(1)} units`// toFixed(1) insures that result is displayed
```
LonMcGregor: Have a look at the comment above. It says the function should "return their Body Mass Index to 1 decimal place". Your function works, but does it meet this specific criteria?

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -1,15 +1,19 @@
...
-// Your task is to write tests for as many different groups of input data or edge cases as you can, and fix any bugs you find.
+// Your task is to write tests for as many different groups of input data or edge cases as you can, and fix any
+// bugs you find.
 
 function formatAs12HourClock(time) {
-  const hours = Number(time.slice(0, 2));
-  if (hours > 12) {
-    return `${hours - 12}:00 pm`;
-  }
+    if (Number(time.slice(0, 2)) === 12) {
```
LonMcGregor: Is there something you could change here so you don't need to keep slicing the hours repeatedly?

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -23,3 +27,24 @@ console.assert(
   currentOutput2 === targetOutput2,
   `current output: ${currentOutput2}, target output: ${targetOutput2}`
 );
+const currentOutput3 = formatAs12HourClock("00:00");//
```
LonMcGregor, line 40: Have you considered some test cases where the time is not exactly at 00 minutes?

vmoratti: Thank you,  @LonMcGregor, I have not realised that my return in some cases was not responsive to different amount of minutes. I have now fixed that.

Sprint-2/5-stretch-extend/format-time.js
vmoratti: I have replaced frequent slicing with variables.

Sprint-2/3-mandatory-implement/1-bmi.js
vmoratti: Sorry, but i don't understand what am I missing.  The function returns correct BMI with one decimal place using  "bmi.toFixed(1)" method

LonMcGregor: Your calculation is correct, but read carefully what the task is asking for. It says
`It should return their Body Mass Index to 1 decimal place`

Is that what your function does? You are returning a string here.

LonMcGregor:
Good work on the time task - it is done now. See my earlier comment about the BMI calc.

Sprint-2/3-mandatory-implement/1-bmi.js
vmoratti: Oh, I see. I got overly enthusiastic with the result, when only bmi number is expected.
Thank you for pointing that out.

illicitonion:

PR #1430, opened 2026-06-30
London | 26-ITP-May | Tomislav Dukez | Sprint 2 | Coursework

tomdu3:
## Learners, PR Template

Self checklist

## Changelist

Tasks of the Sprin2 of the Course Work for the Module Structuring and Testing Data


Sprint-2/3-mandatory-implement/1-bmi.js
```diff
@@ -15,5 +15,12 @@
 // It should return their Body Mass Index to 1 decimal place
 
 function calculateBMI(weight, height) {
-    // return the BMI of someone based off their weight and height
-}
\ No newline at end of file
+  // return the BMI of someone based off their weight and height
+  const bmi = Math.round((weight / (height * height)) * 10) / 10;
```
LonMcGregor: Here you are multiplying by 10 then dividing by 10, which seems like a complex solution. Is there an easier function you could use to get the 1 decimal place?

Sprint-2/5-stretch-extend/format-time.js
```diff
@@ -4,22 +4,34 @@
 
 function formatAs12HourClock(time) {
   const hours = Number(time.slice(0, 2));
-  if (hours > 12) {
-    return `${hours - 12}:00 pm`;
+  if (hours === 0 || hours === 24) {
+    return `12:${time.slice(3, 5)} am`;
```
LonMcGregor: Here there are a couple of slices that you use repeatedly. Is there something you could change that would mean you only have to do them once? This change would make the code more readable and maintainable.

Sprint-2/3-mandatory-implement/1-bmi.js
tomdu3: Thanks. I have changed the code to use Number and toFixed. I was thinking more of not changing the data type from Number to String and back to Number again.

tomdu3:
awesome. Yeah, I have redundant slicing I could have done at the beginning and just insert the values later after conditional checks. thanks.

illicitonion:

PR #1431, opened 2026-06-30
London| 26-ITP-May | Martin Mwaka | Sprint 3 | Practice tdd

Temceo:
Self checklist

## Changelist

Completed Test Driven Development tasks

### DIFF
--- a/Sprint-3/2-practice-tdd/count.js
+++ b/Sprint-3/2-practice-tdd/count.js
@@ -1,5 +1,10 @@
 function countChar(stringOfCharacters, findCharacter) {
-  return 5
+  if (stringOfCharacters.length === 0 || findCharacter.length === 0) return 0;
+  let count = 0;
+  for (const char of stringOfCharacters) {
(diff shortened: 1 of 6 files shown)


Sprint-3/2-practice-tdd/count.js
```diff
@@ -1,5 +1,10 @@
 function countChar(stringOfCharacters, findCharacter) {
-  return 5
+  if (stringOfCharacters.length === 0 || findCharacter.length === 0) return 0;
```
LonMcGregor: What does this if condition do in this function?

Temceo: This condition returns 0 if either the string or the character are empty.  I have updated the code to make it more readable

Temceo: The first line is not needed as count is initialised to zero later in the function.  If count is not incremented it will return zero.  I have now removed it.

illicitonion:

PR #1435, opened 2026-07-01
Manchester | 26-ITP-May | Abdu Hassen | Sprint 2 | structure and testing data 

Abduhasen:
## Learners, PR Template

Self checklist

## Changelist

1.interpreting of codes with explanation. 
2.fixing errors. 
3.writing a code depending on a given statement.
4.debugging of code.

## Questions

N/A

### DIFF
--- a/Sprint-2/2-mandatory-debug/1.js
+++ b/Sprint-2/2-mandatory-debug/1.js
@@ -1,5 +1,5 @@
 // Predict and explain first...
-//  =============> write your prediction here
+//  =============> write your prediction here: the function have nothing to return so it will log the sentences with undefined value.
 
 function sum(a, b) {
   return;
@@ -8,6 +8,11 @@ function sum(a, b) {
 
 console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
 
-// =============> write your explanation here
+// =============> write your explanation here: firstly the process start accepting that it is a function and takes
+//                                             the parameter to hold the place then it go through the function but by the 5th line
+//                                             the function return and the value will come out undefined
 // Finally, correct the code to fix the problem
-//  =============> write your new code here
+//  =============> write your new code here: function sum(a, b) {
+//                                              return a + b;
+//
+//                                             }
--- a/Sprint-2/2-mandatory-debug/2.js
+++ b/Sprint-2/2-mandatory-debug/2.js
@@ -1,7 +1,7 @@
 // Predict and explain first...
 
 // Predict the output of the following code:
-// =============> Write your prediction here
+// =============> Write your prediction here: the code is already declared a variable that can not be changed so the consol.log always be the same 3.
 
 const num = 103;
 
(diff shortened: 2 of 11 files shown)


Sprint-2/2-mandatory-debug/1.js
```diff
@@ -8,6 +8,11 @@ function sum(a, b) {
 
 console.log(`The sum of 10 and 32 is ${sum(10, 32)}`);
 
-// =============> write your explanation here
+// =============> write your explanation here: firstly the process start accepting that it is a function and takes
+//                                             the parameter to hold the place then it go through the function but by the 5th line
+//                                             the function return and the value will come out undefined
```
Sambuxc: How does javascript see `undefined`?
What does this tell you about js that it returns `undefined`?

Sprint-2/2-mandatory-debug/2.js
```diff
@@ -14,11 +14,15 @@ console.log(`The last digit of 105 is ${getLastDigit(105)}`);
...
+// =============> write the output here: 3
 // Explain why the output is the way it is
-// =============> write your explanation here
+// =============> write your explanation here: the code is already declared a variable that can not be changed in the
+//                                             first place so when ever the process try to access another console.log it will take the same value as before
+//                                             so the consol.log always be the same.
 // Finally, correct the code to fix the problem
-// =============> write your new code here
+// =============> write your new code here: function getLastDigit(num) {
+//                                             return num.toString().slice(-1);
```
Sambuxc, line 24: What does `-1` represent when passing into `slice`?

Sprint-2/3-mandatory-implement/2-cases.js
```diff
@@ -14,3 +14,8 @@
 // You will need to come up with an appropriate name for the function
 // Use the MDN string documentation to help you find a solution
 // This might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
+
+function toCapitalizedSnakeCase(str) {
```
Sambuxc: Is this capitalising each word or does it uppercase the whole word, because your function name tells me one thing but its return output gives me another.

Sprint-2/2-mandatory-debug/1.js
Abduhasen: 1. javascript sees `undefined` as if there is no value assigned and it is returning with out a value.
2. this tell me about javascript if we didn't assigned a value to and if the terminate the function with out returning it will see it as value not given to it and it doesn't know the value.

Sprint-2/2-mandatory-debug/2.js
Abduhasen: 1. when using `slice` if we apple `-1` to the `slice` it represent that it is starting from the last index but not including it.

Sprint-2/3-mandatory-implement/2-cases.js
Abduhasen: 1.the function is about changing  all the given string to uppercase and also to snakecase and i changed the function name to more understandable name `toUpperSnakeCase`

Sprint-2/2-mandatory-debug/1.js
Sambuxc: One nice video explaining it a little further -> https://www.youtube.com/watch?v=EKsheWTwgok

Sprint-2/3-mandatory-implement/2-cases.js
Sambuxc: You will also need to update the `toCapitalize` variable name to match the intention of the function, otherwise its still doesn't read correctly.

Once this is done I can approve your PR :) 

Well done

Sprint-2/2-mandatory-debug/1.js
Abduhasen: So good way to explain  `undefined` is when the variable exists but no values has been assigned to it, also java script assignee `undefined` when the variable have no value.

Sprint-2/3-mandatory-implement/2-cases.js
Abduhasen: I changed the variable name to match the intention of the function I named it `convertToUpperSnakeCase`. Thank you for the feedback's they are constructive and i gain more concepts.

illicitonion:

